PR #6, opened 2024-05-05
Implemented the drop down component.

Skorpios604:
Implemented the drop down component for the Uptime Monitor.


popnfresh234:
Nice work, looks great!

Perhaps consider adding proptypes for props validation?  Not sure what our policy is going to be on prop validation.

```

DropDown.propTypes = {
  id: PropTypes.string.isRequired,
  label: PropTypes.string.isRequired,
  options: PropTypes.array.isRequired,
  onChange: PropTypes.func.isRequired,
  value: PropTypes.string.isRequired,
};
```

`React` import can be safely removed too I think, appears unused

Skorpios604:
That is a great idea! Thank you for that. I have updated the code. Do you mind reviewing it?

popnfresh234:
Pulled the pr branch and ran it, all seems fine and the linter is clear, looks good to me!

PR #8, opened 2024-05-06
Updated the root structure of Client side

veyselboybay:
Created routing structure using react-router-dom and added some links and updated the folder structure.

### DIFF
--- a/Client/src/App.jsx
+++ b/Client/src/App.jsx
@@ -1,43 +1,23 @@
-import * as React from 'react';
-import { useState } from 'react'
+
+import { Routes, Route } from 'react-router-dom'
 import './App.css'
-import '@fontsource/roboto/300.css';
-import '@fontsource/roboto/400.css';
-import '@fontsource/roboto/500.css';
-import '@fontsource/roboto/700.css';
-import DropDown from './Dropdown';
+import NotFound from './Pages/NotFound'
+import Login from './Pages/Login'
+import Register from './Pages/Register'
+import HomeLayout from './Layouts/HomeLayout'
+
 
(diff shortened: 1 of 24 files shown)


Client/src/App.jsx
```diff
@@ -1,43 +1,23 @@
...
-      <div className="app-container" style={{ width: '500px' }}>
-      <DropDown
-        id="team-member-autocomplete"
-        options={teamMembers}
-        label="Select team member"
-        value={selectedTeamMember}
-        onChange={handleTeamMemberChange}
-      />
-    </div>
+      <Routes>
```
popnfresh234, line 13: Nice to see the router in place

Client/src/main.jsx
```diff
@@ -2,9 +2,12 @@ import React from 'react'
 import ReactDOM from 'react-dom/client'
 import App from './App.jsx'
 import './index.css'
+import { BrowserRouter as Router, HashRouter } from 'react-router-dom'
```
popnfresh234, line 5: HashRouter unused at this time

popnfresh234 (approved):
Looks good to me!

Client/src/main.jsx
veyselboybay: The reason I put HashRouter is that route location is not detected when we use useLocation hook, I thought maybe we could switch it later for a solution if we encounter an issue.

popnfresh234: Got it, thanks for the explanation :+1:

veyselboybay: Thanks for the review 👍

PR #9, opened 2024-05-06
feat/theme provider

popnfresh234:
Added theme provider to the app.  Not sure if Utils is the right place for the theme, but I put it there for now.


veyselboybay:
Looks good Alex, thank you!

PR #11, opened 2024-05-06
Added start and end icon props to buttons to allow for images, made p…

popnfresh234:
…rop type for level more specific, added onClick prop so a function can be passed to onClick


Skorpios604:
Looks good to me! No conflicts. Well done.

PR #12, opened 2024-05-06
Added a link component, currently only have tertiary style

popnfresh234:
Added a Link component, I only see a tertiary link in the figma doc, so I've only provided style for the tertiary link at the moment.


Skorpios604:
Works good with no issues!

PR #17, opened 2024-05-09
Added user model, joi validation and register new user function

veyselboybay:


Server/models/user.js
```diff
@@ -0,0 +1,45 @@
...
+        type: Date,
+        default: Date.now
+    },
+    created_at: {
+        type: Date,
+        default: Date.now
+    }
+})
+
+UserSchema.pre('save', async function (next) {
```
popnfresh234, line 37: This is cool, I did not know about the `pre` function

popnfresh234 (approved):
Awesome, looks like a great start to me

Server/models/user.js
veyselboybay: Yeah It's pretty useful.

PR #18, opened 2024-05-09
Feat/stepper

popnfresh234:
Added a prorgress stepper component


Skorpios604 (approved):
I checked out your branch for the stepper feat and it seems in line with the standards and fully functional. Approved!

PR #19, opened 2024-05-10
Deleted Dropdown.jsx as it has been moved to the components directory.

Skorpios604:
It now has its own folder in the components directory.


popnfresh234 (changes requested):
It looks like the node modules potentially related to the MUI icons was inadvertently installed in the root project directory.  There's no gitignore in the root project directory so all node modules have been commited, and I don't believe a node project should be initiialized there anyway.

Recommend uninstalling the packages, deleting package.json and package.lock from the root directory if they are still there, and pushing again with the changes related to dropdown.

PR #23, opened 2024-05-10
Feat/monitors

popnfresh234:
Added monitors route and a few endpoints.  Added some documentation for those endpoints to readme, can be refactored if we decide on a different format for documentation.


Server/routes/monitorRoute.js
```diff
@@ -0,0 +1,8 @@
+const router = require("express").Router();
+const monitorController = require("../controllers/monitorController");
+const { getMonitorsByIdValidation } = require("../validation/joi");
```
veyselboybay: `getMonitorsByIdValidation` is not used, I think we can delete it.

Server/index.js
```diff
@@ -1,28 +1,32 @@
...
-const authRouter = require('./routes/authRoute')
-const { connectDbAndRunServer } = require('./configs/db')
-require('dotenv').config()
+const express = require("express");
+const helmet = require("helmet");
+const cors = require("cors");
+const authRouter = require("./routes/authRoute");
+const monitorRouter = require("./routes/monitorRoute");
+const { connectDbAndRunServer } = require("./configs/db");
+require("dotenv").config();
```
veyselboybay, line 7: Double quotes and semi-colons 😄 Looks great!

veyselboybay (commented):
I think you didn't put the mongoose operations in `monitorController` on purpose, we can keep them as is for now, and in our following pr we make sure adding them.

veyselboybay:
> I think you didn't put the mongoose operations in `monitorController` on purpose, we can keep them as is for now, and in our following pr we make sure adding them.

Please disregard my above comment, I just noticed your Fakedb file.

Server/index.js
popnfresh234: I think my linter is just set up to do that on save :laughing:  Happy to use whatever our team standard is!

popnfresh234:
> > I think you didn't put the mongoose operations in `monitorController` on purpose, we can keep them as is for now, and in our following pr we make sure adding them.
> 
> Please disregard my above comment, I just noticed your Fakedb file.

Yep, I just wanted to put something together super quick to test the schema and endpoints with postman.  We should be able to easily swap out when it's time to connect up to the database!

veyselboybay:
> > > I think you didn't put the mongoose operations in `monitorController` on purpose, we can keep them as is for now, and in our following pr we make sure adding them.
> > 
> > 
> > Please disregard my above comment, I just noticed your Fakedb file.
> 
> Yep, I just wanted to put something together super quick to test the schema and endpoints with postman. We should be able to easily swap out when it's time to connect up to the database!

I super liked your idea, but I think it adds an extra layer of complexity, and how are we gonna keep it up when our codebase got bigger?

popnfresh234:
> > > > I think you didn't put the mongoose operations in `monitorController` on purpose, we can keep them as is for now, and in our following pr we make sure adding them.
> > > 
> > > 
> > > Please disregard my above comment, I just noticed your Fakedb file.
> > 
> > 
> > Yep, I just wanted to put something together super quick to test the schema and endpoints with postman. We should be able to easily swap out when it's time to connect up to the database!
> 
> I super liked your idea, but I think it adds an extra layer of complexity, and how are we gonna keep it up when our codebase got bigger?

Yes there is definitely a benefit and a cost I think:

Benefit:  There is a layer of abstrtaction between the database and the controller.  If we call a DB class then we can easily swap out the DB without modifying the controller at all.  All the database methods will be in one location as well, which is nice organizationand willl be easier on maintenance.  Whatever uses the DB doesn't know anything about how the DB is implemetned, only that it has some methods that it can call like `connect()` `getAllUsers()` `getUserById()` etc.  This will make testing the DB much easier as well.

Cost:  Added layer of complexity.  There is now another file to manage, and you can't immediately see what the DB is actually doing from the controller.  Team members also have to remember to work through the DB since Javascript doesn't natively support interfaces.

If you'd like I can make an example DB interface with Mongo and so we can see how the swapping out of MongoDB and FakeDB would work, or if this approach is too complex for now we can drop it and perhaps revisit it later.

veyselboybay:
> > > > > I think you didn't put the mongoose operations in `monitorController` on purpose, we can keep them as is for now, and in our following pr we make sure adding them.
> > > > 
> > > > 
> > > > Please disregard my above comment, I just noticed your Fakedb file.
> > > 
> > > 
> > > Yep, I just wanted to put something together super quick to test the schema and endpoints with postman. We should be able to easily swap out when it's time to connect up to the database!
> > 
> > 
> > I super liked your idea, but I think it adds an extra layer of complexity, and how are we gonna keep it up when our codebase got bigger?
> 
> Yes there is definitely a benefit and a cost I think:
> 
> Benefit: There is a layer of abstrtaction between the database and the controller. If we call a DB class then we can easily swap out the DB without modifying the controller at all. All the database methods will be in one location as well, which is nice organizationand willl be easier on maintenance. Whatever uses the DB doesn't know anything about how the DB is implemetned, only that it has some methods that it can call like `connect()` `getAllUsers()` `getUserById()` etc. This will make testing the DB much easier as well.
> 
> Cost: Added layer of complexity. There is now another file to manage, and you can't immediately see what the DB is actually doing from the controller. Team members also have to remember to work through the DB since Javascript doesn't natively support interfaces.
> 
> If you'd like I can make an example DB interface with Mongo and so we can see how the swapping out of MongoDB and FakeDB would work, or if this approach is too complex for now we can drop it and perhaps revisit it later.

I think It is worth trying and seeing how it works. It'd be hard to revisit later since everything will be complicated enough. Let's connect on Monday before we merge this.

veyselboybay:
It looks great 👌 , thank you for your changes.

PR #24, opened 2024-05-11
Implemented the search feature.

Skorpios604:


MuhammadKhalilzadeh (approved):

Client/src/Components/Search/index.jsx
```diff
@@ -0,0 +1,41 @@
...
+import TextField from '@mui/material/TextField';
+import Chip from '@mui/material/Chip';
+import Box from '@mui/material/Box';
+
+const teamMembers = [
+  { title: 'John Doe'},
+  { title: 'Jane Smith'},
+  { title: 'Alex Johnson'},
+];
+
```
popnfresh234, line 11: Would be nice to have the JSdocs here, if they are included then VScode will suggest and autocomplete any params for the component.  It looks like this component doesn't take any params right now, but I imagine it will in the future.

```
/**
 * @component
 * @param {Object} props
 * @param {'primary' | 'secondary' | 'tertiary' | 'error'} props.level - The level of the button
 * @param {string} props.label - The label of the button
 * @param {boolean} [props.disabled] - Whether the button is disabled
 * @returns {JSX.Element}
 * @example
 * // Render an error button
 * <Button level="error" label="Error" disabled />
 */
```

It will definitely be helpful later on!

Client/src/Components/Search/index.jsx
```diff
@@ -0,0 +1,41 @@
...
+          <TextField
+            {...params}
+            label="Team Members"
+            placeholder="Favorites"
+          />
+        )}
+        renderTags={(value, getTagProps) =>
+          value.map((option, index) => (
+            <Chip
+              key={index}
```
popnfresh234, line 31: I think using the index as a key makes react throw a warning, might want to change this to something else, perhaps a proeprty of `option`?

popnfresh234 (approved):
Looks good to me!

Client/src/Components/Search/index.jsx
```diff
@@ -0,0 +1,41 @@
...
+
+const teamMembers = [
+  { title: 'John Doe'},
+  { title: 'Jane Smith'},
+  { title: 'Alex Johnson'},
+];
+
+export default function Search() {
+  return (
+    <Box padding={10}> {/* Add padding to the container */}
```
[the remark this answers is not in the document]
Skorpios604: That is a great idea! I will work on that right away.

Client/src/Components/Search/index.jsx
Skorpios604: Excellent idea! Implementing into the code now.

MuhammadKhalilzadeh (approved):

Client/src/Components/Search/index.jsx
```diff
@@ -26,12 +34,12 @@ export default function Search() {
           />
         )}
         renderTags={(value, getTagProps) =>
-          value.map((option, index) => (
+          value.map((option) => (
             <Chip
-              key={index}
+              key={option.title}
               variant="outlined"
```
popnfresh234, line 40: Not sure if we're ever going to need to use the other variants, but the variant type could potentially be passed in as a prop.  Probably not necessary for now, but something to keep in mind if we need to customize this component further

Client/src/Components/Search/index.jsx
```diff
@@ -26,12 +34,12 @@ export default function Search() {
           />
         )}
         renderTags={(value, getTagProps) =>
-          value.map((option, index) => (
+          value.map((option) => (
             <Chip
-              key={index}
+              key={option.title}
```
popnfresh234, line 39: Nice, console will be happy now :+1:

Client/src/Components/Search/index.jsx
```diff
@@ -2,16 +2,24 @@ import Autocomplete from '@mui/material/Autocomplete';
...
   { title: 'John Doe'},
   { title: 'Jane Smith'},
   { title: 'Alex Johnson'},
 ];
 
-export default function Search() {
+/**
+ * @component
+ * @returns {JSX.Element}
+ */
```
popnfresh234, line 16: I like to include an example in the JSdoc as well, it's always nice to have one around for reference

```
/**
* @example
* // Renders a Search component
* <Search variant="outlined" />
*/

Probably not neccessary at this point since this component doesn't take any props right now, but an example always helps clear up documentation that might be unclear or difficult to explain

popnfresh234 (approved):
Nice work, I think this is ready to merge in!

Client/src/Components/Search/index.jsx
Skorpios604: Gotchya gotchya! I will keep that in mind for next time.

PR #26, opened 2024-05-11
Syncings

MuhammadKhalilzadeh:
I tried to homogenize my code with yours. Please review it carefully to make sure I'm not missing anything.


popnfresh234 (changes requested):
It appears that node modules have been mistakenly installed in the root project directory.

Node modules for the front end should be installed in the `Client` directory, while modules for the server should be installed in the `Server` directory.

As the root folder does not have a `.gitignore` file the `node_modules` directory has been committed, which it should not be.

Recommend running `npm uninstall <module_name>` in the root directory to uninstall the modules, and then deleting `package.json` and the `node_modules` directory from the root directory if they still exist.

After installing the node modules in the `Client` directory perhaps submit a new pull request so we can review the rest of the code?  Thanks!

PR #27, opened 2024-05-12
Fixing for merge

MuhammadKhalilzadeh:
Synced and polished.
Please review carefully.


Client/src/components/TextFields/Website/websiteTextField.css
```diff
@@ -0,0 +1,86 @@
+:root {
+  --website-color-0: #344054;
+  --border-color: #d0d5dd;
+  --label-font-color: #475467;
+  --icon-color: #98a2b3;
+  --text-field-margin-0: 10px 20px;
+  --text-field-margin-1: 10px 20px 5px;
+  --font-family: "Roboto", "Helvetica", "Arial", sans-serif;
+}
```
popnfresh234, line 9: I think the font family can be set in the Material UI theme, that would be preferable to setting it at the component level.

If we set it at the Theme level and then reference the theme at the component level we can update the font everywhere at once if we need to.

I haven't looked deeply into this but it looks like [Typography is a customizable property](https://mui.com/material-ui/customization/theming/#theme-configuration-variables) of the theme.

popnfresh234 (changes requested):
Looks pretty good to me for the most part, a couple of requests for changes though:

1.  Components should go in the `Components` folder, right now the majority of your components are in a `components` folder.  This should be a quick refactor :+1: 

2.  Looks like the Bluewave repo's `App.jsx` has been overwritten, can you revert that file and update the pull request?

I think we're good to merge after that.


In general might I suggest making smaller pull requests?  It's a lot easier to review pull requests in small chunks :smile: 

Personally I think one component per pull request would be a nice level for us to work with.  If there are any issues with file strucutre or naming conventions or things like that we can catch them early while it's easy to fix.

I'm open to other ideas though if people have a different workflow they prefer!

veyselboybay:
In addition to @popnfresh234 reviews, the Tooltip folder has a .tsx file. You need to take a look at it. 
Overall, It looks good to me. I am closing this PR and after you make your changes, submit another PR and we will take a look at it and finalize the merge.

PR #31, opened 2024-05-14
Syncings 2

MuhammadKhalilzadeh:
Note: the .tsx file that @veyselboybay mention can not be changed because the built-in component of  tooltips in MUI only is compatible in a tsx file.


.vite/deps_temp_6dcee04d/package.json
```diff
@@ -0,0 +1,3 @@
+{
+  "type": "module"
+}
```
popnfresh234, line 3: I don't think the `.vite` folder should be committed to git.  I'm not even sure why it's here :thinking:

Client/src/components/Tooltips/TooltipWithTail/TooltipWithTail.tsx
```diff
@@ -0,0 +1,54 @@
+import "./tooltipWithTail.css";
+import React from "react";
+import { styled } from "@mui/material/styles";
+import Tooltip, { TooltipProps, tooltipClasses } from "@mui/material/Tooltip";
+import HelpOutlineIcon from "@mui/icons-material/HelpOutline";
+import { useTheme } from "@mui/material";
+
+const CustomizedTooltip = styled(
+  ({ className, placement, arrow, ...props }: TooltipProps) => (
```
popnfresh234, line 9: I believe all you need to do to convert this from Typescript to Javascript is remove this `TooltipProps` type.

PR #32, opened 2024-05-14
Updated gitignore to ignore log files, Refactored Authcontroller to u…

popnfresh234:
I refactored the authController so that it talks to the database through the db interface, let me know what you think.

I also added some try/catch blocks as Joi was throwing an exception that wasn't being caught and was causing the server to crash.  I also added the username field to Joi validation, it appears that all fields must be specified even if optional or it throws an error.

Let me know if any changes required or if I've broken anything!

### DIFF
--- a/Server/.gitignore
+++ b/Server/.gitignore
@@ -1,2 +1,3 @@
 node_modules
 .env
+*.log
--- a/Server/controllers/authController.js
+++ b/Server/controllers/authController.js
@@ -1,32 +1,43 @@
-const express = require('express')
-const UserModel = require('../models/user')
-const { authValidation } = require('../validation/joi')
+const express = require("express");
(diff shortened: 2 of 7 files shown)


veyselboybay:
Overall, It looks good to me. Thank you for your changes and refactorings. We can add logger documentations inside controllers.

popnfresh234:
> Overall, It looks good to me. Thank you for your changes and refactorings. We can add logger documentations inside controllers.

Sounds good to me, I forgot all about the logger 😅

veyselboybay:
> > Overall, It looks good to me. Thank you for your changes and refactorings. We can add logger documentations inside controllers.
> 
> Sounds good to me, I forgot all about the logger 😅

It's all good. 👍

PR #33, opened 2024-05-15
update on joi schema, added logger docs, and updated user model

veyselboybay:
- Updated the user model, added firstname and lastname fields,
- Updated joi schema, split the authValidation into loginValidation and registerValidation,
- Added logger docs,
- Updated MongoDB file, deselected password when returning user.


popnfresh234 (approved):
Looks great, thanks for adding the logging. 

Removing password from the user queries is a good call too :+1:

PR #36, opened 2024-05-15
Added login route, updated readme

popnfresh234:
Added a login route and some minor refactoring in DB and auth controller

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,6 +6,11 @@ BlueWave uptime monitoring application
 
 - Clone this repository to your local machine
 
+1.  [Installation (Client)](#client)
+1.  [Installation (Server)](#server)
+1.  [Configuration(Server)](#config-server)
+1.  [Endpoints](#endpoints)
+
 ---
 
 ### Client
@@ -30,7 +35,7 @@ BlueWave uptime monitoring application
 
 ---
 
-#### Configuration
+#### Configuration {#config-server}
 
 Configure the server with the following environmental variables
 
--- a/README.md
+++ b/README.md
@@ -34,10 +34,10 @@ BlueWave uptime monitoring application
 
 Configure the server with the following environmental variables
 
-| ENV Varialbe Name    | Required/Optional | Type   | Description                       |
-| -------------------- | ----------------- | ------ | --------------------------------- |
-| DB_CONNECTION_STRING | Required          | string | Specfies URL for MongoDB Database |
-| MAILERSEND_API_KEY   | Required          | string | Specfies API KEY for MailerSend service |
+| ENV Varialbe Name    | Required/Optional | Type   | Description                                          |
+| -------------------- | ----------------- | ------ | ---------------------------------------------------- |
+| DB_CONNECTION_STRING | Required          | string | Specfies URL for MongoDB Database                    |
+| MAILERSEND_API_KEY   | Required          | string | Specfies API KEY for MailerSend service              |
 | SYSTEM_EMAIL_ADDRESS | Required          | string | Specfies System email to be used in emailing service |
 
 ---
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -79,51 +98,186 @@ Example:
...
+    "__v": 0
+  }
+}
+```
+
+</details>
+
 ---
 
-##### GET /api/v1/monitors
```
Skorpios604, line 84: Just a minor adjustment to the readme as the the path should be "/api/v1/auth" rather than "/api/v1" according to the code:
```javascript
app.use("/api/v1/auth", authRouter);
```

Skorpios604 (changes requested):
I checked the login and registration and they are indeed working. They are both fully functional. I also tried registering with the same user and got the "user already exists!" message. User existence works and so does password matching. Error handling also works.

README.md
popnfresh234: Thanks for catching that, will update and commit

popnfresh234: On second glance, I checked out the readme and it appears to be correct :thinking: 

There's two routers right now, `auth` and `monitors`

```
app.use("/api/v1/auth", authRouter);
app.use("/api/v1/monitors", monitorRouter);
```

The auth routes should be

`/api/v1/auth`

and the monitor routes should be

`/api/v1/monitors/`


For example login is

`/api/v1/auth/login`

the endpoint for all monitors is

`/api/v1/monitors`

and the endpoint for a specific monitor is

`/api/v1/monitors/{monitorId}`

Can you double check and let me know?  Thanks!

Skorpios604: Yup that is correct!

Skorpios604 (approved):

MuhammadKhalilzadeh (approved):

Server/controllers/authController.js
```diff
@@ -23,28 +23,73 @@ const registerController = async (req, res) => {
...
+
+    // If user not found, throw an error
+    if (!user) {
+      throw new Error("User not found!");
+    }
+    // Compare password
+    const match = await user.comparePassword(req.body.password);
+    console.log(user);
+    if (!match) {
+      throw new Error("Password does not match!");
```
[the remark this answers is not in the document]
veyselboybay: Looks good to me. 👍

Server/db/MongoDB.js
```diff
@@ -14,18 +15,22 @@ const connect = async () => {
 
 const insertUser = async (req, res) => {
   try {
-    const newUser = await UserModel.create({ ...req.body }).select('-password');
-    return newUser;
+    const newUser = new UserModel({ ...req.body });
+    await newUser.save();
+    return await UserModel.findOne({ _id: newUser._id }).select("-password"); // .select() doesn't work with create, need to save then find
```
veyselboybay, line 20: Thanks, that's a nice catch. 👌

Server/controllers/authController.js
```diff
@@ -23,28 +23,73 @@ const registerController = async (req, res) => {
...
+    // We can move this to the DB layer if we want
+    // Probably not neccessary at all as we'll likely return a JWT instead of a user
+    const userWithoutPassword = { ...user._doc };
+    delete userWithoutPassword.password;
+
+    // Happy path, return user
+    // In the future we'll probably return a JWT instead of a user
+    return res
+      .status(200)
+      .json({ success: true, msg: "Found user", data: userWithoutPassword });
```
veyselboybay, line 87: Looks good and let's keep it for now, we can remove it when we implement JWT.

Server/models/user.js
```diff
@@ -45,7 +45,13 @@ UserSchema.pre("save", async function (next) {
...
-  this.password = bcrypt.hash(this.password, salt);
+  this.password = await bcrypt.hash(this.password, salt); // hash is also async, need to eitehr await or use hashSync
+  next();
 });
 
+UserSchema.methods.comparePassword = function (submittedPassword) {
+  console.log(submittedPassword, this.password);
+  return bcrypt.compare(submittedPassword, this.password);
+};
+
```
veyselboybay, line 56: This is good. 👌

veyselboybay (commented):
Everything looks good, we can merge it.

PR #38, opened 2024-05-15
Implemented proposed ping service, resolves #37

### LINKED ISSUE #37: Add Ping Service for Server
The server likely needs some sort of service for pinging all the URLs specified in the monitors.

We've discusse dusing [BullMQ](https://docs.bullmq.io/) to set up a queue to handle this operation

Eventually we may want to have a number of options for notifying users of up status changes (SMS, Email, websocket, etc)

popnfresh234:
Proposal for a ping service.

This is totally a prototype, very open to feedback and ideas!

### DIFF
--- a/Server/service/pingService.js
+++ b/Server/service/pingService.js
@@ -73,9 +73,10 @@ const cleanup = async () => {
 
     // Clean out the queue
     const jobs = await queue.getJobs(["active", "waiting", "delayed"]);
+    // There could be a great many jobs, handle this asynchronously with Promise.all
     await Promise.all(jobs.map((job) => job.remove()));
 
-    // Close workers
+    // Close workers.  Same as jobs, there could be many workers we don't want to wait for each one to close.
     await Promise.all(workers.map((worker) => worker.close()));
 
     // Close the queue
--- a/Server/service/pingService.js
+++ b/Server/service/pingService.js
@@ -8,9 +8,9 @@ const connection = {
 };
 
 const JOBS_PER_WORKER = 5;
+const MAX_CREATE_WORKER_RETRIES = 3;
 
 const SERVICE_NAME = "pingService";
-
 const QUEUE_NAME = "monitors";
 
 let queue;
@@ -30,10 +30,11 @@ const enqueueJob = async (queue, monitor) => {
 };
 
 const pingMonitor = (monitor) => {
+  // TODO Implement pinging the monitor
   console.log(`{url: ${monitor.url},`, `interval: ${monitor.interval}}`);
 };
 
-const createWorker = (queue) => {
+const createWorker = (queue, retries = 0) => {
   try {
     const worker = new Worker(
       QUEUE_NAME,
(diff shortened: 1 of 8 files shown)


Server/service/pingService.js
```diff
@@ -19,9 +21,9 @@ const enqueueJob = async (queue, monitor) => {
       { monitor },
       { delay: monitor.interval * 1000 }
     );
-    logger.info("Job enqueued", { service: "pingService" });
+    logger.info("Job enqueued", { service: SERVICE_NAME });
```
Skorpios604, line 35: I suppose the reason for this is because a constant is better than a hard-coded string?

Server/service/pingService.js
```diff
@@ -70,16 +70,13 @@ const startPingService = async (db) => {
...
+    await Promise.all(jobs.map((job) => job.remove()));
 
     // Close workers
-    for (const worker of workers) {
-      await worker.close();
-    }
+    await Promise.all(workers.map((worker) => worker.close()));
 
     // Close the queue
     await queue.close();
```
Skorpios604, line 116: Good for freeing up memory and CPU usage. Good use of async.

Skorpios604:
I like the over all idea in the sense that a pinging service ensures availability and continuously monitors services before they impact other users significantly which is the very basis of our uptime monitor. It also seems like a more automated and intuitive approach than other monitoring services like APM.

Server/service/pingService.js
popnfresh234: Yep, don't want to update lots of hardcoded strings if we change the name of the service one day

Server/service/pingService.js
popnfresh234: Ya my biggest concern is running out of resources since we could have a very large number of monitors we need to ping conccurently

Skorpios604 (approved):

Server/service/pingService.js
gorkem-bwl: > Ya my biggest concern is running out of resources since we could have a very large number of monitors we need to ping conccurently

I would say we don't have to think deeply about those two: 

- Multi-tenancy
- Scalability

We can always get those sorted out once there is an MVP we can play with.

popnfresh234: Agreed, we shouldn't get too bogged down in performance details at this point.  Plenty of time for that once things are up and running :laughing:

Server/service/pingService.js
```diff
@@ -50,6 +51,17 @@ const createWorker = (queue) => {
...
+    if (retries < MAX_CREATE_WORKER_RETRIES) {
+      logger.info(
+        `Retrying to create worker (${
+          retries + 1
+        }/${MAX_CREATE_WORKER_RETRIES})`,
+        { service: SERVICE_NAME }
+      );
+      return createWorker(queue, retries + 1);
+    }
+
```
veyselboybay, line 73: This is great. 👌

Server/service/pingService.js
```diff
@@ -2,6 +2,15 @@ const { Queue, Worker } = require("bullmq");
...
 
+/**
+ * NOTES
+ * This code will be resource heavy as we will be creating
+ * many workers and jobs.  Operations should be robust and have good
+ * error handling and retry functionality.
+ * There will be a lot of I/O operations as well, so we should be
+ * aware of asynchronous operations and handle them properly.
+ */
+
```
veyselboybay, line 13: 👍

Server/service/pingService.js
```diff
@@ -0,0 +1,74 @@
...
+    }
+  );
+};
+
+const startPingService = async (db) => {
+  queue = new Queue("monitors", {
+    connection,
+  });
+
+  const monitors = await db.getAllMonitors();
```
veyselboybay, line 83: @popnfresh234 Are we returning only `active` monitors here?

popnfresh234: Good point, I'm not sure if there's going to be a case where we need to return *all* montiors, maybe we only ever need to return only active monitors?

veyselboybay: I think in this case, we will only need active ones.

MuhammadKhalilzadeh (approved):

PR #49, opened 2024-05-21
Feat/monitor routes

ajhollid:
Added two new routes:

```
POST api/v1/monitors/
POST /api/v1/monitors/delete/{monitorId}
```

to create and delete monitors, respectively.

Updated `GET /api/v1/monitors` to only return active monitors

### DIFF
--- a/Server/db/MongoDB.js
+++ b/Server/db/MongoDB.js
@@ -61,7 +61,7 @@ const getUserByEmail = async (req, res) => {
  */
 const getAllMonitors = async (req, res) => {
   try {
-    const monitors = await Monitor.find();
+    const monitors = await Monitor.find({ isActive: true });
     return monitors;
   } catch (error) {
     throw error;


Server/controllers/monitorController.js
```diff
@@ -82,4 +85,68 @@ const getMonitorsByUserId = async (req, res) => {
...
+  if (error) {
+    return res
+      .status(400)
+      .json({ success: false, msg: error.details[0].message });
+  }
+  try {
+    const monitor = await req.db.deleteMonitor(req, res);
+    return res
+      .status(200)
+      .json({ success: true, msg: "Monitor deleted", data: monitor });
```
veyselboybay, line 137: A couple of things worth discussing;
1. Should we return 204 instead of 200 as a status code? 
2. Should we delete the related checks and alerts after deleting the monitor inside this controller?
3. More importantly, should we force users to only delete inactive monitors (they need to pause the service first)?

Server/db/MongoDB.js
```diff
@@ -36,17 +51,31 @@ const getUserByEmail = async (req, res) => {
...
+ * @async
+ * @param {Express.Request} req
+ * @param {Express.Response} res
+ * @returns {Promise<Array<Monitor>>}
+ * @throws {Error}
+ */
 const getAllMonitors = async (req, res) => {
   try {
-    const monitors = await Monitor.find();
+    const monitors = await Monitor.find({ isActive: true });
```
veyselboybay, line 64: 👌

veyselboybay (commented):
Overall looks good to me.

Server/controllers/monitorController.js
popnfresh234: Good points, my thoughts:

1.  Very good point, I should update the various status codes to be more semantically accurate.  200 is semantically correct but not very specific.  I'll update codes to provide more information.

2.  Yes, I don't think it makes sense to keep the checks and alerts for a monitor around once the monitor has been deleted.  I'll add comments to remind us to do that once we've implemented checks and alerts

3.  I think this can be hidden from the user and handled on the backend.  Their intention when deleting a monitor is to turn it off anyways, so we can handle stopping the service for them, otherwise it's an extra step on their end.

I'll handle 1 and 2 and update the PR, for 3 I'll add some comments to remind us to deal with it once we've got the service hammered out :+1:

veyselboybay: Awesome! 👍

PR #50, opened 2024-05-21
Implemented the integrations component.

Skorpios604:


popnfresh234 (changes requested):
There are some hardcoded dimension values that may give us some trouble down the road, I'm not sure how many of these are one off's or are values that we should be reusing throught the app.

If they are one offs no big deal, but if they are values that will be reused elsewhere we should think about refactoring them to the theme.

The one major issue I see is that we already have a `Button` component implemented, we should use that instead of coding up a new `Button` here as I believe this is just a standard button.

Skorpios604:
Refactored code and addressed the comments. Please have a look. Thanks!

Client/src/Components/Integrations/index.jsx
```diff
@@ -0,0 +1,56 @@
...
+            maxWidth: '300px',
+            wordWrap: 'break-word',
+            textAlign: 'left'
+          }}
+        >
+          This is a paragraph that provides more detail about the header.
+        </Typography>
+      </Box>
+      
+      <Button label="Click Me" level="primary" />
```
popnfresh234, line 51: Sweet, I think this is the first component that uses one of our own components in it!  Nice work :+1:

popnfresh234 (approved):
Looks good to me!

Client/src/Components/Integrations/index.jsx
Skorpios604: Ya I noticed that too! Thanks!

PR #54, opened 2024-05-22
Feat/jwt resolves #41

### LINKED ISSUE #41: Authentication and Tokens
I believe we are going to develop a stateless RESTful API, if that's the case we'll likely need some authentication strategies, at least two:

1.  Provide local login (username/password)
2. 3d party OAuth (Google, Facebook, etc)

Are we going to use a library for this like [Passport.js](https://www.passportjs.org/)?  Or roll our own?

I assume we'll issue a token once a user authenticates, will we use [JWT](https://jwt.io/) for this?  Looks like [jsonwebtoken](https://github.com/auth0/node-jsonwebtoken) is still the go to for this.

We can then write some middleware to verify the JWT on protected routes, or just use `Passport.js`.

In summary it might be helpful to:

*  Decide on authentication strategies
* Decide which libraries (if any) to use for Oauth and JWT (if we decide on that standard)

popnfresh234:
Register and Login routes now issue and return a JWT with user info encoded, the frontend will need this for display and requests that need user ID.

There is a JWT verification middleware that can be added to any route or router to protect the route/routes.  The middleware also adds the User model to the `request` object so we have access to it in the protected route.

### DIFF
--- a/Server/controllers/authController.js
+++ b/Server/controllers/authController.js
@@ -2,8 +2,17 @@ const express = require("express");
 const { registerValidation, loginValidation } = require("../validation/joi");
 const logger = require("../utils/logger");
 require("dotenv").config();
-
 var jwt = require("jsonwebtoken");
+
+/**
+ * Creates and returns JWT token with an arbitrary payload
+ * @function
+ * @param {Object} payload
+ * @returns {String}
+ */
+const issueToken = (payload) => {
+  return jwt.sign(payload, process.env.JWT_SECRET, { expiresIn: "1h" });
+};
 
 /**
  * @function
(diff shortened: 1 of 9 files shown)


Server/controllers/authController.js
```diff
@@ -69,25 +90,22 @@ const loginController = async (req, res) => {
...
     if (!match) {
       throw new Error("Password does not match!");
     }
 
-    // Copy user to remove password.  More memory, but better than mutating user?
-    // We can move this to the DB layer if we want
-    // Probably not neccessary at all as we'll likely return a JWT instead of a user
+    // Remove password from user object.  Should this be abstracted to DB layer?
     const userWithoutPassword = { ...user._doc };
     delete userWithoutPassword.password;
```
veyselboybay, line 100: This approach looks fine right now. Let's just keep it that way, later we can refactor it if we want.

Server/middleware/verifyJWT.js
veyselboybay, line 1: Looks neat and clear. 👌

Server/index.js
```diff
@@ -58,8 +60,7 @@ app.use((req, res, next) => {
 
 //routes
 app.use("/api/v1/auth", authRouter);
-
-app.use("/api/v1/monitors", monitorRouter);
+app.use("/api/v1/monitors", verifyJWT, monitorRouter);
```
veyselboybay, line 63: Nice to see token check in place 👍

veyselboybay (commented):
Overall looks good to me.

PR #60, opened 2024-05-24
Implemented the navbar dropdown section for the user.

Skorpios604:


Client/src/Components/NavBar/index.jsx
```diff
@@ -1,10 +1,48 @@
...
-import './index.css'
+import { useState } from 'react';
+import './index.css';
+import Button from '@mui/material/Button';
+import Menu from '@mui/material/Menu';
+import MenuItem from '@mui/material/MenuItem';
 
 const NavBar = () => {
+    const [anchorEl, setAnchorEl] = useState(null);
+    const open = Boolean(anchorEl);
```
popnfresh234, line 9: To me this is a little bit unclear, it seems we're using the side effect of setting the anchor element to obtain a truthy/falsey value for the `open` variable.

It might be clearer to handle the open variable separately using state, as it is representing the state of the NavBar.

I'm always wary of using side effects of one operation to handle another, it can become hard to follow at best and unpredictable at worst. 

Not a huge deal, but might be worth thinking about. JavaScript and it's truthy/falsey way of evaluation can create problems that are very hard to track down!

popnfresh234: Disregard, MUI Docs approve

popnfresh234 (approved):

PR #61, opened 2024-05-24
Register page

MuhammadKhalilzadeh:
the UI of `register page` is completed and also some issues that was mention in our meeting about the `login page`.
**Next**, I'll be doing some **polishing** and trying to **minimize the styling (CSS) code as much as possible**. I'll try to reach the **shortest version of global styling values** from what I've coded so far and then add it to the `Theme`.


popnfresh234 (approved):
Awesome, thanks for the docs!

PR #66, opened 2024-05-25
Refactored the Navbar with an avatar and a full bar.

Skorpios604:
This is closer to match the overall sought after design

### DIFF
--- a/Client/src/Components/NavBar/index.jsx
+++ b/Client/src/Components/NavBar/index.jsx
@@ -1,48 +1,138 @@
 import { useState } from 'react';
-import './index.css';
-import Button from '@mui/material/Button';
+import AppBar from '@mui/material/AppBar';
+import Box from '@mui/material/Box';
+import Toolbar from '@mui/material/Toolbar';
+import IconButton from '@mui/material/IconButton';
+import Typography from '@mui/material/Typography';
 import Menu from '@mui/material/Menu';
+import MenuIcon from '@mui/icons-material/Menu';
+import Container from '@mui/material/Container';
+import Avatar from '@mui/material/Avatar';
+import Tooltip from '@mui/material/Tooltip';
 import MenuItem from '@mui/material/MenuItem';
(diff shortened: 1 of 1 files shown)


Client/src/Components/NavBar/index.jsx
```diff
@@ -1,48 +1,138 @@
...
-
-    return (
-        <div>
-            <Button
-                id="basic-button"
-                aria-controls={open ? 'basic-menu' : undefined}
-                aria-haspopup="true"
-                aria-expanded={open ? 'true' : undefined}
-                onClick={handleClick}
+const settings = ['Profile', 'Team', 'Invite Colleagues', 'Logout'];
```
popnfresh234, line 15: Would these be passed in as props?  Not sure how often this component will be used, but if it is going to be used more than once I imagine they would be :thinking:

Client/src/Components/NavBar/index.jsx
```diff
@@ -1,48 +1,138 @@
...
             >
-                <MenuItem onClick={handleClose}>Profile</MenuItem>
-                <MenuItem onClick={handleClose}>Team</MenuItem>
-                <MenuItem onClick={handleClose}>Invite Colleagues</MenuItem>
-                <MenuItem onClick={handleClose}>Log Out</MenuItem>
+              {settings.map((setting) => (
+                <MenuItem key={setting} onClick={handleCloseUserMenu}>
+                  <Typography textAlign="center">{setting}</Typography>
+                </MenuItem>
+              ))}
```
popnfresh234, line 129: Nice, this is a good refactor

popnfresh234 (approved):
Nice refactor!

I'm not sure whether it makes sense to pass the items for the menu in as Props, but if the menu items are going to be defined in a higher level component then it probably does.

I'm sure that issue will make itself known if and when the time is ripe :laughing:

Skorpios604:
Going to be passing in as props for now.

Issue #62, opened 2024-05-24
Add Endpoints to TOC in Documentation


popnfresh234:
Each endpoint should have a link in the table of contents.

Example:

1.  [Installation (Client)](https://github.com/bluewave-labs/bluewave-uptime#client)
2.  [Installation (Server)](https://github.com/bluewave-labs/bluewave-uptime#server)
3.  [Configuration(Server)](https://github.com/bluewave-labs/bluewave-uptime#config-server)
4.  [Endpoints](https://github.com/bluewave-labs/bluewave-uptime#endpoints)
    * [/api/v1/auth/register]()
    * [/api/v1/auth/login]()

PR #67, opened 2024-05-25
Add Server endpoints to ToC in README

alvarotorrestx:
I went ahead and linked server endpoints in the README in the layout that was shown in issue #62 

Let me know if this would suffice or should we make further updates?

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,6 +10,14 @@ BlueWave uptime monitoring application
 1.  [Installation (Server)](#server)
 1.  [Configuration(Server)](#config-server)
 1.  [Endpoints](#endpoints)
+	- <code>POST</code> [/api/v1/auth/register](#post-register)
+	- <code>POST</code> [/api/v1/auth/login](#post-login)
+	- <code>GET</code> [/api/v1/monitors](#get-monitors)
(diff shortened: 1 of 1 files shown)


alvarotorrestx:
Maybe add the front end routes? Or that can probably wait until further is completed with server

popnfresh234 (approved):
Fantastic, thanks for taking care of this, it's a small task but it's a load off my shoulders!

Documentation can be boring but it really makes a project stand out.  Thanks your for help!

popnfresh234:
> Maybe add the front end routes? Or that can probably wait until further is completed with server

Good point, but I don't think we've settled on routes for the front end yet, so it might be a bit premature to get documentation going there.

We'll open another issue when the time is right!

popnfresh234:
PS  If you add something like "resolves #62" Github will automatically close that issuse when the pull request is accepted :+1:

alvarotorrestx:
For sure, happy to start off with some sort of contribution! As things progress and I learn more about your project I'd love to see where else I can contribute too! 

I appreciate the tips also, I'll know that for future PRs 👌

Issue #53, opened 2024-05-22
Style Consistency


popnfresh234:
As the number of components grows it's probably worth thinking about how to maintain consistency across our library of components.

I suggest going over the frontend components before we get too far along and refactor out any common dimensions to the theme, or at least a root level CSS file.

I'm mostly thinking of values like:

* margin and padding
* font sizes
* widths/heights

I'm sure there's going to be exceptions and one off's, but I imagine there are also a large number of values that are going to be shared between many components

The sooner we get to this the less painful it will be :laughing:

alvarotorrestx:
Hey there, can I help contribute to your project? If so, please assign me 😄

Skorpios604:
Ya your right this is important. It already sounds painful😄

alvarotorrestx:
Hi, I'm still willing to contribute. I'll check this out late today, thanks! 😊

Skorpios604:
welcome @alvarotorrestx

alvarotorrestx:
Hey everyone, thanks for your patience on allowing me to get back with you all. I've been trying to skim through everything and learn about your app while I work. I really like the code layout by the way, it's almost like we learned the same way and very similar to how I organize/code. 😅

So to my understanding, are we wanting to factor out repeating styles such as the mentioned:
- margin and padding
- font sizes
- widths/heights

And either come out with a master css file or work with the theme for reusability? 

And this is starting off with the components in the Client front end folder?
![image](https://github.com/bluewave-labs/bluewave-uptime/assets/79110517/4fb171ee-9a54-4e06-8b71-81838b7d3ec5)

Of course like mentioned, I know there will be some usage where the components will need to be manually updated on their style which makes sense. I'm just trying to get as much accurate information as I can before properly making changes/submitting PRs.

Further Notes/Questions:
- Do you all have a mockup or wireframe of how the outcome of the project?
I was able to check out the demo most components being displayed.

I have more questions on the project, but those will come over time so I don't overwhelm ya! Thanks so much! 😊

Skorpios604:
Ya so the idea is that we have one consistent style by using MUI theme component which will customize the appearance of our application by overriding the default styles and creating a consistent look and feel. This is also good for maintainability because if values change down the line we won't have to go and change each and every one of them individually. 

Yes that is correct! So the manual ones we made an index.css file for them. 

Ya if you have anymore questions please ask.

alvarotorrestx:
[...]

PR #76, opened 2024-05-28
Created :root variables for common component styles

alvarotorrestx:
Adds a resolution for #53 but not resolving the entire issue.

I wanted your thoughts on this, I was able to find consistent usage of the variables that I created and went ahead and added them to the index.css :root portion so that it can be used throughout all components.

I saw a lot more consistent styles after skimming through more and correcting some of the updated changes, I feel that it could be updated similarly. 

Although, this could eventually build up a lot. I wanted your thoughts on the naming convention, if it'll be easy enough as is for usage or should it be simpler? More explanatory?

Furthermore, if you all like this concept, after updating more/all I could add to the README or a separate page such as the demo to share all the common styles for ease of access.

Please let me know your thoughts, thanks.

### DIFF
--- a/Client/src/index.css
+++ b/Client/src/index.css
@@ -11,6 +11,27 @@
   text-rendering: optimizeLegibility;
   -webkit-font-smoothing: antialiased;
   -moz-osx-font-smoothing: grayscale;
+
+  --size-40: 40px;
+  --size-38: 38px;
+  --size-35: 35px;
+  --size-28: 28px;
+  --size-24: 24px;
+  --size-20: 20px;
+  --size-16: 16px;
+  --size-13: 13px;
+
+  --font-size-1: 13px;
+  --font-size-2: 1rem;
+  --font-size-3: 16px;
+  --font-size-4: 11px;
+
+  --border-2-solid: 2px solid;
+
+  --margin-1: 10px;
+  --margin-2: 20px;
+
+  --cursor-pointer: pointer;
 }
 
 a {
@@ -44,7 +65,7 @@ button {
   font-weight: 500;
   font-family: inherit;
   background-color: #1a1a1a;
-  cursor: pointer;
+  cursor: var(--cursor-pointer);
   transition: border-color 0.25s;
 }
 
--- a/Client/src/Components/Announcements/AnnouncementUpdateSubscription/announcementUpdateSubscription.css
+++ b/Client/src/Components/Announcements/AnnouncementUpdateSubscription/announcementUpdateSubscription.css
@@ -31,7 +31,7 @@
 }
 
 .subscription-dialog-title {
-  font-size: 1rem;
+  font-size: var(--font-size-2);
   font-weight: 700;
   color: var(--update-subscription-color-0);
 }
@@ -45,7 +45,7 @@
 }
 
 .subscription-dialog-text {
-  font-size: 13px;
+  font-size: var(--font-size-1);
   color: var(--update-subscription-color-1);
 }
 
@@ -77,7 +77,7 @@
   color: var(--update-subscription-color-6);
   padding: 0 var(--spacing-2);
   font-weight: 700;
-  font-size: 13px;
+  font-size: var(--font-size-1);
   margin-bottom: 10px;
 }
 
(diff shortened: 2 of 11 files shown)


alvarotorrestx:
I wanted to ask for review on this but it's not popping up for me for some reason, sorry about that.
@popnfresh234 @Skorpios604

Client/src/index.css
```diff
@@ -11,6 +11,27 @@
...
   -moz-osx-font-smoothing: grayscale;
+
+  --size-40: 40px;
+  --size-38: 38px;
+  --size-35: 35px;
+  --size-28: 28px;
+  --size-24: 24px;
+  --size-20: 20px;
+  --size-16: 16px;
+  --size-13: 13px;
```
popnfresh234, line 22: Overall I think this is generally what we had in mind, but if we name the vars in a way that reflects their values we kind of defeat the purpose refacotring the values.

The idea behind refactoring out the values is to give us the flexibility to change them all in one place if we want to later.  If for example we want to change the `13px` value to `16px` the var will now be

`--size-13: 16px`

which works, but semantically no longer makes sense.  I think it would be best to name these something like `--size-small` `--size-extra-small` etc.

I'll leave it up to the front end guys to chime in on this @Skorpios604  @MuhammadKhalilzadeh

Client/src/index.css
```diff
@@ -11,6 +11,27 @@
...
+
+  --font-size-1: 13px;
+  --font-size-2: 1rem;
+  --font-size-3: 16px;
+  --font-size-4: 11px;
```
popnfresh234, line 27: This is great :+1:

Client/src/index.css
```diff
@@ -11,6 +11,27 @@
...
+  --size-38: 38px;
+  --size-35: 35px;
+  --size-28: 28px;
+  --size-24: 24px;
+  --size-20: 20px;
+  --size-16: 16px;
+  --size-13: 13px;
+
+  --font-size-1: 13px;
+  --font-size-2: 1rem;
```
popnfresh234, line 25: Not a fan of mixed units, this is on us though, not on your refactoring :joy:  We should fix this at some point

popnfresh234 (changes requested):
Overall it looks really good to me, I think generally it's what we're looking for.

I have some opinions about the variable names, I've left my comments but I'll defer to the front end team. 

Thanks for your hard work!

Client/src/index.css
MuhammadKhalilzadeh: This is just a temporary naming. My main goal is to circularly give my stylings general names (Like what you mentioned) and add my styles to the Theme after completing the design of every couple of pages.
I just first want to see how frequently each value gets repeated and what's the range it can have and what name could best describe it.

popnfresh234: @MuhammadKhalilzadeh I think these vars were created by @alvarotorrestx ?

In any event I'll leave this one up to you guys to decide on :+1:

alvarotorrestx: I agree on the names being more general like @popnfresh234 stated with --size-small, etc. Although, I had created them with multiple names since there were many different, yet repeated, styles. I'm sure like @MuhammadKhalilzadeh said, once there's consistent styles this will come with ease to rename and readjust the values.

popnfresh234: @alvarotorrestx I'm OK with that :+1: Once our team lead @veyselboybay has a chance to review I think we can probably merge this in

Client/src/index.css
alvarotorrestx: Same here lol, I prefer rem/em values but I agree, I feel this will be updated shortly. I'll await further review

Client/src/index.css
Skorpios604: I think the overall idea of the naming convention that @alvarotorrestx came up with is good practice because of reusability, maintainability, and it will also be good for our theming. I also like the readability of it. I also think it is important to consider what @popnfresh234 said with regard to the semantics of the naming convention.

ajhollid:
We'll revisit this one in the future

PR #77, opened 2024-05-28
Implemented the dashboard settings icon with its dropdown menu.

Skorpios604:


Client/src/Layouts/HomeLayout/index.jsx
```diff
@@ -3,6 +3,7 @@ import { Outlet } from 'react-router'
 import Home from '../../Pages/Home'
 import './index.css'
 import ChartsOverviewDemo from '../../Components/DashboardProgressBars'
+import DashboardSettings from '../../Components/DashboardSettings'
```
popnfresh234, line 6: I think this was meant to be 

`DashboardSettingsIcon`

popnfresh234 (dismissed):
Currently won't compile due to an import error.  I think that `DashboardSettings` was meant to be `DashboardSettingsIcon` in the home layout.

PR #84, opened 2024-05-29
Feat/check db

popnfresh234:
Added DB operations for `Checks`

These operations will probably be called by internal service, but there are API endpoints for them as well for now.  This is helpful for testing and give us the possibiltiy of extracting the Ping Service to an external server at some point if we wish.

### DIFF
--- a/Server/db/MongoDB.js
+++ b/Server/db/MongoDB.js
@@ -250,7 +250,7 @@ const getChecks = async (monitorId) => {
  * Delete all checks for a monitor
  * @async
  * @param {string} monitorId
- * @returns {boolean}
+ * @returns {number}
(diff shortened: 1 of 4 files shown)


Server/db/MongoDB.js
```diff
@@ -247,19 +249,19 @@ const getChecks = async (req, res) => {
 /**
  * Delete all checks for a monitor
  * @async
- * @param {Express.Request} req
- * @param {Express.Response} res
+ * @param {string} monitorId
+ * @returns {boolean}
```
veyselboybay, line 253: I think this operation returns Number rather than boolean?

veyselboybay (approved):

PR #88, opened 2024-05-30
Navbar now takes up the full viewport width.

Skorpios604:


veyselboybay (approved):
Looks good, thanks!

Client/src/Components/NavBar/index.jsx
```diff
@@ -46,8 +46,8 @@ function NavBar() {
   };
 
   return (
-    <AppBar position="static">
-      <Container maxWidth="xl">
+    <AppBar position="fixed" sx={{ width: '100%' }}>
+      <Container maxWidth="xl" sx={{ width: '100%' }}>
```
popnfresh234, line 50: does maxWidth and width 100% have any conflict here?  I assume ecplicitly stating the width overrides whatever maxWidth is set to?

If that's the case we can probably just drop the maxWidth prop all together?  That would be less confusing for future developers

PR #92, opened 2024-06-01
Implemented the sidebar component

Skorpios604:
Implemented the sidebar component and then added it to the home screen layout.


popnfresh234 (approved):
Looks good to me :+1:

PR #99, opened 2024-06-05
Implemented the emailing service

veyselboybay:
Changed the emailing service to `SendGrid` from `Twilio`. Tested the service.

### DIFF
--- a/Server/utils/sendEmail.js
+++ b/Server/utils/sendEmail.js
@@ -1,43 +1,38 @@
-const { MailerSend, EmailParams, Sender, Recipient } = require('mailersend')
-const logger = require('../utils/logger')
-
-
-const mailersend = new MailerSend({
-    apiKey: process.env.MAILERSEND_API_KEY,
-})
+const sgMail = require('@sendgrid/mail')
+const logger = require('./logger')
+const dotenv = require('dotenv')
+dotenv.config({path:__dirname+'/../.env'})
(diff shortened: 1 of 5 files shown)


Server/utils/sendEmail.js
```diff
@@ -1,43 +1,38 @@
...
-const logger = require('../utils/logger')
-
-
-const mailersend = new MailerSend({
-    apiKey: process.env.MAILERSEND_API_KEY,
-})
+const sgMail = require('@sendgrid/mail')
+const logger = require('./logger')
+const dotenv = require('dotenv')
+dotenv.config({path:__dirname+'/../.env'})
```
popnfresh234, line 4: I think this path is specifying .env should be in the root project folder?  If that's the case we should make a note in the readme to tell end user's where to put it.

I've been using the Server root folder so no config is necessary there.  Might be good to keep the server config and client config separate?

popnfresh234 (approved):
Looks good to me :+1:

Server/utils/sendEmail.js
veyselboybay: I ran into an issue that it wasn't reading the .env file. I just figured I was testing it as a standalone, not inside the server. I will make the changes.

popnfresh234 (approved):

popnfresh234:
Looks good to merge!

PR #100, opened 2024-06-05
Feat/job queue

ajhollid:
Added an improved Job Queue that handles scaling workers based on worker load

Inlcludes a controller so you can call endpoints to test the queue, to be removed later

Test endpoints:

`GET /api/v1/job` - gets all jobs in the queue
`POST /api/v1/job` - creates a job and adds it to the queue
`POST /api/v1/job/obliterate` - obliterates the queue

Currently the Queue simply adds test jobs, will integrate adding `Monitors` to the queue once we've settled on the queue implementation


veyselboybay (approved):
Looks good to me!

PR #103, opened 2024-06-06
Re navbar side menu

MuhammadKhalilzadeh:
Navbar and side menu are fixed and redesigned, although there are some parts of it that needs to be completed as we move forward
Screenshot (81)


veyselboybay (approved):

MuhammadKhalilzadeh:
I just changed the font for drop down menu on navbar and also managed to remove the animation effect with talk about for the menu toggle, but couldn't find a way to remove it for menu items too.

Client/src/Components/NavBar/index.css
```diff
@@ -1 +1,29 @@
...
+.icon-button-toggle-pic {
+  width: 10px;
+  height: 5px;
+}
+
+#icon-button {
+  color: transparent;
+  -webkit-transition: none;
+  transition: none;
+}
```
Skorpios604, line 29: Hey! If you want to get rid of the black outline that pops up when you click on the user dropdown menu you can use this here:
```CSS
outline: none; 
box-shadow: none;

Skorpios604: You can also create a new CSS class like this to get rid of the greyish hue that pops up when you hover over the user dropdown. 
```CSS
#icon-button:hover {
  background-color: transparent; 
}

Skorpios604 (changes requested):

Client/src/Components/NavBar/index.css
MuhammadKhalilzadeh: The first part worked, but the second part did not make a different and still effects on drop down menu do not disappear. By the way I think it's pretty much what we wanted on the UI.
Next, I'll be adding routing and navigation on this pr.

MuhammadKhalilzadeh: I just added navigation for side menu using useNavigate and now pages render one the right side

veyselboybay (approved):
Overall looks good to me. There are 2 styling affect issues, we can come back for fix later. Moving forward.

ajhollid (approved):

PR #111, opened 2024-06-07
Encoding text field inputs from users into JSON objects. 

Skorpios604:


Client/src/Pages/Login/index.jsx
```diff
@@ -25,15 +43,17 @@ const Login = () => {
...
             id="login-email-input"
           />
           <div className="login-form-v2-spacing" />
           <PasswordTextField
+            onChange={handleInput}
             error={false}
             placeholder="Password"
-            id="login-email-input"
+            id="login-password-input"
           />
```
ajhollid, line 57: Good catch

ajhollid (approved):
Looks good on my end, input is captured and stored in state.  It should be ready to fire off to the backend now

Client/src/Pages/Login/index.jsx
Skorpios604: Thanks haha!

PR #114, opened 2024-06-09
Anding routing fixing bugs

MuhammadKhalilzadeh:


Client/src/Components/DashboardMenu/index.jsx
```diff
@@ -6,6 +6,8 @@ import Incidents from "../../assets/Images/Icon-warning-gray.png";
 import SensorsIcon from "../../assets/Images/Icon-signal-gray.png";
 import AllInclusiveIcon from "../../assets/Images/Icon-link-gray.png";
 import SettingsIcon from "../../assets/Images/Icon-setting-gray.png";
+import React, { useState, useEffect } from "react";
+import { useNavigate, useLocation } from "react-router-dom";
```
ajhollid, line 11: Do we want to be using redux to handle the navigation state as well?  @veyselboybay

Client/src/Components/DashboardMenu/index.jsx
```diff
@@ -16,13 +18,56 @@ import SettingsIcon from "../../assets/Images/Icon-setting-gray.png";
...
+        "/status",
+        activeButton === "/status",
+        () => handleClick("Status Pages", "/status")
+      )}
+      {DashboardMenuButton(
+        AllInclusiveIcon,
+        "Integrations",
+        "/integrations",
+        activeButton === "/integrations",
+        () => handleClick("Integrations", "/integrations")
```
ajhollid, line 62: Rather than passing the path and the title, what about just passing the path and having a map for the title:

```
const pathMap = {
    "/integrations" : "Integrations",
    "/settings": "Settings",
    ...
}
``` 

You can then lookup the titles easily in the `handleClick` function:

```
const title = pathMap[path]
```

Then if we ever want to change the titles we just have to change the lookup map rather than going through each button.

What do you think @MuhammadKhalilzadeh ?

veyselboybay:
When I ran the code, the following error was thrown,
`Uncaught TypeError: styled_default is not a function`
The problem is solved when I added MUI dep to vite.config.js file. The final vite.config.js looks like this;
`export default defineConfig({
  plugins: [react()],
  optimizeDeps: {
    include: ['@mui/material/Tooltip', '@emotion/styled'],
  },
})`

Could you also change the vite.config.js file and update your PR?

Client/src/Components/DashboardMenu/index.jsx
veyselboybay: It'd be nice to have a central navigation state however, let's skip it for now. We can come back to it later.

Client/src/Components/DashboardMenu/index.jsx
MuhammadKhalilzadeh: That's a great idea, where do you want be to place the routing map?

MuhammadKhalilzadeh:
@veyselboybay Sure, I'll add it. I had the same problem and was fixed by running `npm install`

veyselboybay:
> @veyselboybay Sure, I'll add it. I had the same problem and was fixed by running `npm install`

I tried that as well, but It didn't solve the issue on my end. Is it just me or do you guys have the same issue @ajhollid @Skorpios604 ?

Client/src/Components/DashboardMenu/index.jsx
MuhammadKhalilzadeh: I'll probably add it in the same file or maybe create a new file for it in `utils` folder

ajhollid: I think it's fine in the same file as it should only be used here I believe

veyselboybay:
> It's running fine for me 🤔 I have had to clear out `node_modules` a few times for NPM issues though, so maybe give that a try?

I did try to clear out the `node_modules`, but the issue seemed to stay. @MuhammadKhalilzadeh already added the `optimizeDeps`, everything should be good for all of us now.

ajhollid (approved):

PR #117, opened 2024-06-10
Added individual field validation for register page

ajhollid:
Individual field validation was missing from registration page, not sure what happened to it.

Fields will now update their error state as information is entered


veyselboybay (approved):
Looks good to me. 👍

PR #127, opened 2024-06-11
Implemented the toastify notifications.

Skorpios604:


Client/src/Components/Notifications/index.jsx
```diff
@@ -0,0 +1,10 @@
+import { ToastContainer, toast } from 'react-toastify';
+import 'react-toastify/dist/ReactToastify.css';
+
+const Notification = () => {
+  return <ToastContainer />;
+};
+
+export const notify = (message) => toast(message);
```
ajhollid, line 8: ESlint doesn't like exporting both a component and a function from the same file.  Apparently it will disable fast refresh for the component.  I don't know what fast refresh is or how it works, but it seems a simple fix to avoid disabling it, so we may as well.

Recommend moving this to a separate JS file, maybe `notify.js` to resovle the warning.

```
import { toast } from "react-toastify";
export default (message) => toast(message);
```

should do it.

ajhollid (changes requested):
Cool component, I like the toasts.  

Small change RE ESlint warning requested, but other than that looks fine to me

Client/src/Components/Notifications/index.jsx
Skorpios604: I addressed this issue.

Skorpios604:
If you go to the playground route and you click on the show toast button in the bottom right you will be able to see an example of what the toast notification will look like. I made it to look like the error that would pop up that @MuhammadKhalilzadeh implemented. 

image

The dismiss and X button also have functionality.

Client/src/Components/Toast/index.jsx
```diff
@@ -1,63 +1,30 @@
...
+const ToastComponent = () => {
+  const displayMsg = () => {
+    toast(
+      <AnnouncementsDualButtonWithIcon
+        icon={<ComplexAlert theme="red" />}
+        subject="There was a problem with that action"
+        body="Lorem ipsum dolor sit amet consectetur adipisicing elit. Aliquid pariatur, ipsum dolor."
+        esc="Dismiss"
+        primary="Learn more"
+        dismiss={() => toast.dismiss()}
```
ajhollid, line 16: Dismissing toasts doesn't seem to work at the moment, am I clicking on the wrong thing maybe? :thinking:

Client/src/Components/Toast/index.jsx
```diff
@@ -1,63 +1,30 @@
...
-        clearInterval(interval);
-      };
-    }
-  }, [showToast]);
-
-  const toggleToast = () => {
-    setShowToast(!showToast);
-    setProgress(100); // Reset progress when showing toast
+const ToastComponent = () => {
+  const displayMsg = () => {
```
ajhollid, line 8: As I recall from the docs this needs to take `closeToast` and `toastProps` props to function correctly?

```
const Msg = ({ closeToast, toastProps }) => (
  <div>
     My Content
  </div>
);
```

Like that

ajhollid (changes requested):
I think the nested component needs to take some props in order to function properly, can you confirm?  Thanks!

Client/src/Components/Toast/index.jsx
Skorpios604: Ya you are right I remember that I will take a look.

Client/src/Components/Toast/index.jsx
Skorpios604: Oh no! Really? It was working for me before. Maybe I changed something the last second. Let me take a look and address this issue.

ajhollid (approved):

PR #139, opened 2024-06-14
Feat/server error msgs, resloves #138

ajhollid:


Server/utils/messages.js
Skorpios604, line 1: Like this file!

Skorpios604 (approved):
standardizing the error messages by making it centralized with constant values is a good alternative to having the hard-coded values. The error handling will be much more meaningful this way. It also enhances the logging.

PR #140, opened 2024-06-14
Implemented ping type for monitors, added cleanup for job queue, resolves #136

### LINKED ISSUE #136: Check Monitor Status via Ping
Ping should be one of the options for monitors along with http

ajhollid:
Monitors can now have a type specified, currently

* ping
* http

JobQueue will handle these two types appropriately.

Added a cleanup method for the JobQueue

### DIFF
--- a/Server/index.js
+++ b/Server/index.js
@@ -93,6 +93,16 @@ const startApp = async () => {
 
   await connectDbAndRunServer(app, db);
   const jobQueue = await JobQueue.createJobQueue(db);
+
+  const cleanup = async () => {
+    console.log("Shutting down gracefully");
+    await jobQueue.obliterate();
+    console.log("Finished cleanup");
+    process.exit(0);
+  };
+
+  process.on("SIGINT", cleanup);
+  process.on("SIGTERM", cleanup);
 };
 
 startApp().catch((error) => {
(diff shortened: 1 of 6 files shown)


Server/service/jobQueue.js
```diff
@@ -52,9 +56,25 @@ class JobQueue {
     const worker = new Worker(
```
Skorpios604, line 56: Maybe a try-catch block for this function in case the HTTP request or ping fails so that the worker doesn't crash?

Server/index.js
```diff
@@ -93,6 +93,16 @@ const startApp = async () => {
 
   await connectDbAndRunServer(app, db);
   const jobQueue = await JobQueue.createJobQueue(db);
+
+  const cleanup = async () => {
```
Skorpios604, line 97: Maybe a try-catch here and logging of errors just in case any errors occur?

Skorpios604 (approved):
I left a couple of comments for you to address if you feel the need to do so at this time. Otherwise I have approved your PR.

Server/service/jobQueue.js
ajhollid: This is a good point, I plan on refactoring out the entirety of the network operations in a soon-ish PR, so this will be much more fleshed out then.

Right now its just a proof of concept, it won't be in this state for long!

Server/index.js
ajhollid: Good call, that's worth doing!  I will add that to the next PR rather than reopen this one :+1:

PR #149, opened 2024-06-17
Added delete cascade for Montiors.  Alerts and Checks now also deleted, resolves #87

### LINKED ISSUE #87: Cascading Delete Operations
I believe Delete operation for `Monitor` should cascade to `Checks` and `Alerts`

In other words, if a user deletes a `Monitor` they created all `Checks` and `Alerts` associated with that monitor should be deleted.

ajhollid:


danielcj2:
Code looks good, but I had a question slightly unrelated. I'm unfamiliar with express, do you use post requests for updating and deleting?

ajhollid:
Express can handle all of `post`, `put`, and `delete`.  To be REST compliant we really should use `put` for updating resources and `delete` for deleting resources.

That's a good issue to open up if you want to get your feet wet on the back end :+1:  Shouldn't take much work to refactor.

Issue #141, opened 2024-06-14
Backend - Insert checks to DB on monitor status update


ajhollid:
Implement inserting checks into DB based on response from monitor.

PR #150, opened 2024-06-17
Checks now inserted into db during monitor status check, reslovles #141

ajhollid:
Insert checks into DB on monitor status update

Moved logic for cascading delete to `monitorController`.  Checks must be deleted here _after_ the queue is emptied, or there may be an incomplete job left in the queue that is carreid out after the `monitor` is deleted.  If this occurs, a dangling check without an associated monitor will be left in the database.  Now we are deleting associated checks only after the queue is drained, so this bug should be resolved.

Added JSDoc to network service


Skorpios604:
I couldn't find any issues with this code. Looks good to merge to me.

PR #156, opened 2024-06-18
Added checks to monitors for getMonitorsById and getMonitorsByUserId.…, resolves #155

### LINKED ISSUE #155: Back End - Monitors queried by UserID should also return checks
When a user's monitors are fetched, the associated checks should also be returned.

ajhollid:
* `getMonitorsByUserId` now returns `Monitor` with all its associated `Check` objects attached in ascending order (oldest first).
* `getMonitorById` also returns `Check` with the same sort order

Added a Time To Live field to the `Check` object with default time of 30 days.  Checks older than that will be automatically purged from the DB.  The duration of TTL can be discussed

### DIFF
--- a/Server/db/MongoDB.js
+++ b/Server/db/MongoDB.js
@@ -195,7 +195,9 @@ const getAllMonitors = async (req, res) => {
 const getMonitorById = async (req, res) => {
   try {
     const monitor = await Monitor.findById(req.params.monitorId);
-    return monitor;
+    const checks = await Check.find({ monitorId: monitor._id });
+    const monitorWithChecks = { ...monitor.toObject(), checks };
+    return monitorWithChecks;
   } catch (error) {
     throw error;
   }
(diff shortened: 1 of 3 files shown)


Server/db/MongoDB.js
```diff
@@ -195,7 +195,9 @@ const getAllMonitors = async (req, res) => {
 const getMonitorById = async (req, res) => {
   try {
     const monitor = await Monitor.findById(req.params.monitorId);
-    return monitor;
+    const checks = await Check.find({ monitorId: monitor._id });
+    const monitorWithChecks = { ...monitor.toObject(), checks };
+    return monitorWithChecks;
```
Skorpios604, line 200: Should we send a response to the client here?
```res.status().json()```

File: Server/db/MongoDB.js [hunk 2]
```diff
@@ -195,7 +195,9 @@ const getAllMonitors = async (req, res) => {
...
   } catch (error) {
     throw error;
```
Skorpios604, line 202: Are we going to be sending a response to the client for the error or?

Skorpios604 (approved):
Nice checks added to monitors and nice expiry date feature.

Server/db/MongoDB.js
ajhollid: Responses are sent back to thee server in the Controller, but yep that's the correct format 💯!

The DB returns its response to the controller which in turn returns to the client

Server/db/MongoDB.js
ajhollid: Errors are thrown to the controller, and then they are thrown and caught by the server middleware.

So this error will ultimately be sent to the client!

Issue #143, opened 2024-06-14
Front End - Profile Page


ajhollid:
Implement Profile & Profile Settings Page

PR #157, opened 2024-06-18
Implemented Profile settings design with some limited functionality (Client side), #143

danielcj2:
You may ignore most of the functions since they're used as placeholder at the current moment, or provide limited logic (not yet finished). I do plan to refactor a majority of the code for readability and accessibility purposes, as well as for better naming convention.


Client/src/Pages/Settings/index.jsx
```diff
@@ -1,9 +1,978 @@
...
+    description: "After updating, you'll receive a confirmation email.",
+    type: "input",
+  },
+  photo: {
+    id: "edit-photo",
+    label: "Your photo",
+    description: "This photo will be displayed in your profile page.",
+    type: "photo",
+  },
+};
```
ajhollid, line 63: I'm not sure we need a config object here, these fields are never going to change, we can probably hardcode them in?

I generally approve of keeping this as modular as possible but I think in this case the downside of the component being harder to read outweighs the benefits of keeping this config separate from the UI

Client/src/Pages/Settings/index.jsx
```diff
@@ -1,9 +1,978 @@
...
+  //(tab) 2 - Team
+  const [tab, setTab] = useState("0");
+  const handleTabChange = (event, newTab) => {
+    setTab(newTab);
+  };
+
+  const theme = useTheme();
+
+  const [isLoading, setIsLoading] = useState(false);
+  //TODO - implement delete profile picture function
```
ajhollid, line 164: Loading state should be handled by the redux store, please see login and register pages for examples.

The relevant redux code is in `Features/auth/authSlice`

Client/src/Pages/Settings/index.jsx
```diff
@@ -1,9 +1,978 @@
...
+            </Box>
+          </Stack>
+        </TabPanel>
+        <TabPanel
+          value="1"
+          sx={{ padding: "0", marginTop: theme.spacing(6.25), width: "100%" }}
+        >
+          <form className="edit-password-form" noValidate spellCheck="false">
+            <div className="edit-password-form__wrapper">
+              <AnnouncementsDualButtonWithIcon
```
ajhollid, line 439: Required prop is missing here:

```
index.jsx:439 Warning: Failed prop type: The prop `closeToast` is marked as required in `AnnouncementsDualButtonWithIcon`, but its value is `undefined`.
```

Client/src/Pages/Settings/index.jsx
```diff
@@ -1,9 +1,978 @@
...
+                  sx={{ flex: 1, marginRight: "10px" }}
+                >
+                  <Typography variant="h4" component="h1">
+                    {field.label}
+                  </Typography>
+                </Stack>
+                {/* TODO - Refactor Password Input Component */}
+                {field.type === "input" ? (
+                  <FormControl sx={{ flex: 1, minWidth: theme.spacing(30) }}>
+                    <OutlinedInput
```
ajhollid, line 459: We have components for dealing with Text and Password inputs implemented already, for consistency's sake we should use those throughout the app.

Please see the register and login components to see examples of their usage.

danielcj2:
Oh and I forgot to add JSdocs documentation and proptypes. I'll work on adding that soon ish.

ajhollid (changes requested):
The component looks spot on, nice work on the design :+1: 

I have a couple of concerns with the page:

1.  It is quite large and difficult to read - Can we refactor into a Settings page with each tab a separate component?  I think that would improve readability and maintainability

2.  We have our own components for text inputs as well as the Avatar component.  I believe you have noted that in your TODOs, but we need to use those for consistency across the app.  Please see the register and login components to see examples of dealing with errors and validation for text fields.

3.  We should use the redux store for auth state.  The auth state includes the user, so we should be using that to display user info, otherwise we will have out of sync user data in the app state and the settings page state.

4.  It would be nice to log the JSON data that is to be submitted when the `save` button is pressed, that way we can easily hook up to the backend when ready.

Client/src/Pages/Settings/index.jsx
danielcj2: Yea you're not wrong. The reason why I added the config was in case we wanted to add more fields in the future and it would be easily accessible. Will remove on my next pr.

ajhollid: Makes sense!

Client/src/Pages/Settings/index.jsx
danielcj2: At the current moment, just for testing purposes I used a single loading state for testing purposes, primarily to have the loading spinners work. Will tackle it in the future once I start tinkering with the redux store.

Client/src/Pages/Settings/index.jsx
danielcj2: I believe that error is caused by not passing a toast close function. Will likely be fixed when Monzer finishes updating the toast component.

ajhollid: Got it

Client/src/Pages/Settings/index.jsx
danielcj2: Yeah I somehow missed those, will look over it on my next pr.

danielcj2:
Refactored the three tab panels into their own files for better readability. Will change the general input fields and add error handling and validation on my next pr. Redux implementation will come likely after that.

Client/src/Components/TabPanels/ProfileSettings/ProfilePanel.jsx
```diff
@@ -0,0 +1,294 @@
...
+              label="Cancel"
+              onClick={() => setIsOpen(false)}
+              sx={{ fontSize: "13px" }}
+            />
+            <ButtonSpinner
+              level="error"
+              label="Delete account"
+              onClick={handleDeleteAccount}
+              isLoading={isLoading}
+              sx={{ fontSize: "13px" }}
```
ajhollid, line 281: There are quite a few hardcoded styles and values throughout the components/pages.

For maintainability and flexiblity we should try to define these ideally in the `Theme` if they are shared across pages, or in the speciic page's CSS file if it is sepcific to the component/page only.

Client/src/Pages/Settings/index.jsx
```diff
@@ -1,9 +1,81 @@
...
+            ))}
+          </TabList>
+        </Box>
+        <ProfilePanel />
+        <PasswordPanel />
+        <TeamPanel />
+      </TabContext>
+    </Box>
+  );
+};
```
ajhollid, line 75: Looks good!

ajhollid (approved):
This is a nice improvement, much easier to read.

There are quite a few hardcoded styles and values that can be refacotred out to either the theme or the CSS file for the page/component in question.

This is a WIP so I'm not too fussed about it, as long as it gets done eventually.

I will wait to hear from the rest of the front-end team to weigh in as this area is not my specialty, but I think this is good to merge in.

Client/src/Components/TabPanels/ProfileSettings/ProfilePanel.jsx
danielcj2: Yeah I can do that. Wasn't quite sure if we were primarily gonna use `Theme` for styling or not so I was torn between using the two.

ajhollid: I think rule of thumb is if it's being used across multiple components add it to the theme, otherwise if it's component specific it goes in the components css file

Issue #146, opened 2024-06-16
Frontend - Root Layout 


ajhollid:
It looks like the main layout is aligned using margins to offset content, it may be beneficial to refactor the layout into a grid or a flexbox.

Open for disucssion, I'm not sure if there is a particular reason it was set up that way.

PR #158, opened 2024-06-18
Implemented flexbox styling for root layout, #146

danielcj2:
Changed the navbar and sidebar to be sticky, restructured home-layout, now main-content spans the full width of the page and the content is aligned at the top of the element. No more margins. Might need to look over your individual pages and make sure the padding/margins of the components you are working on are good.

### DIFF
--- a/Client/src/Components/DashboardSidebar/index.css
+++ b/Client/src/Components/DashboardSidebar/index.css
@@ -1,19 +1,19 @@
 .dashboard-sidebar {
-  height: 100vh;
+  height: calc(100vh - 64px);
   width: 250px;
   border: 1px solid var(--color-border-0);
-  border-radius: var(--border-radius-0);
-  position: fixed;
-  top: 0;
-  left: 0;
+  border-top: none;
+  border-bottom: none;
+  top: 64px;
+  position: sticky;
+  z-index: 100;
   display: flex;
   flex-direction: column;
   justify-content: space-between;
-  padding: 20px 0;
 }
 
-.menu-container {
-  margin-top: 60px;
+.dashboard-menu-container{
+  padding-top: 40px;
 }
 
 .support-container {
--- a/Client/src/index.css
+++ b/Client/src/index.css
@@ -69,10 +69,8 @@ a:hover {
 
 body {
   margin: 0;
-  display: flex;
-  place-items: center;
   min-width: 320px;
-  min-height: 100vh;
+  height: 100%;
 }
 
 h1 {
--- a/Client/src/App.css
+++ b/Client/src/App.css
@@ -1,10 +1,3 @@
-#root {
-  max-width: 1280px;
-  margin: 0 auto;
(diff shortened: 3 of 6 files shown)


Client/src/Components/DashboardSidebar/index.css
```diff
@@ -1,19 +1,19 @@
...
   flex-direction: column;
   justify-content: space-between;
-  padding: 20px 0;
 }
 
-.menu-container {
-  margin-top: 60px;
+.dashboard-menu-container{
+  padding-top: 40px;
 }
```
ajhollid, line 17: Did this change from 60px to 40px because of the removal of the padding?

Client/src/index.css
```diff
@@ -69,10 +69,8 @@ a:hover {
 
 body {
   margin: 0;
-  display: flex;
-  place-items: center;
   min-width: 320px;
-  min-height: 100vh;
+  height: 100%;
```
ajhollid, line 73: Does this behave correclty if there is not enough content to fill the screen?

I assume the `min-height: 100vh` was set to make sure the body always at least fills the screen, even if there is no content.

Client/src/Components/DashboardSidebar/index.css
```diff
@@ -1,19 +1,19 @@
 .dashboard-sidebar {
-  height: 100vh;
+  height: calc(100vh - 64px);
   width: 250px;
   border: 1px solid var(--color-border-0);
-  border-radius: var(--border-radius-0);
```
ajhollid, line 5: What was this border radius for?

ajhollid (commented):
I don't see any specific issues and everything seems to display properly.

A couple of changes I have some questions about, but more out of curiousity than concern.

@Skorpios604 @MuhammadKhalilzadeh please chime in and if you two are OK with the changes I think we can merge this in.

Client/src/Components/DashboardSidebar/index.css
danielcj2: No idea tbh, it wasn't visible before the changes as the navbar was overlapped on top of it, so I decided to remove it.

Client/src/index.css
danielcj2: Well it's more of a redundant line. The sidebar is set to `height: calc(100vh - 64px)`  64 pixels being the height of the navBar (will need to add a media query to handle navBar height changes) and the main-content which wraps the sidebar has a height of `max-content` so automatically the main-content div will fill the remainder of the screen + overflow if there is any.

Client/src/Components/DashboardSidebar/index.css
danielcj2: Correct. Although it is more of an arbitrary value I picked. After checking the figma design, it looks like that should be 20 pixels instead.

MuhammadKhalilzadeh (approved):

PR #167, opened 2024-06-20
Added and configured redux-persist, resolves #164

### LINKED ISSUE #164: Front End - Persist Redux State
Redux state is lost on direct navigation or on refresh.

[Redux-Persist](https://github.com/rt2zz/redux-persist) 

looks like a good option

ajhollid:
Added redux-persist to handle persisting redux state.

I'd like to open this one up for disucssion, I've never used `redux-persist` before but it looks like it will cover our needs.

If anyone has opinions or suggestions around this please let everyone know.

### DIFF
--- a/Client/package-lock.json
+++ b/Client/package-lock.json
@@ -27,7 +27,8 @@
         "react-redux": "9.1.2",
         "react-router": "^6.23.0",
         "react-router-dom": "^6.23.1",
-        "react-toastify": "^10.0.5"
+        "react-toastify": "^10.0.5",
+        "redux-persist": "6.0.0"
       },
       "devDependencies": {
         "@types/react": "^18.2.66",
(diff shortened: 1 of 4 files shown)


danielcj2:
It seems like a viable option for state persistence. Just worried about how it may impact the performance with larger state sizes.

ajhollid:
> It seems like a viable option for state persistence. Just worried about how it may impact the performance with larger state sizes.

Agreed, I'm not familiar enough with it to make a strong call there.

The other option that comes to mind is that we could use local storage to save the minimum state necessary, I beleive that would be the auth state.

If we wanted to really go barebones we could just save the JWT from the auth state.

Thoughts?

danielcj2:
Either or works honestly. Heck we could also just use redux-persist strictly for essential states that make the app function or enhance it. Stuff like auth state, form states, temporary UI states (e.g. open tab panels, alerts, modals maybe?, perhaps if we make a collapsible sidebar?, and whatever may be),

ajhollid:
> Either or works honestly. Heck we could also just use redux-persist strictly for essential states that make the app function or enhance it. Stuff like auth state, form states, temporary UI states (e.g. open tab panels, alerts, modals maybe?, perhaps if we make a collapsible sidebar?, and whatever may be),

Alright, I think we can give this a shot, we can whitelist the states we want to persist (or blacklist those what we don't want), that should resolve performance issues.

BIg caveat here though is that this library appears to be unmaintained, last commit was 3 years ago.  Redux-toolkit developers seem to still think it's fine to use though, so I think we can use it, but we should be prepared to swap it out if necessary.

danielcj2:
> > Either or works honestly. Heck we could also just use redux-persist strictly for essential states that make the app function or enhance it. Stuff like auth state, form states, temporary UI states (e.g. open tab panels, alerts, modals maybe?, perhaps if we make a collapsible sidebar?, and whatever may be),
> 
> Alright, I think we can give this a shot, we can whitelist the states we want to persist (or blacklist those what we don't want), that should resolve performance issues.
> 
> BIg caveat here though is that this library appears to be unmaintained, last commit was 3 years ago. Redux-toolkit developers seem to still think it's fine to use though, so I think we can use it, but we should be prepared to swap it out if necessary.

Sounds like a plan. The implementation doesn't seem too complicated so if we come across issues we can easily swap it out.

danielcj2 (approved):

MuhammadKhalilzadeh:
I'm interested in this approach and what your doing looks interesting to me!

MuhammadKhalilzadeh (approved):

PR #176, opened 2024-06-22
Refactor Profile Settings elements with their respective components, #172

danielcj2:
Refactored input, avatar elements and added some validation.

### DIFF
--- a/Client/src/Components/TabPanels/ProfileSettings/PasswordPanel.jsx
+++ b/Client/src/Components/TabPanels/ProfileSettings/PasswordPanel.jsx
@@ -2,22 +2,12 @@ import TabPanel from "@mui/lab/TabPanel";
 import React, { useState } from "react";
 import AnnouncementsDualButtonWithIcon from "../../Announcements/AnnouncementsDualButtonWithIcon/AnnouncementsDualButtonWithIcon";
 import { useTheme } from "@emotion/react";
-import {
-  Box,
-  Divider,
-  FormControl,
-  IconButton,
-  InputAdornment,
-  OutlinedInput,
-  Stack,
-  Typography,
-} from "@mui/material";
-import VisibilityOff from "@mui/icons-material/VisibilityOff";
-import Visibility from "@mui/icons-material/Visibility";
+import { Box, Divider, Stack, Typography } from "@mui/material";
 import InfoOutlinedIcon from "@mui/icons-material/InfoOutlined";
 import WarningAmberOutlinedIcon from "@mui/icons-material/WarningAmberOutlined";
 import ButtonSpinner from "../../ButtonSpinner";
 import PasswordTextField from "../../TextFields/Password/PasswordTextField";
+import { editPasswordValidation } from "../../../Validation/validation";
 
 /**
  * PasswordPanel component manages the form for editing password.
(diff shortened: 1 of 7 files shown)


MuhammadKhalilzadeh:
This is something that I've also learned myself recently, for the sake of applying best practices as much as possible.
Try passing values through component values true props instead of the current style of your code. What I mean is like this:
`const myComponent = (props) => {};`

And for props try to assign their types at least for though that we know what type they would be, like this:

```
myComponent.propTypes = {
  arg1: PropTypes.string(for exampes!).isRequired,
....
}
```

and also their default values:

```
myComponent.defaultTypes = {
  arg1: "something",
....
}
```

And then generate your `JSDoc`.
In that fashion, our code will be at it's best

MuhammadKhalilzadeh (approved):

Client/src/Components/TabPanels/ProfileSettings/PasswordPanel.jsx
```diff
@@ -31,6 +32,76 @@ const PasswordPanel = () => {
...
+            ? "*This field is required."
+            : value.length < 8
+            ? "*Password must be at least 8 characters long."
+            : !/[A-Z]/.test(value)
+            ? "*Password must contain at least one uppercase letter."
+            : !/\d/.test(value)
+            ? "*Password must contain at least one number."
+            : !/[!@#$%^&*]/.test(value)
+            ? "*Password must contain at least one symbol."
+            : "";
```
ajhollid, line 80: We use the [Joi](https://joi.dev/api/?v=17.13.3) library for field validation, and I believe we already have validation schemas set up for username and password.

Please see register/validation pages for implementaiton examples, it's pretty easy to use

ajhollid (changes requested):
Looks good to me for the most part :+1: 

We use the Joi library for field validation though, no need to roll our own.  There's some examples of that in the login and register pages, and I believe we already have validation schema defiend for those fields.

Once the Joi validation is set up I think we can merge this in.

As @MuhammadKhalilzadeh mentioned please do include all prop types and JSdocs if there are any missing.

danielcj2:
Integrated joi validation for input fields. Each field checks itself as you type (using an onChange). If there are any errors, the submit button becomes disabled until the issues are resolved. 

As to @MuhammadKhalilzadeh 's comment, I double-checked, and I couldn't find any missing props. Unless I'm blind and completely overlooked them. In that case I apologize and kindly ask you to point them out 😂

Client/src/Components/TabPanels/ProfileSettings/ProfilePanel.jsx
```diff
@@ -23,63 +24,38 @@ const ProfilePanel = () => {
   const [isLoading, setIsLoading] = useState(false);
 
   //for testing, will tweak when I implement redux slice
+  const idToName = {
+    "edit-first-name": "firstName",
+    "edit-last-name": "lastName",
+    "edit-email": "email",
```
ajhollid, line 30: Can you double check these mappings? I believe the backend is is expecting all lower case, ie `firstname` and `lastname`.

ajhollid (changes requested):
Validation looks good, but I think the form field names need to be adjusted as backend expects all lower case names.

Client/src/Components/TabPanels/ProfileSettings/ProfilePanel.jsx
danielcj2: Sure! on that topic should I also change the naming for the password panel form fields? I'm not sure how the backend is currently set up and what fields should correspond to what. 

This is how I have the mappings for the new password form set atm:
`"edit-current-password": "currentPassword"` - current (old) password
`"edit-new-password": "password"` 
`"edit-confirm-password": "confirm"`

ajhollid: I can't remember exactly what they are either, they are all in the docs for the respective API endpoints though.

If there isn't an endpoint for whichever request you're working on then you can leave them as is and we can revise if needed when the time comes.

danielcj2: I don't believe there is an endpoint for resetting password w/o a recovery token currently so I'll leave it as is.

ajhollid: I believe you'll be hitting

POST/api/v1/auth/user/{user ID}

for editing users,so naming conventions should match with that one.

I don't think confirm password field needs to be passed to the backend as match should already be verified on client side.

We can update the backend if we want to verify there as well too though.

danielcj2: Makes sense. I'll update the names in that case to match the convention. As for the newpassword and confirm fields I'll leave them as TBD for now and will touch base once I start working on that part.

ajhollid (approved):
Looks good on my end

PR #179, opened 2024-06-24
Implemented logout funcationality, resolves #178

### LINKED ISSUE #178: Frontend - Logout
Implement logout functionality

ajhollid:
Auth state now cleared out when Logout is selected from the navbar

### DIFF
--- a/Client/src/Features/Monitors/monitorsSlice.js
+++ b/Client/src/Features/Monitors/monitorsSlice.js
@@ -46,7 +46,14 @@ export const getMonitorsByUserId = createAsyncThunk(
 const monitorsSlice = createSlice({
   name: "monitors",
   initialState,
-  reducers: {},
+  reducers: {
+    clearMonitorState: (state) => {
+      state.isLoading = false;
+      state.monitors = [];
+      state.success = null;
+      state.msg = null;
+    },
+  },
   extraReducers: (builder) => {
     builder
       // *****************************************************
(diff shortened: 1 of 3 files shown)


danielcj2:
After clearing the auth state, I'm concerned whether the monitors state or other remnants (if we add more slices) might persist in localStorage since we're using `redux-persist`. I'm not sure if that's even how it works but just throwing it out there.

ajhollid:
> After clearing the auth state, I'm concerned whether the monitors state or other remnants (if we add more slices) might persist in localStorage since we're using `redux-persist`. I'm not sure if that's even how it works but just throwing it out there.

Working on this right now :smile:

Client/src/Features/Monitors/monitorsSlice.js
```diff
@@ -46,7 +46,14 @@ export const getMonitorsByUserId = createAsyncThunk(
 const monitorsSlice = createSlice({
   name: "monitors",
   initialState,
-  reducers: {},
+  reducers: {
+    clearMonitorState: (state) => {
+      state.isLoading = false;
+      state.monitors = [];
+      state.success = null;
+      state.msg = null;
```
danielcj2, line 54: You could technically reset the monitor state to initialState in case we add more values, something like `clearMonitorState: () => initialState`, but I doubt it's needed.

danielcj2 (commented):
Looks good to me!

danielcj2 (approved):

Skorpios604 (approved):
Pretty straight forward. I see no issues here. Good to merge!

PR #181, opened 2024-06-24
Feat/protected routes, resolves #180

### LINKED ISSUE #180: Frontend - Protect routes
Implement protected route component

ajhollid:
Implemented a protected route component
Fixed some redirects to point to `/login` instead of `/`


danielcj2:
Good stuff 🙌 Will the redirect be active in our environment, and will we need to disable it/ log in to continue working on the code?

ajhollid:
> Good stuff 🙌 Will the redirect be active in our environment, and will we need to disable it/ log in to continue working on the code?

Yep it will be active, you would have to disable it or log in if you need to work on components. 

One of the other pull requests is for a TTL on the token though, so you can set it to expire in 99 days or something for dev purposes.  Then you only need to login once

PR #190, opened 2024-06-25
Create new monitor without dropdowns

MuhammadKhalilzadeh:
the Create New Monitor page is complete except dropdowns since I could not make it work to get its value or give it list of options, although I added a text field as a placeholder for dropdowns for now. And, your able to capture input values, too.

### DIFF
--- a/Client/src/Components/Checkbox/Checkbox.css
+++ b/Client/src/Components/Checkbox/Checkbox.css
@@ -12,10 +12,26 @@
   cursor: pointer;
 }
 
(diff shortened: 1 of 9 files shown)


ajhollid (approved):
UI looks good to me :)

I think we can merge this in, then I will take a look at the functionality and those dropdowns and see if we can get it all working.

Thanks for your hard work!

danielcj2 (approved):
Looks great! Good stuff 🙌

PR #197, opened 2024-06-25
Moved CreateNewMonitor to its own directory as it is a page unto itself

ajhollid:
Create New Monitor button now redirects to `/monitors/create` and displays the `CreateNewMonitor` page

TODO:

`CreateNewMonitor` page uses the styles from the `Monitors page`.  I simply copied the whole index.css file to the new `CreateNewMonitor` directory, so there are probably unused styles in that file.  We should remove any unused styles.


danielcj2 (commented):
Looks good. Maybe have @MuhammadKhalilzadeh look into it since he's familiar with the code?

Skorpios604 (approved):
Looks great! Functional but not responsive. If you drop down to mobile size the text goes into each other. I suppose that will be fixed when the styling is fixed.

ajhollid:
> Looks great! Functional but not responsive. If you drop down to mobile size the text goes into each other. I suppose that will be fixed when the styling is fixed.

Good call, responsiveness is something we need to address in the future, I think there are many components and pages that aren't responsive yet.

MuhammadKhalilzadeh (approved):
Nice code you've wrote. Looks good to me, Kudos.

PR #206, opened 2024-06-26
Feat/terry component

terryjung0244:
Feat : implement new feature for AlertMessage component

- Added new feature to handle dynamic icon color changes.
- Updated AlertMessage component with additional props.
- Included new utility functions for color handling.

Please revie my code. Thank you.

### DIFF
--- a/Client/src/Components/TerryComponent/index.jsx
+++ b/Client/src/Components/TerryComponent/index.jsx
@@ -1,20 +1,72 @@
 // eslint-disable-next-line no-unused-vars
-import React from "react";
+import React, { useState } from "react";
 import { Button, Container } from "@mui/material";
 
 const index = () => {
+  const [notification, setNotification] = useState({
+    open: false,
+    title: "",
+    message: "",
+    type: "",
+    alertMessageType: "",
+    actionText: "",
+  });
+
+  const handleOpenAlertMessage = (type, title, message, actionText) => {
+    setNotification({
+      ...notification,
+      open: true,
+      title,
+      type,
+      message,
+      actionText,
+    });
+  };
+
   return (
     <div>
       <Container style={{ marginTop: "100px", marginLeft: "70px" }}>
-        <Button variant="contained">Feature Alert</Button>
+        <Button
+          variant="contained"
+          onClick={() =>
+            handleOpenAlertMessage(
+              "release",
+              "We've just released a new feature",
+              "Lorem ipsum dolor sit amet consectetur adipiscing elit. Aliquid pariatur, ipsum dolor.",
+              "View changes"
+            )
+          }
+        >
+          Feature Alert
+        </Button>
         <Button
           style={{ margin: "10px 20px 10px 20px" }}
           className="button"
           variant="contained"
+          onClick={() =>
+            handleOpenAlertMessage(
+              "problem",
+              "There was a problem with that action",
+              "Lorem ipsum dolor sit amet consectetur adipiscing elit. Aliquid pariatur, ipsum dolor.",
+              "Learn more"
+            )
+          }
         >
           Problem Alert
         </Button>
-        <Button variant="contained">Update Alert</Button>
+        <Button
+          variant="contained"
+          onClick={() =>
+            handleOpenAlertMessage(
+              "update",
+              "Successfully updated profile",
+              "Lorem ipsum dolor sit amet consectetur adipiscing elit. Aliquid pariatur, ipsum dolor.",
+              "View changes"
+            )
+          }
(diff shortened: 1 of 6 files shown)


Client/src/Components/TerryComponent/reusable/AlertMessage.jsx
```diff
@@ -0,0 +1,74 @@
...
+  type,
+  title = "",
+  message,
+  dismissText = "Learn more",
+  actionText,
+  close,
+  dismissClose,
+}) => {
+  // Function to determine icon color based on the title
+  const switchIconColor = (type) => {
```
ajhollid, line 18: Functions should al linclude JSDocs so that their usage is well documented and so that we benefit from itellisense suggestion and autocomplete features.

Please see other components and functions for implementations if you've not worked iwth JSdocs before

Client/src/Components/TerryComponent/reusable/AlertMessage.jsx
```diff
@@ -0,0 +1,74 @@
...
+      <Alert
+        onClose={close}
+        severity="info"
+        variant="filled"
+        sx={{
+          width: "100%",
+          backgroundColor: "white",
+          color: "black",
+          border: "2px solid lightgrey",
+          "& .MuiAlert-icon": {
```
ajhollid, line 42: This is a sample component so it's not a big deal, but if this were a production component we would not want to hardcode styles.

Styles are ideally built off our MUI theme if they are used across components, or in the component's CSS file if it is a one off component

Client/src/Components/TerryComponent/reusable/AlertMessage.jsx
```diff
@@ -0,0 +1,74 @@
...
+
+AlertMessage.propTypes = {
+  open: PropTypes.bool.isRequired,
+  type: PropTypes.string.isRequired,
+  title: PropTypes.string,
+  message: PropTypes.string.isRequired,
+  dismissText: PropTypes.string,
+  actionText: PropTypes.string,
+  close: PropTypes.func,
+  dismissClose: PropTypes.func,
```
ajhollid, line 71: Nice, thanks for including proptypes here :+1:

Client/src/Components/TerryComponent/index.jsx
```diff
@@ -1,7 +1,95 @@
...
+  const [notification, setNotification] = useState({
+    open: false,
+    title: "",
+    message: "",
+    type: "",
+    alertMessageType: "",
+    actionText: "",
+  });
+
+  const handleOpenAlertMessage = (type, title, message, actionText) => {
```
ajhollid, line 16: Given the number of parameters here I might consider passing this function a config object instead:
```
const handleOpenAlertMessage = (config) => {
  ...
}
```
then you can simply pass it all your params:

```
{
  type: "type",
  title: "title",
  message: "message",
  actionText: "actionText
}
```

The JSdocs would then describe the shape of the config object.

This is not a hard and fast requirement, but it's something to think about when your function start taking many parameters.

At the very least the function requires JSdocs so that we know what parameters to give it.

Client/src/Components/TerryComponent/index.jsx
```diff
@@ -1,7 +1,95 @@
...
+            handleOpenAlertMessage(
+              "update",
+              "Successfully updated profile",
+              "Lorem ipsum dolor sit amet consectetur adipiscing elit. Aliquid pariatur, ipsum dolor.",
+              "View changes"
+            )
+          }
+        >
+          Update Alert
+        </Button>
```
ajhollid, line 78: There's a lot of duplicated code here, you could consider declaring the buttons in an array and then mapping them to a component.  Again not a hard and fast requirement, but it's always nice to reduce code repetition wherever possible. 

It makes the code cleaner to read and if we need to update the buttons we can do it all at once rather than 3 times seperately.  Something to keep in mind :+1:

ajhollid (commented):
Overall the component looks great, I'm really happy with how it looks.  I don't see any major issues with the implementation.

In general my suggestions are to think about reusability and maintainability of code when writing your components.  I see some hardcoded styles, some missing JSdocs, and some duplicated code that could be refactored.

They are small details, but added together they make the difference between code that is maintainable and code that is not.

Great work, this is a fantastic start.  I'm glad you got the Git workflow figured out and are ready to contribute!

Just a side note we won't actually merge this branch as building the component was mostly for getting you up to speed with our stanards and Git workflow.

Skorpios604:
I think we can close this @ajhollid since we got what needed to be done.

PR #209, opened 2024-06-26
Monitors

MuhammadKhalilzadeh:
`Bugs` mentioned `in the UI` by Gorkem are `fixed`.
`Dropdown` is `replaced` with the Placeholder (TextField I put their so far). but, **still could not bind it to see the list of options**, still not sure why!


Client/src/Pages/CreateNewMonitor/index.jsx
```diff
@@ -228,7 +257,14 @@ const CreateNewMonitor = () => {
...
           <div className="advanced-setting-config">
             <div className="adv-setting-title">Check frequency</div>
-            <FlexibileTextField />
+            <div className="monitors-gaps-small"></div>
+            <Dropdown
+              id="freq-dropdown"
+              label="1 minutes"
+              onChange={handleFrequencyOptionChange}
+              options={frequencyOptions}
+              value={frequencyOptions}
```
ajhollid, line 266: I think we can use the `Select` component instead of `Autocomplete` as users should be picking from a list.  That should resolve some of the errors here.

Client/src/Pages/Monitors/CreateNewMonitor.jsx
```diff
@@ -28,6 +40,17 @@ const CreateNewMonitor = () => {
...
 
+  const [port, setPort] = useState("Select a port to check");
+  const [frequncy, setFrequency] = useState("1 minute");
+
+  const handlePortOptionChange = (newValue) => {
+    setPort(newValue);
+  };
+
+  const handleFrequencyOptionChange = (newValue) => {
+    setPort(newValue);
```
Skorpios604: Should this not be 
```javascript
const handleFrequencyOptionChange = (newValue) => {
    setFrequency(newValue);
```

Client/src/Pages/Monitors/CreateNewMonitor.jsx
```diff
@@ -28,6 +40,17 @@ const CreateNewMonitor = () => {
   const [urlToMonitor, setUrlToMonitor] = useState("");
   const [friendlyName, setFriendlyName] = useState("");
 
+  const [port, setPort] = useState("Select a port to check");
+  const [frequncy, setFrequency] = useState("1 minute");
```
Skorpios604: Typo here
Should be frequency

Client/src/Components/Button/index.jsx
```diff
@@ -48,10 +48,11 @@ const levelConfig = {
  * <Button type="submit" level="error" label="Error" disabled sx={{marginTop: "1rem"}}/>
  */
 
-const Button = ({ type, level, label, disabled, img, onClick, sx }) => {
+const Button = ({ id, type, level, label, disabled, img, onClick, sx }) => {
   const { variant, color } = levelConfig[level];
```
Skorpios604, line 52: Should we have a default value here? just incase variant and color become undefined if level is not found?

Client/src/Pages/Monitors/index.css
```diff
@@ -4,6 +4,10 @@
   padding: 4vw;
 }
 
+.monitors {
+  font-family: var(--env-var-font-familt-1);
```
Skorpios604, line 8: Typo here where it says familt instead of family

Client/src/Pages/Monitors/index.css
```diff
@@ -74,10 +82,21 @@
...
+
+#create-new-monitor-btn {
+  padding-top: 6px;
+}
+
+#ports-dropdown,
+#freq-dropdown {
+  width: 226.8px;
+  height: 20px;
+}
```
Skorpios604, line 102: get rid of hardcoded values

Client/src/Pages/Monitors/CreateNewMonitor.jsx
```diff
@@ -162,9 +185,15 @@ const CreateNewMonitor = () => {
...
             />
-            <div className="monitors-gaps-medium"></div>
+            <div className="monitors-gaps-small-plus"></div>
             <div className="monitors-dropdown-holder">
-              <FlexibileTextField placeholder="Select a port to check" />
+              <Dropdown
+                id="ports-dropdown"
+                label="Select a port to check"
+                onChange={handlePortOptionChange}
+                options={portOptions}
```
Skorpios604: should this just be port instead of port options? Maybe that will fix the issue?

Client/src/Pages/Monitors/CreateNewMonitor.jsx
```diff
@@ -228,7 +257,14 @@ const CreateNewMonitor = () => {
         rightLayout={
           <div className="advanced-setting-config">
             <div className="adv-setting-title">Check frequency</div>
-            <FlexibileTextField />
+            <div className="monitors-gaps-small"></div>
+            <Dropdown
+              id="freq-dropdown"
+              label="1 minutes"
+              onChange={handleFrequencyOptionChange}
+              options={frequencyOptions}
```
Skorpios604: should this just be frequency instead of frequency options? Maybe that will fix the issue?

Skorpios604 (changes requested):

ajhollid (approved):
I think we can merge this in and then work on it further

Skorpios604 (approved):

Client/src/Pages/Monitors/CreateNewMonitor.jsx
MuhammadKhalilzadeh: Yeah, right. But either way won't work

Client/src/Pages/Monitors/index.css
MuhammadKhalilzadeh: There are some of the hardcoded values that we can not remove, since there are paddings and margins built-in for the MUI components and we need to set these hardcoded values according to that

Client/src/Pages/Monitors/CreateNewMonitor.jsx
MuhammadKhalilzadeh: No, apparently this is an array type that needs to be passed.

Client/src/Pages/Monitors/CreateNewMonitor.jsx
MuhammadKhalilzadeh: No, apparently this is an array type that needs to be passed.

Issue #193, opened 2024-06-25
Front End - Good First issue


danielcj2:
The password input field styling appears to be broken.

![image](https://github.com/bluewave-labs/bluewave-uptime/assets/122758391/42b90b43-a4c2-4a4c-9f6d-7cbc5f7cb656)

faizan-o:
I Can Fix The Styling If You Please Assign The Issue To Me.

ajhollid:
> I Can Fix The Styling If You Please Assign The Issue To Me.

Hi @faizan-o, if you'd like to work on this you can fork the repo and make a pull request with your changes.  We will review your PR once it is made :+1:

PR #210, opened 2024-06-26
Client side upload picture, #193

danielcj2:
Design is not final, just wanted to get everyone's input on it.

For testing, the loading states are currently managed using a timeout mechanism but will swap to redux once everything is streamlined. The progress bar is managed through `setUploadProgress`, a `useState` hook, which increments linearly at fixed intervals (`x` ms) using `setInterval` . This is likely to be changed in the future to properly and accurately reflect progress. This could be done using redux through the `auth` slice or setting up a separate slice. I would prefer the latter just to keep `auth` as clean as possible, especially if we end up using different image renders, and for editting the picture. What do you all think?

Stuff not implemented yet: 
 - File (type/size) restrictions.
 - Connect to redux store.
 - Edit picture functionality (stuff like crop/ centering the image) - since it's not as urgent we can leave this on the backlog for now
 Will create separate issues for these.

### DIFF
--- a/Client/src/Components/ProgressBars/index.css
+++ b/Client/src/Components/ProgressBars/index.css
@@ -1,0 +1,18 @@
+.progress-bar-container{
+    background-color: #fafafa;
+    padding: var(--env-var-spacing-1-plus);
+}
+.progress-bar-container h2.MuiTypography-root,
+.progress-bar-container p.MuiTypography-root {
+  color: var(--env-var-color-2);
+  font-size: var(--env-var-font-size-medium);
+}
+.progress-bar-container p.MuiTypography-root {
+  opacity: 0.6;
+}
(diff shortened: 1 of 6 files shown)


Skorpios604 (approved):
Looks pretty neat. I like the idea of the loading bar for the image. So currently it loads according to a set timer but eventually it would load according to speed and size?

ajhollid (approved):
I like the design, looks good to me.

Definitely need some restrictions and validation in the future, looking good for now though :+1:

danielcj2:
> Looks pretty neat. I like the idea of the loading bar for the image. So currently it loads according to a set timer but eventually it would load according to speed and size?

@Skorpios604 Exactly! that would be the goal.

PR #230, opened 2024-06-29
Implementation of the Incidents page. 

Skorpios604:
Still needs a little bit of work. Got to put it into a div to make it look more like the figma file. Still need pagination on the bottom but the component is being reworked. Still need to add the filter by status dropdown near the top.


Client/src/Pages/Incidents/index.jsx
```diff
@@ -1,9 +1,72 @@
-import React from 'react'
+import { styled } from '@mui/material/styles';
+import Table from '@mui/material/Table';
+import TableBody from '@mui/material/TableBody';
+import TableCell, { tableCellClasses } from '@mui/material/TableCell';
+import TableContainer from '@mui/material/TableContainer';
+import TableHead from '@mui/material/TableHead';
+import TableRow from '@mui/material/TableRow';
+import Paper from '@mui/material/Paper';
+import StatusLabel from "../../Components/Label/";
```
ajhollid, line 9: Currently doesn't compile as `Label` has no default export, so unless we provide one we need to import it as a named import:

import { StatusLabel } from "../../Components/Label/";

Client/src/Pages/Incidents/index.jsx
```diff
@@ -1,9 +1,72 @@
...
+  '&:nth-of-type(n+1)': {
+    backgroundColor: theme.palette.common.white, 
+  },
+  '&:last-child td, &:last-child th': {
+    border: 0,
+  },
+}));
+
+function createData(name, calories, fat, carbs, protein) {
+  return { name, calories, fat, carbs, protein };
```
ajhollid, line 35: I guess this is mock data for now so not a big deal, but the names definitely don't make sense in this context :joy:

Client/src/Pages/Incidents/index.jsx
```diff
@@ -1,9 +1,72 @@
...
+      <Table sx={{ minWidth: 700 }} aria-label="customized table">
+        <TableHead>
+          <TableRow>
+            <StyledTableCell>Status</StyledTableCell>
+            <StyledTableCell align="right">Date & Time</StyledTableCell>
+            <StyledTableCell align="right">Message</StyledTableCell>
+          </TableRow>
+        </TableHead>
+        <TableBody>
+          {rows.map((row, index) => (
```
ajhollid, line 59: A number of errors in the console here:

image

The first couple are related to `StatusLabel`, we need to add statuses to its acceptable status field.

The key error is because `row.name` is having `toString()` called on it under the hood, so it is parsed to `[object object]` for every row, hence the duplicate keys.

ajhollid (changes requested):
Page looks great!  Couple of minor bug fixes and I think it's good to go :+1:

Client/src/Pages/Incidents/index.jsx
```diff
@@ -57,7 +57,7 @@ export default function CustomizedTables() {
         </TableHead>
         <TableBody>
           {rows.map((row, index) => (
-            <StyledTableRow key={row.name}>
+            <StyledTableRow key={index}>
```
ajhollid, line 60: I think this is still a WIP while we wait for data to be passed in, but index should not be used for keys.  Since the order of the items isn't necessarily fixed the key is unstable, so we really should use a unique and stable key.  We can change this when the component is ready for use

Client/src/Components/Label/index.jsx
```diff
@@ -153,7 +153,7 @@ const StatusLabel = ({ status, customStyles }) => {
 };
 
 StatusLabel.propTypes = {
-  status: PropTypes.oneOf(["Seen", "Waiting", "New", "Active"]),
+  status: PropTypes.oneOf(["Seen", "Waiting", "New", "Active", "Down", "Cannot resolve"]),
```
ajhollid, line 156: Awesome, thanks for adding those :+1:

ajhollid (approved):

Client/src/Pages/Incidents/index.jsx
Skorpios604: Sounds good thank you for your review!

Client/src/Components/Label/index.jsx
Skorpios604: No worries thank you!

danielcj2 (approved):
Seems like Alex went through everything already. Looks good to me 👍

PR #231, opened 2024-07-01
Feat/image upload, resolves #221

### LINKED ISSUE #221: Backend - Image for user profile
Save user's profile image.  Need to consider storage options:

* Direct in DB
* Third party storage (S3) and save URL to db

ajhollid:
Backend implementation for image upload
 - Stores image bytes directly to DB

### DIFF
--- a/Server/db/MongoDB.js
+++ b/Server/db/MongoDB.js
@@ -32,7 +32,15 @@ const connect = async () => {
  */
 const insertUser = async (req, res) => {
   try {
-    const newUser = new UserModel({ ...req.body });
+    const userData = { ...req.body };
+    console.log(req.file);
+    if (req.file) {
+      userData.profileImage = {
+        data: req.file.buffer,
+        contentType: req.file.mimetype,
+      };
+    }
+    const newUser = new UserModel(userData);
     await newUser.save();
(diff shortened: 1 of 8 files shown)


Server/db/MongoDB.js
```diff
@@ -32,7 +32,15 @@ const connect = async () => {
...
 const insertUser = async (req, res) => {
   try {
-    const newUser = new UserModel({ ...req.body });
+    const userData = { ...req.body };
+    console.log(req.file);
+    if (req.file) {
+      userData.profileImage = {
+        data: req.file.buffer,
+        contentType: req.file.mimetype,
+      };
```
danielcj2, line 41: Was there a reason to include the image upload logic in `insertUser`? If I'm not mistaken this is the function that handles user registration correct?

ajhollid: I assume users should be able to select a profile image when they create an account?

danielcj2: That would make sense, I was only asking because the figma design plans doesn't include that upload feature.

danielcj2 (approved):

Skorpios604 (approved):

PR #242, opened 2024-07-02
Feature/issue 225 incidents comp

terryjung0244:
I created an Incidents component with a basic structure. I defined a 'maintenanceItems' array and mapped it to 'ListItem' components with icons.

Please review my codes.

### DIFF
--- a/Client/src/Pages/Incidents/index.css
+++ b/Client/src/Pages/Incidents/index.css
@@ -13,13 +13,16 @@
 }
 
 .maintenance-title {
-  font-size: 20px;
+  color: var(--env-var-color-2);
+  font-size: 16px;
   font-weight: bold;
   margin-top: 40px;
   margin-bottom: 30px;
 }
 
 .checklist-item {
+  font-size: 13px;
+  color: var(--env-var-color-2);
   display: flex;
   margin-bottom: 10px;
 }
--- a/Client/src/Pages/Incidents/index.css
+++ b/Client/src/Pages/Incidents/index.css
@@ -1,13 +1,33 @@
-.maintenance-checklist-card {
-  max-width: 314;
+.maintenance-checklist-main {
+  max-width: 350px;
+  min-height: 390px;
   margin: auto;
-  padding: 2;
-  box-shadow: 3;
+  padding: 5px;
 }
 
-.maintenance-checklist-box {
-  height: 100;
-  background-color: grey;
-  border-radius: 1px;
-  margin-bottom: 2px;
+.maintenance-image {
+  width: 280px;
+  height: 150px;
+  background-color: whitesmoke;
+  border-radius: 10px;
+}
+
+.maintenance-title {
+  font-size: 20px;
+  font-weight: bold;
+  margin-top: 40px;
+  margin-bottom: 30px;
+}
+
+.checklist-item {
+  display: flex;
+  margin-bottom: 10px;
+}
+
+.checklist-item-text {
+  margin-left: 10px;
+}
+
+.maintenance-checklist-main .maintenance-checklist-button {
+  margin-top: 30px;
 }
--- a/Client/src/Pages/Incidents/index.css
+++ b/Client/src/Pages/Incidents/index.css
@@ -1,0 +1,13 @@
+.maintenance-checklist-card {
+  max-width: 314;
+  margin: auto;
+  padding: 2;
+  box-shadow: 3;
+}
+
+.maintenance-checklist-box {
+  height: 100;
+  background-color: grey;
+  border-radius: 1px;
+  margin-bottom: 2px;
+}
--- a/Client/src/Pages/Incidents/index.jsx
+++ b/Client/src/Pages/Incidents/index.jsx
@@ -1,9 +1,18 @@
-import React from 'react'
+import React from "react";
+import { Card, CardContent, Box, Typography, Button } from "@mui/material";
 
 const Incidents = () => {
-    return (
-        <div>Incidents</div>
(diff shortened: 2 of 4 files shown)


Client/src/Pages/Incidents/index.jsx
```diff
@@ -1,9 +1,52 @@
...
-export default Incidents
\ No newline at end of file
+  return (
+    <Card>
+      <CardContent>
+        <Box />
+        <Typography>Create a maintenance window to</Typography>
+        <List>
+          {maintenanceItems.map((item, index) => {
+            <ListItem key={index}>
```
ajhollid, line 30: Index [should not be used for keys i](https://react.dev/learn/rendering-lists#why-does-react-need-keys)n a React component:

From the react docs:

> You might be tempted to use an item’s index in the array as its key. In fact, that’s what React will use if you don’t specify a key at all. But the order in which you render items will change over time if an item is inserted, deleted, or if the array gets reordered. Index as a key often leads to subtle and confusing bugs.
> 
> Similarly, do not generate keys on the fly, e.g. with key={Math.random()}. This will cause keys to never match up between renders, leading to all your components and DOM being recreated every time. Not only is this slow, but it will also lose any user input inside the list items. Instead, use a stable ID based on the data.
> 
> Note that your components won’t receive key as a prop. It’s only used as a hint by React itself
> If your component needs an ID, you have to pass it as a separate prop: <Profile key={id} userId={id} />.

Client/src/Pages/Incidents/index.css
```diff
@@ -0,0 +1,13 @@
...
+  margin: auto;
+  padding: 2;
+  box-shadow: 3;
+}
+
+.maintenance-checklist-box {
+  height: 100;
+  background-color: grey;
+  border-radius: 1px;
+  margin-bottom: 2px;
```
ajhollid, line 12: Do these values exist in the theme?  If so those should be used instead of hardcoding.

ajhollid (changes requested):
In general looks good!

Index of an array should not be used for keys when rendering an array of items though, please see link in review

Client/src/Pages/Incidents/index.jsx
```diff
@@ -1,51 +1,37 @@
...
 import CheckCircleOutlineIcon from "@mui/icons-material/CheckCircleOutline";
 
 const Incidents = () => {
   const maintenanceItems = [
-    "Mark your maintenance periods",
-    "Eliminate any misunderstandings",
-    "Stop sending alerts in maintenance windows",
+    { id: 1, text: "Mark your maintenance periods" },
+    { id: 2, text: "Eliminate any misunderstandings" },
+    { id: 3, text: "Stop sending alerts in maintenance windows" },
```
ajhollid, line 10: This works :+1:

ajhollid (approved):
Looks good to me :+1:

Client/src/Pages/Incidents/index.jsx
```diff
@@ -1,9 +1,38 @@
...
+  ];
 
-export default Incidents
\ No newline at end of file
+  return (
+    <div className="maintenance-checklist-main">
+      <div className="maintenance-image"></div>
```
danielcj2, line 15: On Figma there is an option to export/ copy frames or components as svgs, so you can actually take the skeleton frame from the design and use it as the background for `maintenance-image`.

Client/src/Pages/Incidents/index.css
```diff
@@ -0,0 +1,33 @@
...
+  background-color: whitesmoke;
+  border-radius: 10px;
+}
+
+.maintenance-title {
+  font-size: 20px;
+  font-weight: bold;
+  margin-top: 40px;
+  margin-bottom: 30px;
+}
```
danielcj2, line 20: The font color of the text should be `#475467` and a size of `16px`. Please see the Figma design.

In terms of measurements, stuff like paddings, margins, font-sizes and what not, we aim to stick closely to the original Figma design. Makes life easier when you're not always guesstimating. We have some pre-defined variables in `:root` inside the `Client/src/index.css` file, or if you are more inclined to use inline styling we also have a theme schema set up, in `Client/src/Utils/Theme.js`. I would recommend looking over those and applying the styles where possible.

Client/src/Pages/Incidents/index.css
```diff
@@ -0,0 +1,33 @@
...
+
+.checklist-item {
+  display: flex;
+  margin-bottom: 10px;
```
danielcj2, line 24: The font color of the text should be `#475467` and a size of `13px`. Please see the Figma design.

danielcj2 (changes requested):
Looks about right, just a few styling changes and should be good to merge!

Please use styling variables where appropriate.

Client/src/Pages/Incidents/index.jsx
terryjung0244: 👍

Client/src/Pages/Incidents/index.jsx
terryjung0244: Thanks for your comment.

Client/src/Pages/Incidents/index.css
terryjung0244: Fixed.

Client/src/Pages/Incidents/index.css
terryjung0244: Thank you.

danielcj2 (approved):
Looks great! Thank you for implementing the changes.

PR #250, opened 2024-07-03
Added admin role to registration, resolves #249

### LINKED ISSUE #249: Frontend - Add role to user created in Registration page
The registration page should only be shown once for the admin user, so the role created there should be an admin role

ajhollid:
Admin role added to user created by the Registration page


danielcj2 (approved):
Looks good

Issue #235, opened 2024-07-02
Backend - Alert on status change


ajhollid:
Alerts should be fired when a service's status changes

Skorpios604:
Complete!

PR #251, opened 2024-07-03
Monitor now has a status, status will be updated on new check, #235

ajhollid:
Monitor now has a status that is updated whenever a new check is created.

Logic in place for sending an alert if status changes from up to down or down to up.

### DIFF
--- a/Server/models/Check.js
+++ b/Server/models/Check.js
@@ -31,19 +31,25 @@ const CheckSchema = mongoose.Schema(
   }
 );
 
+/**
+ * When a check is created, we should update the stauts of the associated monitor
+ * if it has changed.  Monitor starts with default status: true
+ */
+
 CheckSchema.pre("save", async function (next) {
   const monitor = await mongoose.model("Monitor").findById(this.monitorId);
+  if (monitor.status !== this.status) {
+    if (monitor.status === true && this.status === false) {
+      // TODO issue alert
+      console.log("Monitor went down");
+    }
 
-  if (monitor.status === true && this.status === false) {
-    console.log("Monitor went down");
+    if (monitor.status === false && this.status === true) {
+      // TODO issue alert
+      console.log("Monitor went up");
+    }
+    await monitor.save();
   }
-
-  if (monitor.status === false && this.status === true) {
-    console.log("Monitor went up");
-  }
-
-  monitor.status = this.status;
-  await monitor.save();
 
   next;
 });
--- a/Server/models/Check.js
+++ b/Server/models/Check.js
@@ -34,6 +34,7 @@ const CheckSchema = mongoose.Schema(
 /**
  * When a check is created, we should update the stauts of the associated monitor
  * if it has changed.  Monitor starts with default status: true
+ * @param {function} next - Callback to proceed to the next middleware.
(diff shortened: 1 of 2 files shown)


Server/models/Check.js
```diff
@@ -31,4 +31,27 @@ const CheckSchema = mongoose.Schema(
...
+    }
+
+    if (monitor.status === false && this.status === true) {
+      // TODO issue alert
+      console.log("Monitor went up");
+    }
+    await monitor.save();
+  }
+
+  next;
```
danielcj2, line 54: Is this a typo? Shouldn't it be `next()` instead?

Server/models/Check.js
```diff
@@ -31,4 +31,27 @@ const CheckSchema = mongoose.Schema(
...
 );
 
+/**
+ * When a check is created, we should update the stauts of the associated monitor
+ * if it has changed.  Monitor starts with default status: true
+ */
+
+CheckSchema.pre("save", async function (next) {
+  const monitor = await mongoose.model("Monitor").findById(this.monitorId);
+  if (monitor.status !== this.status) {
```
MuhammadKhalilzadeh, line 41: please put this whole code block inside an if statement for error handling purposes, like this:

```
if (monitor) {
// other if(s)
} else {
// report that no monitors are found
}
```

Server/models/Check.js
```diff
@@ -31,4 +31,27 @@ const CheckSchema = mongoose.Schema(
   }
 );
 
+/**
+ * When a check is created, we should update the stauts of the associated monitor
```
MuhammadKhalilzadeh, line 35: I'm trying to learn more and more about documentation these days and try to take advantage of AI and Chat-GPT tools for such purposes. So, here's what I got as another suggestion of comments and documentation for this code snippet:

```
/**
 * Middleware function executed before saving a check.
 * Updates the status of the associated monitor if it has changed.
 *
 * @param {function} next - Callback to proceed to the next middleware.
 */
```

What do you think?
Can we replace it with this one?

Server/models/Check.js
MuhammadKhalilzadeh: Yeah. its a function:  `next()`

MuhammadKhalilzadeh (approved):
@ajhollid  You've done a cool job as always buddy. 👍😊
I hope my comments do help to improve ✌️

Server/models/Check.js
ajhollid: AI can be a helpful tool, but I prefer to write my own documentation.  For example in this case it has missed out on the fact that the default value of the monitor status is true, which is useful information for anyone working on this method.

The js DOC is good though, we can include that :+1:

Server/models/Check.js
ajhollid: Good catch, typing too fast :P

Server/models/Check.js
ajhollid: Good call, I've added some error handling

Server/models/Check.js
MuhammadKhalilzadeh: Yeah. We can combine them and improve it 👍🏻

PR #255, opened 2024-07-03
Replace job in queue when monitor is edited, resolves #254

### LINKED ISSUE #254: Backend - When a monitor is edited, the job needs to be edited or restarted as well
Currently the job keeps running with the instance of the old monitor

ajhollid:


Skorpios604 (approved):

Server/controllers/monitorController.js
```diff
@@ -214,6 +214,10 @@ const editMonitor = async (req, res, next) => {
 
   try {
     const editedMonitor = await req.db.editMonitor(req, res);
+    // Delete the old job(editedMonitor has the same ID as the old monitor)
+    req.jobQueue.deleteJob(editedMonitor);
```
MuhammadKhalilzadeh, line 218: My suggestion is that we could use async/await more consistently throughout the function for cleaner code structure.

And also for a more specific error handling and haunting the error on a more exact spot, you may want to write this line like this:

```
// Delete the old job from the job queue
    try {
      await req.jobQueue.deleteJob(editedMonitor);
    } catch (error) {
      // Log the error and continue, as the monitor was still successfully edited
      console.error(some constant text to show the error type , error);
    }
```

Server/controllers/monitorController.js
```diff
@@ -214,6 +214,10 @@ const editMonitor = async (req, res, next) => {
...
+    // Add the new job back to the queue
+    req.jobQueue.addJob(editedMonitor._id, editedMonitor);
```
MuhammadKhalilzadeh, line 220: Similarly, we could this here like this:

```
// Add the new job back to the queue
    try {
      await req.jobQueue.addJob(editedMonitor._id, editedMonitor);
    } catch (error) {
      // Log the error and continue, as the monitor was still successfully edited
      console.error('Failed to add job to queue:', error);
    }
```

MuhammadKhalilzadeh (approved):
Looks great to me ✅✅
Adding some small details can make it better organized and cleaner.

Server/controllers/monitorController.js
ajhollid: Good call, I forgot this was an async function, I will add the await.

The error handling is done further upstream in the express middleware, so throwing the error is sufficient here and the middleware will take care of it.

MuhammadKhalilzadeh: Awesome 👍🏻✌️✅

Server/controllers/monitorController.js
ajhollid: If we have a separate try/catch block for each async call in the function then if any one call fails, the others will still execute. 

That would lead to jobs possibly being duplicated or deleted and not added back in.  

```
const doOperation = () => {
  try {
    // delete old job
  } catch (error) {
    console.log(error);
  }
  try {
    // add new job
  } catch (error) {
    console.log(error);
  }
  ```
  
If we do this and deleting the job fails, we will still add the job again, and now we've got duplicate jobs, or we'll throw a different error all togehter.
  
 What we want in this case is one try/catch block with all the risky operations in the try block, so if any of them fails, the whole thing fails

MuhammadKhalilzadeh: Oh, I thought the error in inner try/catch propagates and makes others fail as well and makes finding errors more detailed!
Cool then 👍🏻

ajhollid: It does allow us to do more detailed error handling, but control returns to the calling function after the catch block and the code continues to run.

In cases where we do need to do something specific with the error handling we can catch the error, do whatever we need to do with it, and then throw it again.

That will short circuit the function as intended and allow the middleware to handle the error upstream.

I don't think it's necessary here though, the error itself will already contain enough information for debugging I believe.

The client should not receive errors from the JobQueue, these errors are for our debugging use only I believe.

PR #257, opened 2024-07-04
Backend completion

MuhammadKhalilzadeh:
I just added `verifyOwnership` `middleware` in `authRoute.js` for the **delete user endpoint** by looking at how it is done in `monitorRoute.js`.
Please review it to see if there is anything needs to be added or fixed.


Server/controllers/authController.js
```diff
@@ -261,6 +262,73 @@ const resetPasswordController = async (req, res, next) => {
...
+ * @returns {Object} The response object with success status and message.
+ * @throws {Error} If user validation fails or user is not found in the database.
+ */
+const deleteUserController = async (req, res, next) => {
+  try {
+    // Validate user
+    await deleteUserParamValidation.validateAsync(req.body);
+
+    // Check if the user exists
+    const user = await req.db.getUserByEmail(req);
```
ajhollid, line 281: There's a bug here, what happens if a user uses their token maliciously and adds a different user's email to their request?

The malicious use will be able to delete any other user's monitors. 

I think to resolve this we need to extract the user's ID and Email from their token rather than allowing it to be passed in the body or request params

ajhollid (changes requested):
There's a major bug here that would allow a malicious user to delete other user's data.  I believe we need to use the JWT to fetch user ID and email to resolve

Server/controllers/authController.js
MuhammadKhalilzadeh: So, what do you have in mind for it to added?
Could you please explain your idea a bit more with a code snippet?

ajhollid: Sure, when the user hits the `delete` endpoint they must have a valid JWT token.  This is included in the headers:

```
{
  host: 'localhost:5000',
  'user-agent': 'insomnia/2023.5.8',
  authorization: 'Bearer eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJfaWQiOiI2NjdkZWQxOGFlNjExNDFlNjg5NzQ0NjQiLCJmaXJzdG5hbWUiOiJBbGV4IiwibGFzdG5hbWUiOiJIb2xsaWRheSIsImVtYWlsIjoidGVzdEB0ZXN0LmNvbSIsImlzQWN0aXZlIjp0cnVlLCJpc1ZlcmlmaWVkIjpmYWxzZSwiY3JlYXRlZEF0IjoiMjAyNC0wNi0yN1QyMjo1MjowOC44MDBaIiwidXBkYXRlZEF0IjoiMjAyNC0wNi0yN1QyMjo1MjowOC44MDBaIiwiX192IjowLCJpYXQiOjE3MTk1Mjg5OTUsImV4cCI6MTcyODA4MjU5NX0.FBd_TmyCfMHKpLm_1cyzgcpeDLIZaT14XkkNaOka1LQ',
  accept: '*/*'
}
```

The JWT token contains both the user ID and the user's email.  Decoded JWT:

```
{
  "_id": "667ded18ae61141e68974464",
  "firstname": "Alex",
  "lastname": "Holliday",
  "email": "test@test.com",
  "isActive": true,
  "isVerified": false,
  "createdAt": "2024-06-27T22:52:08.800Z",
  "updatedAt": "2024-06-27T22:52:08.800Z",
  "__v": 0,
  "iat": 1719528995,
  "exp": 1728082595
}
```

So my suggestion is that we remove all parameters and the request body from this request and obtain all information we need from the JWT.

That way a user cannot maliciously affect another user's data unless they have stolen their token.  That's beyond the scope of our concern though, there's not much we can do about that other than set very short expirty times for the token.

In summary, my suggestion is:

1.  Remove the request parameter and request body
2.  Get the JWT token from `req.headers`
3.  Decode the JWT and extract the user ID and email.  We already have the `jsonwebtoken` library included for decoding tokens, see `verifyJWT` middleware for usage example
4.  Continue as normal for the rest of the process.

I think this gives us the greatest amount of security for a very sensitive endpoint

MuhammadKhalilzadeh: Oh, got it 👍🏻
I've seen this before!

MuhammadKhalilzadeh: I tried to implement your idea.
Check to see if that is what you mean

Server/controllers/authController.js
```diff
@@ -261,6 +262,88 @@ const resetPasswordController = async (req, res, next) => {
...
+ *
+ * @param {Object} req - The request object.
+ * @param {Object} res - The response object.
+ * @param {Function} next - The next middleware function.
+ * @returns {Object} The response object with success status and message.
+ * @throws {Error} If user validation fails or user is not found in the database.
+ */
+const deleteUserController = async (req, res, next) => {
+  try {
+    const token = req.headers.authorization.split(' ')[1];
```
ajhollid, line 277: This will throw an erorr if the auth header isn't present, we can leave it for now but we should open an issue to deal with the potential errors here

Server/controllers/authController.js
```diff
@@ -261,6 +262,88 @@ const resetPasswordController = async (req, res, next) => {
...
+    const decodedToken = jwt.decode(token)
```
ajhollid, line 278: I was checking the docs and `jwt.decode` actually doesn't verify, so we need to change this to `jwt.verify`, otherwise a user can just create their own token and delete whatever they want with it

Server/controllers/authController.js
```diff
@@ -261,6 +262,88 @@ const resetPasswordController = async (req, res, next) => {
...
+    console.log("USERID: ", _id, " Email: ", email)
+
+    const decodedTokenCastedAsRequest = {
+      params: {
+        userId: _id
+      },
+      body: {
+        email
+      }
+    }
```
ajhollid, line 290: I don't think there's any reason to make a fake request here?  We should just refactor functions that originally wanted a request to now accept an `id` and `email`.

Otherwise this is potentially confusing for new devs.

Let's leave this for now and I'll do a refactor later on.

Server/routes/authRoute.js
```diff
@@ -23,6 +26,7 @@ router.post(
   userEditController
 );
 router.get("/users/admin", checkAdminController);
+router.delete("/user/:userId", verifyJWT, verifyOwnership(User, "userId") , deleteUserController);
```
ajhollid, line 29: We can remove the `:userId` param from this route, it is no longer required

`router.delete("/user", verifyJWT, verifyOwnership(User, "userId") , deleteUserController);`

is sufficient

ajhollid (changes requested):
Looks pretty good, there are a couple TODOs that we can take care of in the next PR.

We do need to switch from `jwt.decode` to `jwt.verify` to actually verify the JWT though, and the route needs to be updated to remove the parameter

Server/routes/authRoute.js
MuhammadKhalilzadeh: Fixed

MuhammadKhalilzadeh: Let's merge this thing and then figure out what to do next

MuhammadKhalilzadeh: I'm facing a couple of errors now. please text login, register, and delete to check them out

ajhollid: Can you elaborate more on the errors?

How do we replicate them?

MuhammadKhalilzadeh: Don't need to do anything!
Just do make a simple request (Register, Login, and Delete) (On both UI and postman) and errors will be logged on console
The error is for parts that I haven't worked on before, so I'm afraid modifying the code blindly may cause issues.

ajhollid (approved):

PR #265, opened 2024-07-04
Feat/recharts, resolves #264

### LINKED ISSUE #264: Frontend - Monitor chart library
Use a chart library for the monitors page.  Options include:

- [Recharts](https://recharts.org/en-US/)
- [React Charts](https://react-charts.tanstack.com/)
- [Nivo](https://nivo.rocks/)
- [Apex](https://apexcharts.com/)

I was originally leaning towards Apex, but I'm thinking Recharts now since it's already implemented in React.  Apex requires a react wrapper

ajhollid:
- Refactored the Monitors page to use Recharts for charting data.  Native react and responsive
- Improved the normalization of data by adding a floor and ceiling to the data
    - Currently using 5th and 95th percentiles as floor and ceiling.  We can adjust these values if we don't like the display

image


Skorpios604 (approved):
Looks great! I like these bars much better than the previous ones.

PR #278, opened 2024-07-06
Front end small fix v1

MuhammadKhalilzadeh:


ajhollid (changes requested):
This pull request still has all the backend changes in it.

Can you open a PR with just the small frontend fixes?

PR #286, opened 2024-07-07
Updated navbar user dropdown, resolves #224

### LINKED ISSUE #224: Frontend - Update navBar user dropdown
Update user dropdown as per the Figma design

image

- Add icons
- Update links
- Add navigation

danielcj2:
Updated user dropdown to match Figma design

- Added icons.
- Updated links.
- Added navigation.


Client/package.json
```diff
@@ -32,7 +32,8 @@
     "react-router-dom": "^6.23.1",
     "react-toastify": "^10.0.5",
     "recharts": "2.12.7",
-    "redux-persist": "6.0.0"
+    "redux-persist": "6.0.0",
+    "vite-plugin-svgr": "^4.2.0"
```
ajhollid, line 36: Is there a signifcant benefit to usign this plugin rather than an `img` with the svg as the source?

I'm always of the opinion that the fewer dependencies a project has the better, but if there is a real benefit then disregard

ajhollid (approved):
Good on my end.  My only thought is I'm not sure exactly what benefit the svg plugin provides.

Skorpios604 (approved):
Looks good!

Client/package.json
danielcj2: I tend to agree with you but `vite-plugin-svgr` is a super lightweight tool. Reason why I added it is for ease of maintenance in case we ever do decide to mess around with the svgs or import more non @mui icons. Makes it easier to edit within the code, rather than having to locate the files and update them separately. I believe svgs also offer better performance, although I could be wrong.

PR #287, opened 2024-07-08
Feat/m UI table and pagination

ajhollid:
Improved performance

- Uses regular MUI table instead of Datagrid
- Pagination _on frontend_ implemented.  
  - We can think about pagination on the backend too as datasets grow larger


Skorpios604 (approved):
Nice stuff!

danielcj2 (approved):
Looks really clean!

I'm not so sure about the `box shadow` at the bottom of the table but that's more of a design concern.

PR #289, opened 2024-07-08
Limit checks

MuhammadKhalilzadeh:
`checkController.js` and `MongoDB.js` : Limiting getChecks to only fetch the last 25 checks to display on the bar chart.
`checkController.js` : Error handling
`messages.js` : New message constant, `CHECKS_NOT_FOUND`


ajhollid (changes requested):
Good starting point!

Error handling should not be done at the Controller level, we have middleware to do that. If you handle the error here we will not have system logs for the error.

The getChecks method needs a bit more flexibility to work as we need it, please see details in comments!

Server/controllers/checkController.js
```diff
@@ -44,9 +44,15 @@ const getChecks = async (req, res, next) => {
...
-    return res
-      .status(200)
-      .json({ success: true, msg: successMessages.CHECK_GET, data: checks });
+    if (checks) {
+      return res
+        .status(200)
+        .json({ success: true, msg: successMessages.CHECK_GET, data: checks });
+    } else {
+      return res
+        .status(404)
```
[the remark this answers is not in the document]
MuhammadKhalilzadeh: So I'll remove the error handling part from here then

Server/db/MongoDB.js
```diff
@@ -416,7 +416,7 @@ const createCheck = async (checkData) => {
 
 const getChecks = async (monitorId) => {
   try {
-    const checks = await Check.find({ monitorId });
+    const checks = await Check.find({ monitorId }).sort({ createdAt: -1 }).limit(25);
```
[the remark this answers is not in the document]
MuhammadKhalilzadeh: Yep. Option 1 is the best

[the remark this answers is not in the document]
MuhammadKhalilzadeh: Where do you want the parameter then? inside `req.body` or `req.params` or any other options?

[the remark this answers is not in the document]
ajhollid: Probably `req.query` i would imagine.  URL would look something like

`/api/v1/monitor/{monitor_id}?limit=25`

Server/db/MongoDB.js
```diff
@@ -414,9 +414,13 @@ const createCheck = async (checkData) => {
  * @throws {Error}
  */
 
-const getChecks = async (monitorId) => {
+const getChecks = async (monitorId, limitedTo) => {
```
MuhammadKhalilzadeh, line 417: @ajhollid Is this okay?

ajhollid: Sure, but I think you can name that var `limit`

MuhammadKhalilzadeh: Done.
how about other parts?

ajhollid: Have you updated the route to include the query param?  I don't think I saw that here

ajhollid: Frontend will also need to be updated to make the appropriate request, and docs need to be updated too

MuhammadKhalilzadeh:
@ajhollid  
Modifications are applied as discussed

ajhollid (approved):

PR #296, opened 2024-07-09
Monitors pagination styling, resolves #295

### LINKED ISSUE #295: Frontend - Monitor table pagination
Change current pagination to match Figma design.

image

danielcj2:
Updated monitors pagination to match Figma design.

### DIFF
--- a/Client/src/Components/MonitorTable/index.jsx
+++ b/Client/src/Components/MonitorTable/index.jsx
@@ -1,7 +1,6 @@
 import "./index.css";
 import { useState } from "react";
 import PropTypes from "prop-types";
-import OpenIt from "../../assets/Images/Icon-open-in-tab-gray.png";
 import ResponseTimeChart from "../Charts/ResponseTimeChart";
 import {
   Table,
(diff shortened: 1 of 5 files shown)


Client/src/Components/MonitorTable/index.jsx
```diff
@@ -93,20 +94,15 @@ const Status = ({ params }) => {
  * @returns {React.Component} Returns a table with the monitor data.
  */
 const MonitorTable = ({ monitors = [] }) => {
-  const [page, setPage] = useState(0);
-  const [rowsPerPage, setRowsPerPage] = useState(5);
+  const [page, setPage] = useState(1);
```
ajhollid, line 97: This is a bit confusing as the reference pagination internally uses 0 based indexing, but I assume this is offset to show the user 1 based indexing.

I think a less confusing solution would be to leave the actual implementation as 0 based and move the offset to the rendering:

```
const [page, setPage] = useState(0);
...
const mappedRows = monitors.slice(page * rowsPerPage, page * rowsPerPage + rowsPerPage)
````

And then in the pagination component:

```
page={page + 1}
onChange={(event, value) => handleChangePage(event, value - 1)}
```

This is clearer to me since the pagination component expects a 0 based index.

ajhollid (changes requested):
Looks great, much closer to the Figma designs.

I think there is a clearer solution for the 0 based vs 1 based indexing offset issue, please ee my comment and see what you think

Client/src/Components/MonitorTable/index.jsx
danielcj2: Yeah you're correct, it was done this way to simplify the logic calculating the offset, but I see where you're coming from. I'll add the fix 👍

ajhollid (approved):
Awesome, thanks for the update 👌

PR #300, opened 2024-07-09
Added configured axios instance with 401 interceptor, resolves #299

### LINKED ISSUE #299: Frontend - Set up interceptor for 401 responses to revoke token
If a user tries to make a request with an expired token their token should be revoked from the redux store and they should be redirected to login

ajhollid:
Added an axios instance with an interceptor for 401 requests.

If a 401 is returned by the server the auth state will be cleared and the user will be booted to the login screen


Skorpios604 (approved):
Looks great! The man with the plan

PR #301, opened 2024-07-09
Image deletion, resolves #297

### LINKED ISSUE #297: Frontend & Backend - Image deletion

danielcj2:
Finalized the frontend image deletion logic.

### DIFF
--- a/Client/src/Components/TabPanels/Account/ProfilePanel.jsx
+++ b/Client/src/Components/TabPanels/Account/ProfilePanel.jsx
@@ -72,6 +72,10 @@ const ProfilePanel = () => {
     );
     if (error) return;
 
+    setLocalData((prev) => ({
+      ...prev,
+      deleteProfileImage: false,
+    }));
     setProgress((prev) => ({ ...prev, isLoading: true }));
     setFile({
       src: URL.createObjectURL(pic),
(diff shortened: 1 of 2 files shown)


Client/src/Components/TabPanels/Account/ProfilePanel.jsx
```diff
@@ -116,13 +117,23 @@ const ProfilePanel = () => {
...
   const removePicture = () => {
     errors["picture"] && clearError("picture");
+    setFile({ delete: true });
     clearInterval(intervalRef.current); // interrupt interval if image upload is canceled prior to completing the process
     setProgress({ value: 0, isLoading: false });
   };
 
+  // Opens the picture update modal
+  const openPictureModal = () => {
+    setIsOpen("picture");
```
ajhollid, line 127: I see why this is done, we need to distinguish between which operation opened the modal, but the `isModalOpen` method is kind of confusing as a result :thinking: 

We might want to add a TODO here to revisit this and think about a cleaner way of handling this, but no big deal for now

ajhollid (approved):
Looks good to me, the only thing that stands out to me is the handling of the Modal.  We might want to revisit that in the future

Client/src/Components/TabPanels/Account/ProfilePanel.jsx
danielcj2: The alternative is we could drop `isModalOpen` entirely and use `isOpen === value` directly, if that makes it any less confusing. I'm open to suggestions. I wasn't really sure how else I could implement it. It's a little tricky because we're dealing with two modals that are dependent on one another in the sense that only one can be active at a time.

ajhollid: What about just setting the open value of one to be the inverse of the other?

That way they can never be open at the same time?

danielcj2: Wouldn't that just open the 2nd modal when you close the 1st and vice versa 😂 ?

ajhollid: I hearby retract my previous comment and slink away in shame 🤣

ajhollid: I think we can leave it for now, it's not critical. Let's keep it in mind though in case a better solution comes to us

Skorpios604 (approved):
Look clean and fully functional. Good to merge.

PR #302, opened 2024-07-09
Account deletion

danielcj2:
Connected backend to redux auth slice.

### DIFF
--- a/Client/src/Components/TabPanels/Account/ProfilePanel.jsx
+++ b/Client/src/Components/TabPanels/Account/ProfilePanel.jsx
@@ -12,11 +12,16 @@ import {
   imageValidation,
 } from "../../../Validation/validation";
 import { useDispatch, useSelector } from "react-redux";
-import { update } from "../../../Features/Auth/authSlice";
+import {
+  clearAuthState,
+  deleteUser,
+  update,
+} from "../../../Features/Auth/authSlice";
 import ImageField from "../../TextFields/Image";
 import ImageIcon from "@mui/icons-material/Image";
 import ProgressUpload from "../../ProgressBars";
 import { formatBytes } from "../../../Utils/fileUtils";
+import { clearMonitorState } from "../../../Features/Monitors/monitorsSlice";
 
 /**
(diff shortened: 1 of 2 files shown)


Client/src/Components/TabPanels/Account/ProfilePanel.jsx
```diff
@@ -179,7 +184,12 @@ const ProfilePanel = () => {
 
   // Initiates the account deletion process
   const handleDeleteAccount = () => {
-    //TODO - implement delete account function
+    dispatch(deleteUser(authToken)).then((action) => {
+      if (action.payload.success) {
+        dispatch(clearAuthState());
+        dispatch(clearMonitorState());
+      }
+    });
```
ajhollid, line 192: Can we replace this with async/await?  

```
const action = await dispatch(deleteUser(authToken))
if (action.payload.success {
    ...
}
```

I always aim for the least amount of indentaiton possible

Client/src/Features/Auth/authSlice.js
```diff
@@ -74,6 +75,28 @@ export const update = createAsyncThunk(
   }
 );
 
+export const deleteUser = createAsyncThunk(
+  "auth/delete",
+  async (data, thunkApi) => {
+    const user = jwtDecode(data);
+
+    //1.5s delay to show loading spinner
+    await new Promise((resolve) => setTimeout(resolve, 1500));
```
ajhollid, line 84: Is this just for demonstration purposes?  I don't think we need to slow down operations purposefully when we're ready to ship production

ajhollid (changes requested):
Nice work :+1:  Unless there's an issue with using `await` I think we should use that instead of `Promise.then()` for the dispatches to keep things consistent and as readable as possible

Client/src/Features/Auth/authSlice.js
danielcj2: Yea pretty much, without the timeout the operation is instant. The timeout is there to tell the user _hey, a process is occurring_. I would image that with larger datasets there won't be a need for it at all.

ajhollid (approved):

Issue #120, opened 2024-06-10
Front End - Connect `Display One Monitor` with real data


ajhollid:
All data from one monitor should be requested from backend when a single monitor's details are viewed

PR #303, opened 2024-07-10
Update route to pass monitor ID to details component, #120

ajhollid:
Clicking on a monitor from the list now takes you to the details page with the monitor ID passed along


Skorpios604 (approved):
Works good

PR #311, opened 2024-07-10
Connected password form to redux, resolves #214

### LINKED ISSUE #214: Frontend - Connect profile settings password form to backend

danielcj2:
Connected password form to redux. There are sill some checks to be done. Currently any error from the request will display under the `Password` text field, where we're specifically looking for the "Incorrect password" (403) error only. The easy solution would be to simply check for `action.payload.msg === "Incorrect password"` but that doesn't feel right. Any thoughts?

### DIFF
--- a/Client/src/Components/TabPanels/Account/PasswordPanel.jsx
+++ b/Client/src/Components/TabPanels/Account/PasswordPanel.jsx
@@ -6,6 +6,8 @@ import ButtonSpinner from "../../ButtonSpinner";
 import PasswordTextField from "../../TextFields/Password/PasswordTextField";
 import { editPasswordValidation } from "../../../Validation/validation";
 import Alert from "../../Alert";
+import { update } from "../../../Features/Auth/authSlice";
+import { useDispatch, useSelector } from "react-redux";
 
 /**
  * PasswordPanel component manages the form for editing password.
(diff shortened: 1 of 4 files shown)


Client/src/Components/TabPanels/Account/PasswordPanel.jsx
```diff
@@ -44,7 +46,7 @@ const PasswordPanel = () => {
 
     const validation = editPasswordValidation.validate(
       { [name]: value },
-      { abortEarly: false, context: { newpassword: localData.newpassword } }
+      { abortEarly: false, context: { newPassword: localData.newPassword } }
```
ajhollid, line 49: I just realized I totally ignored our current naming convention here 😂 these body params probably should all be camel case instead of all lower case..
ie firstName instead of firstname 🤔thoughts? Worth changing?

ajhollid (approved):
Everything looks good except for my ignoring the current naming convention 😂

Client/src/Components/TabPanels/Account/PasswordPanel.jsx
danielcj2: True, camel case does seem more appropriate. It's up to you, Captain. You say jump, and I'm diving overboard 😄.

ajhollid: 🏴‍☠️ Off you go then! Let's leave it for now and I'll open an issue and we can do this refactor when we have a chance

Skorpios604 (approved):

PR #318, opened 2024-07-11
Status table component

Skorpios604:
Made a component for the status table since it will be used in many places in the app. 

## Summary by CodeRabbit

- **New Features**
  - Introduced a paginated table display with the new `StatusTable` component.
  
- **Refactor**
  - Replaced custom table rendering logic in the Incidents page with the new `StatusTable` component for improved data display and management.

### DIFF
--- a/Client/src/Components/StatusTable/index.jsx
+++ b/Client/src/Components/StatusTable/index.jsx
@@ -1,4 +1,3 @@
-import * as React from 'react';
 import { styled, useTheme } from '@mui/material/styles';
 import Table from '@mui/material/Table';
 import TableBody from '@mui/material/TableBody';
@@ -8,7 +7,6 @@ import TableHead from '@mui/material/TableHead';
 import TableRow from '@mui/material/TableRow';
 import Paper from '@mui/material/Paper';
 import Box from '@mui/material/Box';
-import Typography from '@mui/material/Typography';
 import { StatusLabel } from "../../Components/Label/";
 
 /**
--- a/Client/src/Components/StatusTable/index.jsx
+++ b/Client/src/Components/StatusTable/index.jsx
@@ -81,8 +81,6 @@ export default function IncidentHistoryTable() {
         },
       }}
     >
-      <Typography variant="h6" component="div" sx={{ fontWeight: 600, fontSize: 16, marginBottom: theme.spacing(2) }}>
-      </Typography>
       <TableContainer component={Paper}>
         <Table sx={{ minWidth: 700 }} aria-label="customized table">
           <TableHead>
--- a/Client/src/Components/StatusTable/index.jsx
+++ b/Client/src/Components/StatusTable/index.jsx
@@ -82,7 +82,6 @@ export default function IncidentHistoryTable() {
       }}
     >
       <Typography variant="h6" component="div" sx={{ fontWeight: 600, fontSize: 16, marginBottom: theme.spacing(2) }}>
-        Incident History
       </Typography>
       <TableContainer component={Paper}>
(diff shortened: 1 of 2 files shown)


Client/src/Components/StatusTable/index.jsx
```diff
@@ -0,0 +1,114 @@
...
+  '&:last-child td, &:last-child th': {
+    border: 0,
+  },
+}));
+
+const itemsPerPage = 5; // Number of items per page
+
+export default function StatusTable({ data, columns }) {
+  const theme = useTheme();
+  const [currentPage, setCurrentPage] = React.useState(1);
```
ajhollid, line 40: Can you touch base with @danielcj2 on this one?  We went over 0 vs 1 based indexing in one of his PRs, it makes more sense to follow a 0 based index since that's what the pagination uses internally

Client/src/Components/StatusTable/index.jsx
```diff
@@ -0,0 +1,114 @@
...
+          width: '100%',
+        },
+      }}
+    >
+      <TableContainer component={Paper}>
+        <Table sx={{ minWidth: 700 }} aria-label="customized table">
+          <TableHead>
+            <TableRow>
+              {columns.map((column, index) => (
+                <StyledTableCell key={index} align={column.align || 'left'}>
```
ajhollid, line 69: This should use `column.id` as the key

Client/src/Components/StatusTable/index.jsx
```diff
@@ -0,0 +1,114 @@
...
+  },
+}));
+
+const StyledTableRow = styled(TableRow)(({ theme }) => ({
+  '&:nth-of-type(odd)': {
+    backgroundColor: theme.palette.action.hover,
+  },
+  '&:nth-of-type(even)': {
+    backgroundColor: theme.palette.common.white,
+  },
```
ajhollid, line 30: These tables are not supposed to have alternating row colors are they?

Client/src/Components/StatusTable/index.jsx
```diff
@@ -0,0 +1,114 @@
...
+        width: '80%',
+        paddingX: theme.spacing(6),
+        [theme.breakpoints.down('md')]: {
+          width: '100%',
+        },
+      }}
+    >
+      <TableContainer component={Paper}>
+        <Table sx={{ minWidth: 700 }} aria-label="customized table">
+          <TableHead>
```
ajhollid, line 66: Table header height looks off to me, it should be 43px, shorter than the other rows 

image

Client/src/Components/StatusTable/index.jsx
```diff
@@ -0,0 +1,114 @@
...
+                <StyledTableCell key={index} align={column.align || 'left'}>
+                  {column.header}
+                </StyledTableCell>
+              ))}
+            </TableRow>
+          </TableHead>
+          <TableBody>
+            {paginatedData.map((row, rowIndex) => (
+              <StyledTableRow key={rowIndex}>
+                {columns.map((column, colIndex) => (
```
ajhollid, line 78: This is an O(n*m) operation, we should find a way to simlplify this as this could get really nasty if we have a lot of data.

ajhollid (changes requested):
The table looks good!

There's some styling issues to address and I think we need to sit down and plan out the best way to have a generic table render its rows from a chunk of data.

If you could start on the styling issues I'll put my head to thinking about the best way to get those rows rendering in a performant and reusable way!

ajhollid (approved):
Discussed changes with @Skorpios604, we're going to incorpoate changes into v2 of this

Issue #310, opened 2024-07-10
Validation - Password validation needs to be synced between front and backend


ajhollid:
Password validation on the frontend and backend do not currently match
 - Backend is required and min 8 chars
 - Frontend is required, min 8 chars, at least one upper case, at least one number, at least one symbol

PR #319, opened 2024-07-11
Validation - Password validation on registration page-Frontend #310

vibhu1604:
password validation - Must be at least 8 characters, Must contain one special character, one number and one uppercase

## Summary by CodeRabbit

- **New Features**
  - Enhanced password validation on the registration page to require at least one uppercase letter, one number, and one special character in addition to the existing minimum length requirement.

### DIFF
--- a/Client/src/Pages/Register/index.jsx
+++ b/Client/src/Pages/Register/index.jsx
@@ -150,7 +150,7 @@ const Register = () => {
         <div className="register-form-checks">
           <Check text="Must be at least 8 characters" />
           <div className="register-form-v-spacing-small"></div>
-          <Check text="Must contain one special character" />
+          <Check text="Must contain one special character, one number and one uppercase" />
         </div>
         <div className="login-form-v2-spacing" />
         <div className="register-form-actions">
--- a/Client/src/Validation/validation.js
+++ b/Client/src/Validation/validation.js
@@ -21,9 +21,7 @@ const registerValidation = joi.object({
   password: joi
     .string()
     .min(8)
-    .pattern(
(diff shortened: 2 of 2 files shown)


Client/src/Validation/validation.js
```diff
@@ -18,10 +18,17 @@ const registerValidation = joi.object({
...
     }),
 
-  password: joi.string().min(8).required().messages({
-    "string.min": "Password must be at least 8 characters",
-    "string.empty": "Password is required",
-  }),
+  password: joi
+    .string()
+    .min(8)
+    .pattern(/^(?=.*[A-Z])(?=.*\d)(?=.*[@$!%*?&])[A-Za-z\d@$!%*?&]{8,}$/)
```
ajhollid, line 24: As I mentioned in your other PR we can drop the 8 char requirement from the regex since it is already checked for by `.min(8)`.  No need for the redundant check

Client/src/Pages/Register/index.jsx
```diff
@@ -150,7 +150,7 @@ const Register = () => {
         <div className="register-form-checks">
           <Check text="Must be at least 8 characters" />
           <div className="register-form-v-spacing-small"></div>
-          <Check text="Must contain one special character" />
+          <Check text="Must contain one special character, one number and one uppercase" />
```
ajhollid, line 153: I would change this to:

"Must contain one special character, one number, and one uppercase letter"

I'm a stickler for the [oxford comma](https://en.wikipedia.org/wiki/Serial_comma) and the last "letter" is needed

ajhollid (changes requested):

ajhollid:
Revisit when @vibhu1604 returns

Issue #240, opened 2024-07-02
Frontend - Monitors page redirect


danielcj2:
Monitors page should redirect to `/monitors/create` if the user has no monitors created.

PR #320, opened 2024-07-11
Feat/redirect newmonitor #240

vibhu1604:
If no monitor then redirected to create monitor page

## Summary by CodeRabbit

- **New Features**
	- Added automatic redirection to the "/monitors/create" route if no monitors are present.

- **Enhancements**
	- Updated password requirements in the registration form to include one uppercase letter, one number, and one special character.

- **Bug Fixes**
	- Improved password validation logic to enforce new pattern requirements for enhanced security.

### DIFF
--- a/Client/src/Pages/Register/index.jsx
+++ b/Client/src/Pages/Register/index.jsx
@@ -150,7 +150,7 @@ const Register = () => {
         <div className="register-form-checks">
           <Check text="Must be at least 8 characters" />
           <div className="register-form-v-spacing-small"></div>
-          <Check text="Must contain one special character" />
+          <Check text="Must contain one special character, one number and one uppercase" />
         </div>
         <div className="login-form-v2-spacing" />
         <div className="register-form-actions">
--- a/Client/src/Validation/validation.js
+++ b/Client/src/Validation/validation.js
@@ -18,10 +18,19 @@ const registerValidation = joi.object({
       "string.empty": "Email is required",
     }),
 
-  password: joi.string().min(8).required().messages({
-    "string.min": "Password must be at least 8 characters",
-    "string.empty": "Password is required",
-  }),
+  password: joi
+    .string()
+    .min(8)
+    .pattern(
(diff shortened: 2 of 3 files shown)


Client/src/Validation/validation.js
```diff
@@ -18,10 +18,19 @@ const registerValidation = joi.object({
...
 
-  password: joi.string().min(8).required().messages({
-    "string.min": "Password must be at least 8 characters",
-    "string.empty": "Password is required",
-  }),
+  password: joi
+    .string()
+    .min(8)
+    .pattern(
+      new RegExp("^(?=.*[A-Z])(?=.*\\d)(?=.*[@$!%*?&])[A-Za-z\\d@$!%*?&]{8,}$")
```
ajhollid, line 25: This regex has redundancy in it, we do not need to check for 8 char length, joi is already doing that, it can be removed for clarity

Client/src/Pages/Register/index.jsx
```diff
@@ -150,7 +150,7 @@ const Register = () => {
         <div className="register-form-checks">
           <Check text="Must be at least 8 characters" />
           <div className="register-form-v-spacing-small"></div>
-          <Check text="Must contain one special character" />
+          <Check text="Must contain one special character, one number and one uppercase" />
```
ajhollid, line 153: This should be in a separate pull request as it is resoving a different issue.

The rule of thumb is one issue per pull request, otherwise if we have to revert the merge for one of the issues we have to revert both of them.

ajhollid (changes requested):
Please keep your pull requests limited to one issue per request. 

In this case you can separate out the validation PR into it's own request and the Monitor redirect into it's own request :+1:

ajhollid:
After looking at your other PR I think I see what happened, it looks like you started working on your validation branch, pushed that, and then checked out a new branch for monitor redirect from your validation branch.

That means that you have included the changes from your validation branch in your monitor redirect branch, which is not what we want.

When you are working on more than one feature you should check out master first before starting a new feature branch.

ajhollid:
Revist when @vibhu1604  returns

PR #321, opened 2024-07-12
Product logo added

MuhammadKhalilzadeh:
## Summary by CodeRabbit

- **New Features**
  - Updated the NavBar component to display a new logo image.
  - Improved the Login page by updating the logo to a new image file.
- **Style**
  - Enhanced the NavBar and Login page styling for better logo display and spacing.


Client/src/Components/NavBar/index.css
```diff
@@ -30,3 +30,9 @@
   height: 16px;
   color: var(--env-var-color-25);
 }
+
+#bw-uptime-logo-dashboard {
+  height: 16px;
+  float: left;
+  margin-left: 15px;
```
danielcj2, line 37: The float is not needed here.

Client/src/Components/NavBar/index.jsx
```diff
@@ -103,22 +105,11 @@ function NavBar() {
...
-              textDecoration: "none",
-            }}
-          >
-            Peak Watch
-          </Typography>
+          <img
+            id="bw-uptime-logo-dashboard"
+            src={BWULogo}
+            alt="BlueWave Uptime Logo"
+          />
```
danielcj2, line 112: Please use svg instead of a regular img. Svgs are faster to load, scale better and you can also use inline css styles to edit it. Check `navBar` component to see implementation.

Client/src/Pages/Login/index.css
```diff
@@ -18,7 +18,9 @@
 }
 
 .login-form-header-logo {
-  margin: 0 auto;
+  margin: 30px auto;
+  margin-bottom: 60px;
+  width: 353px;
```
danielcj2, line 23: I don't believe there is a need for `auto` here. I'm not so sure about the margins either, we're only really concerned about spacing the logo and the login header apart so a margin-bottom should suffice.

danielcj2 (changes requested):
Looks good, just a couple issues I've noticed.

Client/src/Components/NavBar/index.css
ajhollid: Yes, this doesn't seem to actually do anything?

Client/src/Components/NavBar/index.jsx
```diff
@@ -23,6 +23,8 @@ import UserSvg from "../../assets/icons/user.svg?react";
 import TeamSvg from "../../assets/icons/user-two.svg?react";
 import LogoutSvg from "../../assets/icons/logout.svg?react";
 
+import BWULogo from "../../assets/Images/bwl-logo.svg";
```
ajhollid, line 26: Nice, great to see the Logo up :partying_face:

ajhollid (changes requested):
Really good to see the Logo up and displaying :100: There seem to be some unnecessary styles, and the SVG can use the SVG tag instead of img tag :+1:

Client/src/Components/NavBar/index.css
MuhammadKhalilzadeh: Yep. It was just a test. Forgot to remove!

Client/src/Pages/Login/index.css
MuhammadKhalilzadeh: In fact, auto is what's helping to make the Logo be centered.
If I remove it, logo will be right aligned!

MuhammadKhalilzadeh: I added margins to make it a bit more clear and clean, since the "Log in to your account" heading is too large and it wouldn't look too nice without these margins.
I suggest keeping it for now. If we did not like it, we'll remove it later.

ajhollid: I've tested as well, the auto isn't required as the parent has text-align center set on it

ajhollid (approved):

danielcj2 (approved):
Looks good

PR #327, opened 2024-07-12
Added toast alerts to login/register pages, resolves #147

### LINKED ISSUE #147: Frontend - Backend - Use Toast component to display errors
Currently errors on login and registration are shown using `alert` as a placeholder, these can all be replaced with the new `Toast` component.

danielcj2:
Added toast alerts to login/register pages and move `ToastContainer` to top of the page so the alerts can be displayed everywhere in the app.

### DIFF
--- a/Client/src/Pages/Login/index.jsx
+++ b/Client/src/Pages/Login/index.jsx
@@ -78,7 +78,10 @@ const Login = () => {
         // validation errors
         createToast({
           variant: "info",
-          body: error.details[0].message,
+          body:
+            error && error.details && error.details.length > 0
+              ? error.details[0].message
+              : "Error validating data.",
           hasIcon: false,
         });
       } else if (error.response) {
--- a/Client/src/Pages/Login/index.jsx
+++ b/Client/src/Pages/Login/index.jsx
@@ -13,6 +13,7 @@ import axiosInstance from "../../Utils/axiosConfig";
 import { loginValidation } from "../../Validation/validation";
 import { login } from "../../Features/Auth/authSlice";
 import { useDispatch } from "react-redux";
+import { createToast } from "../../Utils/toastUtils";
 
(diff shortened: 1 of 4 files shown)


Client/src/Pages/Register/index.jsx
```diff
@@ -91,16 +92,26 @@ const Register = () => {
       }
     } catch (error) {
       if (error.name === "ValidationError") {
-        // TODO Handle validation errors
-        console.log(error);
-        alert(error);
+        // validation errors
+        createToast({
+          variant: "info",
+          body: error.details[0].message,
```
ajhollid, line 98: We should do a check here before trying to access items from the `details` array, it's not guaranteed to be populated

ajhollid (changes requested):
Looks good, we should do a check before accessing details array though for safety's sake

ajhollid (approved):

Client/src/Pages/Login/index.jsx
```diff
@@ -78,7 +78,10 @@ const Login = () => {
         // validation errors
         createToast({
           variant: "info",
-          body: error.details[0].message,
+          body:
+            error && error.details && error.details.length > 0
+              ? error.details[0].message
+              : "Error validating data.",
```
danielcj2, line 84: oh wait I don't think error.details.length > 0 is relevant here

ajhollid: woops, speed merged it, go ahead and open a fix PR

danielcj2: Disregard this, I just had a brain fart 🤣 Length will still be > than 0 when we're accessing the 0 index.

ajhollid: Good ol off by 1 :stuck_out_tongue:

PR #330, opened 2024-07-12
Implemented a basic table that takes a standard arrangement of data

ajhollid:
This table builds of @Skorpios604 's Status Table to build a generic table that can display any number of columns and rows

**TODO Currently Unstyled**

Table takes input arrays of objects in this format:

**headers**
|Name| Type|Note|
|---|---|---|
|id|number|Unique ID|
|name|string|Header name|

Example: 
```
headers: [
    {id:1, name: "Col 1},
    {id:2, name: "Col 2}
]
```

**rowItems**

|Name| Type|Note|
|---|---|---|
|id|number| Unique ID|
|data|React Node| Any valid react node|

Example: 
```
rowItems  = [
    {id: 1, data: <span>Row 1, col 1</span>},
    {id: 2, data: <span>Row 1, col 2</span>},
    {id: 3, data: <span>Row 2, col 1</span>},
    {id: 4, data: <span>Row 2, col 2</span>}
]
```

Rows will be arranged under their respective columns according to the number of headers.

### DIFF
--- a/Client/src/Components/BasicTable/index.jsx
+++ b/Client/src/Components/BasicTable/index.jsx
@@ -1,0 +1,146 @@
+import {
+  TableContainer,
+  Paper,
+  Table,
+  TableHead,
+  TableRow,
+  TableCell,
+  TableBody,
+  TablePagination,
+} from "@mui/material";
+import React, { useState } from "react";
+import PropTypes from "prop-types";
+
+/**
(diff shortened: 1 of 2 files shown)


ajhollid:
See monitors page for usage example

Skorpios604 (approved):

PR #331, opened 2024-07-13
Refactored textfield components into one

danielcj2:
FYI there will be some prop types error popping up here and there but those will all be addressed on my next PR.


Skorpios604 (approved):
Nice refactors! Tons of cool new features too!

PR #336, opened 2024-07-15
Made a schema for the metrics we need from Google Lighthouse 

Skorpios604:


Server/models/Metrics.js
```diff
@@ -0,0 +1,52 @@
...
+
+const MetricsSchema = mongoose.Schema(
+  {
+    monitorId: {
+      type: mongoose.Schema.Types.ObjectId,
+      ref: "Monitor",
+      immutable: true,
+    },
+    accessibility: {
+      type: Object,
```
ajhollid, line 22: These can all be type number as all were interested in from these (for now) is the overall score. If we need more metrics we can create more fields later

Server/models/Metrics.js
```diff
@@ -0,0 +1,52 @@
...
+      required: true,
+    },
+    performance: {
+      type: Object,
+      required: true,
+    },
+    timestamp: {
+      type: Date,
+      default: Date.now,
+    },
```
ajhollid, line 40: We don't need this field as timestamps are enabled below

Server/models/Metrics.js
```diff
@@ -0,0 +1,52 @@
...
+ * @typedef {Object} MetricsSchema
+ * @property {mongoose.Schema.Types.ObjectId} monitorId - Reference to the Monitor model.
+ * @property {Object} accessibility - Metrics related to accessibility category.
+ * @property {Object} bestPractices - Metrics related to best practices category.
+ * @property {Object} seo - Metrics related to SEO category.
+ * @property {Object} performance - Metrics related to performance category.
+ * @property {Date} timestamp - Timestamp when the metrics were recorded.
+ */
+
+const MetricsSchema = mongoose.Schema(
```
ajhollid, line 14: Why don't we name this PageSpeedCheck as that describes exactly what this object is

ajhollid (changes requested):
We can drop all the docker stuff from this PR, those files are leftover from an older iteration of the project and should be deleted.

The type of the four categories in the schema can be number, since for now at least all were interested in from the four categories is score. We can add more fields later if needed.

Overall good start!

Skorpios604:
Addressed issues.

ajhollid (approved):

PR #341, opened 2024-07-16
Feat/#293 maintenance window options

terryjung0244:
1. Fix Autocomplete value comparison in Dropdown component

- Updated isOptionEqualToValue prop in Dropdown component to correctly compare option and value objects by their 'name' property.
- Adjusted PropTypes for the value prop to accept an object instead of a string to match the structure of the options.

2. Add repeatComponents object with repeat option content.

3. Add state management for maintenance options and selected repeat option.

Please review my code to ensure I'm heading in the right direction. Thank you.


Client/src/Pages/Maintenance/CreateNewMaintenance.jsx
```diff
@@ -0,0 +1,45 @@
...
+    setSelectedOption(newValue);
+  };
+
+  return (
+    <div>
+      <div>Create a maintenance window</div>
+      <div>Your pings won&apos;t be sent in this time frame.</div>
+      <div>Repeat</div>
+      <div>
+        <Dropdown
```
ajhollid, line 32: @terryjung0244 We discussed using the MUI `Select` component here, was there some difficulty with the implementation?

Client/src/Pages/Maintenance/index.css
```diff
@@ -33,4 +33,5 @@
 
 .maintenance-checklist-main .maintenance-checklist-button {
   margin-top: 30px;
+  font-size: 13px;
```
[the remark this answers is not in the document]
ajhollid: @terryjung0244 this is good advice, we should avoid hardcoding values, especially those that are used often and in many places.

Font sizes are already declared in the app level CSS file

```
  --env-var-font-size-small: 11px;
  --env-var-font-size-small-plus: 12px;
  --env-var-font-size-medium: 13px;
  --env-var-font-size-medium-plus: 14px;
  --env-var-font-size-large: 16px;
  --env-var-font-size-large-plus: 24px;
  --env-var-font-size-xlarge: 30px;
  ```

File: Client/src/Pages/Maintenance/CreateNewMaintenance.jsx [hunk 3]
```diff
@@ -0,0 +1,45 @@
+import React, { useState } from "react";
+import Dropdown from "../../Components/Dropdown";
+
+const CreateNewMaintenance = () => {
+  const repeatComponents = {
+    "Don't repeat": [<div key={1}>a</div>, <div key={2}>b</div>],
+    "Repeat daily": [<div key={3}>c</div>, <div key={4}>d</div>],
+    "Repeat weekly": [<div key={4}>e</div>, <div key={6}>f</div>],
```
ajhollid, line 8: This is a bit of an odd arrangement, can you explain the usage?

ajhollid (changes requested):

Client/src/Pages/Maintenance/index.css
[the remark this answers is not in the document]
terryjung0244: Corrected. 

font-size: var(--env-var-font-size-medium);  // 13px

Client/src/Pages/Maintenance/CreateNewMaintenance.jsx
terryjung0244: I encountered an issue using the Dropdown MUI Select. I created a 'repeatDropDown' array with three objects to use as select options. When I tried to send the 'repeatDropDown' object in an array, I needed to keep the Dropdown component unchanged, particularly its value. Initially, the Dropdown's propTypes value was set to a string. However, I changed it to an object. There's no error so far, but I'm unsure if this change is allowed.

const repeatDropDown = [
    { name: "Don't repeat" },
    { name: "Repeat daily" },
    { name: "Repeat weekly" },
];

Dropdown.propTypes = {
  value: PropTypes.object.isRequired,
};

Client/src/Pages/Maintenance/CreateNewMaintenance.jsx
terryjung0244: Sorry for the confusion.

I am trying to display the following components when the page initializes:

"Don't repeat": [
  <div key={1}>Date</div>,
  <div key={2}>Start time</div>,
  <div key={3}>Duration</div>,
  <div key={4}>Friendly name</div>,
  <div key={5}>Add monitors</div>
]

These components are based on the Figma maintenance structure. Each <div> would represent a reusable component that can be utilized later on.

'Repeat daily' and 'Repeat weekly' would be similar except for some features such as 'Repeating days'.

"Repeat daily": [
  <div key={2}>Start time</div>,
  <div key={3}>Duration</div>,
  <div key={4}>Friendly name</div>,
  <div key={5}>Add monitors</div>
]

"Repeat weekly": [
  <div key={2}>Repeating days</div>,
  <div key={3}>Start time</div>,
  <div key={4}>Duration</div>,
  <div key={5}>Friendly name</div>
]

Please let me know if I'm going in the wrong direction.

Thank you!

terryjung0244: Sorry for the confusion.

repeat

I am trying to display the following components when the page initializes. The components are based on the Figma maintenance structure. Each <div> would represent a reusable component that can be utilized later on. 

'Repeat daily' and 'Repeat weekly' would be similar except for some features such as 'Repeating days'.

Please let me know if I'm going in the wrong direction. 

Thank you.

ajhollid: Ok thanks for the explanation, so these are meant to be placeholders for components to be written later?

If that's the case we don't need to look at these too closely

Client/src/Pages/Maintenance/CreateNewMaintenance.jsx
ajhollid: What was the issue with the Select component?

We really don't want autocomplete here so it would be best if we could resolve the problem with Select.

It's a straightforward component so I'm sure we can resolve it.

ajhollid (changes requested):
I believe we should resolve the issue with Select and use that.

If you can share some specifics about the issue you encountered while using it we can help resolve it.

Client/src/Pages/Maintenance/CreateNewMaintenance.jsx
terryjung0244: Correct. We don't need to focus on these for now.

ajhollid:
```
import { Select } from "@mui/material";
import { MenuItem } from "@mui/material";
import { useState } from "react";

const CreateNewMaintenance = () => {
  const [repeat, setRepeat] = useState(0);

  return (
    <div>
      <div>Create a maintenance window</div>
      <div>Your pings won&apos;t be sent in this time frame.</div>
      <div>Repeat</div>
      <Select
        value={repeat}
        onChange={(e) => {
          setRepeat(e.target.value);
        }}
      >
        <MenuItem value={0}>Don't Repeat</MenuItem>
        <MenuItem value={1}>Repeat Daily</MenuItem>
        <MenuItem value={2}>Repeat Weekly</MenuItem>
      </Select>
    </div>
  );
};

export default CreateNewMaintenance;
```

Here's a very simple example of how you can use `Select` for your component

PR #364, opened 2024-07-17
Added validation to speedcheck route.

Skorpios604:


ajhollid (changes requested):
As discussed, validation can be moved to controller level

ajhollid (approved):

PR #375, opened 2024-07-18
Add startup script

ajhollid:
run `./uptime.sh` to one command start the application


Skorpios604 (approved):
Up and running!

ajhollid:
Env var bug, will update when fixed

PR #387, opened 2024-07-19
Nodemailer implemented

MuhammadKhalilzadeh:
Nodemailer is Implemented
Email templated are dynamic in variables, but the code inside the route for
```
app.use("/api/v1/mail", (req, res) => {
// my code
})
```
is hard coded, so it needs to be modified to be dynamic according to the situation and the response for the user.

Results:
Screenshot (96)


MuhammadKhalilzadeh:
Screenshot (95)

ajhollid (approved):

PR #388, opened 2024-07-19
Updated validation schemas in the joi and changed validation schema in the controller for the page speed check. 

Skorpios604:


ajhollid (approved):
Looking good!

PR #404, opened 2024-07-21
Feat/table order, resolves #403

### LINKED ISSUE #403: Monitor Checks Sort Order
image

I believe we would like the Graph to display as above left `(oldest) -> right (newest)`.

However the chart displays as `top (oldest) -> bottom (newest)` , but I believe we want this to display as `top (newest) -> bottom (oldest)` as in a last-in-first-out stack.

Can you confirm @gorkem-bwl ?

If that's the case we can manipulate the data to display this properly.

ajhollid:
Data in `BasicTable` can now be displayed in reverse order by adding `reverse` prop


Skorpios604 (approved):
Nice work! Simple and works!

PR #413, opened 2024-07-22
Refactored form structure and added fix to optional field in create monitor

danielcj2:


Client/src/Pages/Monitors/CreateMonitor/index.jsx
```diff
@@ -84,11 +84,11 @@ const CreateMonitor = () => {
     event.preventDefault();
     //obj to submit
     let monitor = {
-      ...generalSettings,
       url:
         checks.type === "http"
           ? "https://" + generalSettings.url
           : generalSettings.url,
+      name: generalSettings.name || generalSettings.url,
```
ajhollid, line 90: Can we replace this with a ternary?  I am really not a fan of falsey evaluation, it makes my skin crawl :P

`generalSettings.name === "" ? generalSettings.url : generalSettings.name`

is a bit more verbose but it is very clear as to what the intent is and there can't be any unexpected evaluations

ajhollid (changes requested):
Can we replace the `||` evaluation with a tenary?  I think it's a bit clearer and a bit safer

ajhollid (approved):

PR #417, opened 2024-07-22
Status fallback

danielcj2:
I noticed that this frame is used quite a bit in the Figma design so I refactored it into a reusable component.


ajhollid (approved):
Reusability ftw

PR #427, opened 2024-07-23
Team panel table

danielcj2:
There is no functionality added yet other than filtering the table. Before I add the role change dropdown, there needs to be a proper select component implemented, so I'll get that sorted out asap.


ajhollid (approved):
Looking good

Skorpios604 (approved):

PR #428, opened 2024-07-23
applying changes

MuhammadKhalilzadeh:


ajhollid (commented):
This all looks fine to me :+1:

Client/src/Components/NavBar/index.jsx
```diff
@@ -99,8 +99,9 @@ function NavBar() {
       position="sticky"
       sx={{
         width: "100%",
-        backgroundColor: "white",
-        boxShadow: "var(--env-var-shadow-1)",
+        backgroundColor: "transparent",
+        boxShadow: "none",
+        borderBottom: "1px solid var(--color-border-0)",
```
danielcj2, line 104: Looks like you're creating overflow by adding a 1px border. This is due to how the height of the dashboard sidebar is being calculated.

`.dashboard-sidebar {
height: calc(100vh - 64px)`

By adding a 1px border you are effectively increasing the height of the navBar to `65px`. 

Also it would be nice to refactor the height of the navBar into a variable.

danielcj2 (commented):
Looks good, just a small issue I found.

Client/src/Components/NavBar/index.jsx
MuhammadKhalilzadeh: It's a single and mostly unique value in our code and making variables out of values like this would generate lines and lines of code and hard to maintain.
But, I do hate overflows too. 🤧 
Let's get ride of it. 
Fixed ✅

ajhollid: If a value is used in two places, IE to set the height of the `NavBar`, and then to calculate the height of something else that depends on the height of the `NavBar` then there is no question, it should be a variable.

Reasoning is if the height of the `NavBar` ever changes now anything that depends on the height of the `NavBar` is no longer correct.

If they are both based on a variable and the variable changes then there is no issue, heights will all be calcualted correctly.

MuhammadKhalilzadeh: Let me explain values like this with an example.
Let's say we have the value of `64px` in several places in our files.
Do we need to make it a `global single variable` and reference all those places to this global variable?
It depends on the situation. Well, coincidentally these values are equal, but their goals are different.
And also in many cases, the goal of all these values might be the same in a certain period of software's life cycle, but each of these values may change in the future due to upgrades in the software to updates in the User Interface.
So I guess its wiser to keep a unique value of `(X)px` hard coded for the moment and it gets repeated several times, then make it dynamic as a `variable`

In our case, `64px`, as you said was the default value for the MUI and can not be changed.
Me and @gorkemcetin reviewed several options for making changes in the `NavBar` and the overall look of the dashboard and adding this border-bottom fitted the most, which appeared to cause a slight overflow on the bottom of the page.

So Now I've made changes to the height of the `DashboardSidebar` which had the height of `height: calc(100vh - var(--env-var-spacing-5));` and in this case `--env-var-spacing-5` equals `65px` to remove the `1px` overflow we had.

The 64px in our `DashboardSidebar` was an independent and unique value with only one purpose.

ajhollid: I believe the issue with the value of `64px` is that it is _not_ an isolated and arbitrary value that is only used in one location.

It is a value that is derived from the `minHeight` of the `NavBar` as set by MUI, and it is used to calculate the height of the `SideBar`.

My thought process is these two values are intextricably linked and as such should share the same value, ergo the value should be extracted to a variable which can be shared between the two components.

It may be fine to have the same value hardcoded in two components for now while we all remember that they both reference the same value, but we may not be the ones that are working on the project in 6 months, or we may all forget as we move forward and work on different parts of the app.

Now as for not being able to change the height `NavBar` height of `64px`, you should be able to simply target the class directly `.MuiToolbar-root` and override the default styles.

I believe that it is a good idea to override the height of the `NavBar` and specify our own value in a CSS variable, even if it is the same `64px`, so that it is under our control and we can guarantee that our calculations _based on that variable_ will be correct.

```
int x = 2;
int y = 2;
int z = y * x;
z / y === x;
z / x === y;
```

is inherently more maintainable than

```
int z = 2 * 2
z / 2 === 2
z / 2 === 2
```

In the first example if we change we change `x` or `y` in we don't need to change anything else, the code is still correct.

In the second example if we change the either of the operands we need to change it in `n` places, every place that it is used.  We might not even remember where we used it or why.  That is much more work and much more congtiive load as we must remember exactly where we used values and why.

Does that all make sense?  I hope that helps explain the value I see in refactoring that specific value into a variable, and the whole idea in general.

Skorpios604: I agree with @ajhollid sentiment and @danielcj2 it is important to use a variable in a case like this.

Client/src/index.css
```diff
@@ -64,6 +65,7 @@
   --env-var-spacing-2: 24px;
   --env-var-spacing-3: 32px;
   --env-var-spacing-4: 40px;
+  --env-var-spacing-5: 65px;
```
ajhollid, line 68: I think it would be best to name this something like `nav-bar-height`, then we can use it to set both the height of the `NavBar` by targeting `.MuiToolbar-root` and the height of the `SideBar`:

```
(100vh - var(--nav-bar-height));
```

I think that's a net improvement :+1:  It's easier to maintain and the it is very clear from the calculation that the height of the `SideBar` is dependant upon the height of the `NavBar`

ajhollid (changes requested):
Minor change with respect to the CSS var and how it can be used to calcualte the height of the `NavBar`

Client/src/index.css
MuhammadKhalilzadeh: No need to target `.MuiToolbar-root`
Overflow is fixed already!

MuhammadKhalilzadeh:
Now we have `--env-var-nav-bar-height` with the value of `64px` for both `NavBar` and `DashboardSidebar`

PR #432, opened 2024-07-23
Feat/invite_registration

ajhollid:
Implemented registration from email with token.

Please double check my updates to the validation, I think I've got it straight but I'd apprecaite more eyes on it

### DIFF
--- a/Server/controllers/authController.js
+++ b/Server/controllers/authController.js
@@ -59,20 +59,21 @@ const registerController = async (req, res, next) => {
     return;
   }
 
-  // Check if an admin user exists, if so, error
-  try {
-    const admin = await req.db.checkAdmin(req, res);
-    console.log(admin);
-    if (admin === true) {
-      throw new Error(errorMessages.AUTH_ADMIN_EXISTS);
-    }
-  } catch (error) {
-    console.log("WEEEEEEE", error.message);
-    error.service = SERVICE_NAME;
-    error.status = 403;
-    next(error);
-    return;
-  }
+  // TODO  Can there be more than one admin?
+  // // Check if an admin user exists, if so, error
+  // try {
+  //   const admin = await req.db.checkAdmin(req, res);
+  //   console.log(admin);
+  //   if (admin === true) {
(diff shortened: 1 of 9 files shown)


Server/db/MongoDB.js
```diff
@@ -185,6 +185,21 @@ const requestInviteToken = async (req, res) => {
   }
 };
 
+const getInviteToken = async (req, res) => {
+  try {
+    console.log(req.body.token);
+    const invite = await InviteToken.findOneAndDelete({
+      token: req.body.token,
+    });
+    if (invite === null) {
```
Skorpios604, line 193: maybe 
```javascript 
if (!invite)
```

Server/controllers/authController.js
```diff
@@ -237,6 +238,21 @@ const inviteController = async (req, res, next) => {
...
+  try {
+    await inviteVerifciationBodyValidation.validateAsync(req.body);
+    const invite = await req.db.getInviteToken(req, res);
+
+    res
+      .status(200)
+      .json({ status: "success", msg: "Invite verified", data: invite });
+  } catch (error) {
+    error.service = SERVICE_NAME;
+    next(error);
```
Skorpios604, line 252: Function will end at ```next(error);```
No need for a return here I don't think.

Skorpios604 (approved):

Server/db/MongoDB.js
ajhollid: I always prefer strict evaluation whenver possible.  

`!invite` allows for all sorts of things like `0`, `""`, 'undefined`, `NaN` etc that I would definitely _not_ want passing that check :P

Skorpios604: Good point!

Server/controllers/authController.js
ajhollid: I belive you are correct here, the return is left over from an earlier function I wrote that had a conditional `next()` call and I wanted to make sure no further code was executed after that so I added the `return` and now I've just gotten in the habit of doing it to be safe.

PR #437, opened 2024-07-24
Select component

danielcj2:
Built select component and added styling. The `items` prop needs to take in an array of objects, with `_id` (value) and `name` (text to display) fields required.

### DIFF
--- a/Client/src/Components/Inputs/Select/index.jsx
+++ b/Client/src/Components/Inputs/Select/index.jsx
@@ -1,6 +1,11 @@
 import PropTypes from "prop-types";
 import { useTheme } from "@emotion/react";
-import { MenuItem, Select as MuiSelect } from "@mui/material";
+import {
+  MenuItem,
+  Select as MuiSelect,
+  Stack,
+  Typography,
+} from "@mui/material";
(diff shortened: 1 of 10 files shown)


Client/src/Components/Inputs/Select/index.jsx
```diff
@@ -0,0 +1,116 @@
...
+ * @param {boolean} props.isHidden - Whether the placeholder should be hidden.
+ * @param {string} props.value - The currently selected value.
+ * @param {object[]} props.items - The array of items to populate in the select dropdown.
+ *    @param {string} props.items._id - The unique identifier of each item.
+ *    @param {string} props.items.name - The display name of each item.
+ * @param {function} props.onChange - The function to handle onChange event.
+ * @param {object} props.sx - The custom styles object for MUI Select component.
+ * @returns {JSX.Element}
+ */
+
```
ajhollid, line 28: Could you add some instructions on how to use this component as well as an example to the JSdocs?

I think pretty much every dev is going to use this componet at some point so some instructions here would be super helpful :+1:

ajhollid (changes requested):
Nice, reusable components make my day.

Can we add instructions and examples to the JSdocs?  Most important being the format of items that are passed to the component

ajhollid (approved):

PR #438, opened 2024-07-24
Hide team panel for users

danielcj2:


Client/src/Components/NavBar/index.jsx
```diff
@@ -51,6 +50,10 @@ function NavBar() {
   const navigate = useNavigate();
   const authState = useSelector((state) => state.auth);
 
+  let settings = ["Profile", "Password", "Team", "Logout"];
+  if (authState.user.role[0] !== "admin")
```
ajhollid, line 54: It's not guaranteed that "admin" will be the first role in the `role` array, this should probably search the array to check if "admin" exists anywhere in the array

ajhollid (changes requested):
Looks good!

The order of the `role` array is not guaranteed though, so searching the array for existance of "admin" is probably the best approach here

ajhollid (approved):

PR #457, opened 2024-07-26
Fixed error that occurs when authState.user is undefined

Skorpios604:
authState.user.role can now be safely accessed if authState.user is undefined preventing the includes error.

### DIFF
--- a/Client/src/Components/NavBar/index.jsx
+++ b/Client/src/Components/NavBar/index.jsx
@@ -8,7 +8,6 @@ import Typography from "@mui/material/Typography";
 import Menu from "@mui/material/Menu";
 import MenuIcon from "@mui/icons-material/Menu";
 import Container from "@mui/material/Container";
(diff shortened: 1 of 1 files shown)


ajhollid (commented):
Curious bug, @danielcj2 have you come across this while working?

I've started with a blank slate several times and I've never come across it.

Could be I just haven't done a clean start since recently enough to run into this bug 🤔

danielcj2:
> Curious bug, @danielcj2 have you come across this while working?
> 
> I've started with a blank slate several times and I've never come across it.
> 
> Could be I just haven't done a clean start since recently enough to run into this bug 🤔

🤔 That's interesting. There shouldn't be a case where user is undefined I don't think. I wonder if the error is thrown because the user data might be outdated? We did update the naming convention. Could be worth to check, since personally I haven't run into this bug.

danielcj2 (approved):
Looks good to me!

PR #458, opened 2024-07-26
Maintenance window options screen

MuhammadKhalilzadeh:
Here's the implementation of maintenance window options screen. 
I tried different ways, but I could not find any ways to change the Date Picker and Time Picker appearance to match the our designs.

Screenshot 2024-07-26 145028

### DIFF
--- a/Client/package-lock.json
+++ b/Client/package-lock.json
@@ -8,19 +8,20 @@
       "name": "client",
       "version": "0.0.0",
       "dependencies": {
-        "@emotion/react": "^11.11.4",
-        "@emotion/styled": "^11.11.5",
+        "@emotion/react": "^11.13.0",
+        "@emotion/styled": "^11.13.0",
         "@fontsource/roboto": "^5.0.13",
         "@mui/icons-material": "^5.15.17",
         "@mui/lab": "^5.0.0-alpha.170",
-        "@mui/material": "^5.15.16",
+        "@mui/material": "^5.16.5",
+        "@mui/styled-engine-sc": "^6.0.0-alpha.18",
         "@mui/x-charts": "^7.5.1",
         "@mui/x-data-grid": "7.3.2",
-        "@mui/x-date-pickers": "7.3.2",
+        "@mui/x-date-pickers": "^7.3.2",
         "@reduxjs/toolkit": "2.2.5",
         "axios": "1.7.2",
         "chart.js": "^4.4.3",
-        "dayjs": "1.11.11",
+        "dayjs": "^1.11.11",
         "joi": "17.13.1",
         "jwt-decode": "^4.0.0",
         "react": "^18.2.0",
@@ -31,6 +32,7 @@
         "react-toastify": "^10.0.5",
         "recharts": "2.12.7",
         "redux-persist": "6.0.0",
+        "styled-components": "^6.1.12",
         "vite-plugin-svgr": "^4.2.0"
       },
       "devDependencies": {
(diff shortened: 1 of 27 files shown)


MuhammadKhalilzadeh:
And also for these dropdowns, I targeted many classes, but disabling neither of them couldn't invisible this blue background-color!

Screenshot 2024-07-26 145544

gorkem-bwl:
HH:MM picker: how does the user experience look like there? Is it possible for the user select hour, minute and AM/PM?

MuhammadKhalilzadeh:
Yes, this is how it looks
Screenshot_20240726_161627_Chrome

MuhammadKhalilzadeh:
@gorkem-bwl how do you like it?

gorkem-bwl:
> @gorkem-bwl how do you like it?

Looks great to me!

MuhammadKhalilzadeh:
@gorkem-bwl  Then let's go with this. 👍🏻
How about other parts?

gorkem-bwl:
> @gorkem-bwl Then let's go with this. 👍🏻 How about other parts?

Can you dynamically search for other monitors at the “add monitors” section?

MuhammadKhalilzadeh:
@gorkem-bwl I believe it's for the same exact purpose, but first we need to connect it to backend and bind it to the UI.
This issue (task) is just for creating the UI component of Maintenance window options screen.

ajhollid:
> And also for these dropdowns, I targeted many classes, but disabling neither of them couldn't invisible this blue background-color!
> 
> Screenshot 2024-07-26 145544

@MuhammadKhalilzadeh we have a select component in the project now you should be able to just drop in here

MuhammadKhalilzadeh:
I don't think so!
My branch is up-to-date and I'm not finding any select components other than this one I'm using.
Which folder is it? Is it added recently?

MuhammadKhalilzadeh:
@ajhollid What's the name of the component we've created with it?

ajhollid:
> @ajhollid What's the name of the component we've created with it?

It's under Inputs.

It appears some of the changes you've made here are targeting global classes which have broken some components though.

We'll have to dig through it and find out what's going on.

Try and keep those CSS selectors as specific as possible, if you target too broadly you'll affect other components

MuhammadKhalilzadeh:
@ajhollid  Okay, then I guess we'll need to set a meeting and review and discuss my code in this part.

ajhollid:
With additions and deletions this PR is close to 800 lines, our target for pull requests should be at a maximum 200 lines.

It's going to take a while to go through the whole thing, we may have to go component by component since there are a lot in this PR.

In future it would be best to have one PR per component, we can get code reviewed faster and feedback can be incorporated.

MuhammadKhalilzadeh:
No! it's not that much. I know the exact components I've worked on.
We can review them easily

ajhollid:
> No! it's not that much. I know the exact components I've worked on.
> We can review them easily



789 lines to go review here, you can see how many lines you have changed by running `git diff` before you commit. 

We have to be very careful in this review because we don't know what components have been affected by the wide CSS targets.

At the very least the Select component appears to have been broken by globally selecting the Input class.

ajhollid:
> Let's me think what we can do and get back to you asap

Sent DM on discord, let's pick up the discussion there 🙏

ajhollid:
> @gorkem-bwl I believe it's for the same exact purpose, but first we need to connect it to backend and bind it to the UI.
> This issue (task) is just for creating the UI component of Maintenance window options screen.

This shouldn't actually require a new backend request, all the monitors are in the redux store since they are fetched when the user first logs in.

You can just grab them from the redux store and build the autocomplete data from them 😎

PR #463, opened 2024-07-26
Feat/monitor endpoint refactor

ajhollid:
Refactored endpoints to remove near duplicate methods

Replaced duplicated functions with query params to modify queries

Added query param validation


danielcj2 (approved):
Nice! Looking much cleaner.

PR #469, opened 2024-07-27
Checkbox component

danielcj2:


ajhollid (approved):
Looks like a good component, what's the negative margin for though? Always curious whenever I see a negative margin 😂

danielcj2:
> Looks like a good component, what's the negative margin for though? Always curious whenever I see a negative margin 😂

It's used to counteract the padding. The padding pushes the element, and the negative margin pulls the element back into place. Basically I'm adding an onHover background change effect with some padding, while still keeping the element aligned with the rest of the content.

ajhollid:
> > Looks like a good component, what's the negative margin for though? Always curious whenever I see a negative margin 😂
> 
> It's used to counteract the padding. The padding pushes the element, and the negative margin pulls the element back into place. Basically I'm adding an onHover background change effect with some padding, while still keeping the element aligned with the rest of the content.

Ahh got it, thanks for the explanation 👌

PR #470, opened 2024-07-27
Page Speed

danielcj2:
Will get the fields hooked up to local state in my next PR. 

Issues: 
 - proptype error. Need to refactor `StatusLabel` component to allow for different status texts


Client/src/Utils/timeUtils.js
```diff
@@ -0,0 +1,59 @@
+export const formatDuration = (ms) => {
```
ajhollid, line 1: Nice, thanks for refactoring this out

Client/src/Pages/PageSpeed/index.jsx
```diff
@@ -0,0 +1,153 @@
...
+  return (
+    <Grid item lg={6} flexGrow={1}>
+      <Stack direction="row" gap={theme.gap.medium} p={theme.gap.ml}>
+        <PageSpeedIcon style={{ width: theme.gap.ml, height: theme.gap.ml }} />
+        <Box flex={1}>
+          <Stack direction="row" justifyContent="space-between">
+            <Typography component="h2" mb={theme.gap.xs}>
+              {data.name}
+            </Typography>
+            {/* TODO - fix prop-type error */}
```
ajhollid, line 40: We could just remove the acceptable `strings` from the label component, I don't think there's any real good reason to limit what strings we can use there :thinking:

ajhollid (approved):
Nice work, appreciate `timeUtils` refactor :100:

Client/src/Pages/PageSpeed/index.jsx
danielcj2: Yea, that's probably the move. I'm thinking we should define some styles types in the component itself as well, that way there's no need to pass a color variable for background/dot at all.

Skorpios604 (approved):
A masterpiece

PR #474, opened 2024-07-29
Fix/remove tq

ajhollid:
I created this file and committed it accidentaly, not sure where it came from :thinking: It contains a git diff so I must have accidentally outputted to file instead of terminal at some point.


Skorpios604 (approved):
good catch

PR #477, opened 2024-07-29
Page speed state

danielcj2:


ajhollid (approved):
Looking good on my end.

This is not connected up yet correct?  TODO multiply the minute value of the Select component by milliseconds per minute (60k)

danielcj2:
> Looking good on my end.
> 
> This is not connected up yet correct? TODO multiply the minute value of the Select component by milliseconds per minute (60k)

Correct, it's only hooked up to local state. No create function yet.

PR #501, opened 2024-07-31
Feat/fe/connect config, resolves #497

### LINKED ISSUE #497: FE - Monitor Config
Monitor URL should not be changeable in the config page.  Previous checks would not make sense if the URL changed

image

the `http://` in the above image can be removed and the field could just show the full URL

Type should not be changeable either as previous checks would not make sense again

ajhollid:
Removed/disabled editing of features that should not be editable

Connected Save and Delete buttons

Fetch moniotrs on successful update to refresh UI


danielcj2 (approved):
Looks good and works as expect. I think we should consider adding an extra step to deleting the monitor, in case the user clicks on the button accidentally. A confirm delete modal like in `ProfilePanel` could work.

PR #508, opened 2024-07-31
Add monitor notifications to state 

danielcj2:


Client/src/Pages/Monitors/CreateMonitor/index.jsx
```diff
@@ -219,33 +212,27 @@ const CreateMonitor = () => {
...
-                onChange={(event) => handleChange(event, "type", setChecks)}
+                checked={monitor.type === "http"}
+                onChange={(event) => handleChange(event)}
               />
-              {checks.type === "http" || checks.type === "https" ? (
+              {monitor.type === "http" ? (
                 <ButtonGroup sx={{ ml: "32px" }}>
                   <Button
                     level="secondary"
                     label="HTTP"
```
ajhollid, line 222: Can we change the order of these?  HTTPS should be the most common so it makes more sense to have that as the leftmost and default value I think

ajhollid (approved):
Looks good and working well!

ajhollid (approved):
Awesome, thanks

Skorpios604 (approved):

PR #530, opened 2024-08-02
Fix/monitor table row click

ajhollid:
Stop propagation of events so row click works as intended

Added inline style for ResponseTimeChart to set cursor to pointer


danielcj2 (approved):
Looking good, thank you for fixing that 😄

PR #532, opened 2024-08-02
Updated create page speed monitor page

danielcj2:
image

There was too much whitespace so I limited the form width to `1000px`. The alternative would be to match the `CreateMonitor` page design, since the two use the same components.


ajhollid (approved):

gorkem-bwl:
Screenshot looks neat and tidy.

PR #538, opened 2024-08-03
Pagespeed details

danielcj2:
This is an early draft, decided to push the branch because it is getting a bit out of hand. The page only displays the checks and a pie chart (no highlight effects as of now). None of the metrics are displayed yet.

The implementation for the centered label is a little hacky but I'll get it sorted in the future.


ajhollid (approved):
Looks good, hacks will get worked out no doubt!

PR #540, opened 2024-08-05
Refactored to use MUI charts

ajhollid:
Refactors the ResponseTime chart to use MUI charts instead of Recharts

I'm on the fence as to whether this is a good refactor or not

Pros 
* It uses MUI library, so no need for another separate library

Cons
* It is not as straightforward to wrangle the data, it requires mapping the data to multiple different arrays in order to correctly display data with different colors for up and down
* Doesn't seem to display quite as nicely as Recharts

I like not having to have a second library involved, but I really don't like having to do all the mapping with the data. 

Appreciate everyones thoughts on this one!

### DIFF
--- a/Client/src/Components/Charts/ResponseTimeChart/index.css
+++ b/Client/src/Components/Charts/ResponseTimeChart/index.css
@@ -1,7 +1,4 @@
 .chart-container {
-  display: flex;
-  justify-content: space-around;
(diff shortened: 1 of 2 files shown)


ajhollid:
After discussion this refactors costs outweigh the benefits in my opinion

PR #543, opened 2024-08-05
Pagespeed pie chart 

danielcj2:
There's some edge cases to be tested but other than that the pie chart is about fully functional.

### DIFF
--- a/Client/src/Pages/PageSpeed/Details/index.css
+++ b/Client/src/Pages/PageSpeed/Details/index.css
@@ -58,7 +58,8 @@
 .page-speed-details .MuiPieArcLabel-faded {
   fill: rgba(0, 0, 0, 0.3);
 }
-.page-speed-details .pie-label {
+.page-speed-details .pie-label,
+.page-speed-details .pie-value-label {
   transition: all 400ms ease;
 }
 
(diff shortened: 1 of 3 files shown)


Client/src/Pages/PageSpeed/Details/index.css
```diff
@@ -51,3 +51,18 @@
...
+.page-speed-details .MuiPieArcLabel-root {
+  font-size: var(--env-var-font-size-small-plus);
+  transition: fill 300ms ease;
+}
+.page-speed-details .MuiPieArcLabel-faded {
+  fill: rgba(0, 0, 0, 0.3);
+}
+.page-speed-details .pie-label,
+.page-speed-details .pie-value-label {
+  transition: all 400ms ease;
```
ajhollid, line 63: Can we slow down the animations a little bit?  Seems a bit jumpy to me when hovering over, might be nice to have the animation take a little longer :thinking:

ajhollid (approved):
Nice work recreating this entirely out of MUI charts, looking good!  My only suggestion is perhaps to slow down the animations a little bit

Client/src/Pages/PageSpeed/Details/index.css
danielcj2: You're referring to the pie arc animations right? I haven't touched those so they're default as they come from @mui. I can possibly override them though. I'll look into it.

PR #549, opened 2024-08-06
Bugfix & Pagespeed

ajhollid:
1.  Adds bugfix for the monitors detail chart
2. Pagespeed details page implemented


danielcj2 (approved):
Send it 🚀

PR #562, opened 2024-08-07
Changed the name here to pageSpeedCheckController as to not be confus…

Skorpios604:
as to not be confused with the actual check controller.


ajhollid (approved):
Good catch, that was indeed confusing!

PR #564, opened 2024-08-07
Loading skeletons

### LINKED ISSUE #550: "No incidents, yet" placeholder shouldn't be visible if there are incidents
The "No incidents, yet" placeholder shows up momentarily for almost half a second before the screen is painted with the incidents table. 

If there are no incidents, this placeholder should be shown.

danielcj2:
The PR may seem a tiny bit more intimidating than it actually is 😂 . A lot of the code was re-layered, which resulted in the high # of lines changed.

Resolves #550

### DIFF
--- a/Client/src/Pages/Incidents/index.jsx
+++ b/Client/src/Pages/Incidents/index.jsx
@@ -12,6 +12,32 @@ import Select from "../../Components/Inputs/Select";
 
 import "./index.css";
 
+/**
+ * Renders a skeleton layout.
+ *
+ * @returns {JSX.Element}
+ */
+const SkeletonLayout = () => {
+  const theme = useTheme();
+
+  return (
+    <>
+      <Stack direction="row" alignItems="center" gap={theme.gap.medium}>
+        <Skeleton variant="rounded" width={150} height={34} />
+        <Skeleton variant="rounded" width="15%" height={34} />
+        <Skeleton
+          variant="rounded"
+          width="20%"
+          height={34}
+          sx={{ ml: "auto" }}
+        />
+      </Stack>
+      <Skeleton variant="rounded" width="100%" height={300} />
+      <Skeleton variant="rounded" width="100%" height={100} />
+    </>
+  );
+};
+
 const Incidents = () => {
   const theme = useTheme();
   const navigate = useNavigate();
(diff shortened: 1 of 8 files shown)


Client/src/Pages/Incidents/index.jsx
```diff
@@ -102,7 +105,22 @@ const Incidents = () => {
...
+              <Skeleton variant="rounded" width="15%" height={34} />
+              <Skeleton
+                variant="rounded"
+                width="20%"
+                height={34}
+                sx={{ ml: "auto" }}
+              />
+            </Stack>
+            <Skeleton variant="rounded" width="100%" height={300} />
+            <Skeleton variant="rounded" width="100%" height={100} />
```
ajhollid, line 121: Could the whole Skeleton be extracted to a component within the class? Just to keep the main render method a bit more readable

Client/src/Pages/Monitors/Configure/index.jsx
```diff
@@ -164,53 +164,88 @@ const Configure = () => {
...
+            {loading ? (
+              <>
+                <Skeleton
+                  variant="rounded"
+                  width="50%"
+                  height={24}
+                  sx={{ mb: theme.gap.small }}
+                />
+                <Skeleton variant="rounded" width="50%" height={18} />
+              </>
```
ajhollid, line 219: Same for this one

ajhollid (changes requested):
Scared me for a minute!

It looks fine to me, but I think it would be nice to extract those skeletons to their own components to keep the render method clean and readable.

If not possible / feasible let me know

danielcj2:
> Scared me for a minute!
> 
> It looks fine to me, but I think it would be nice to extract those skeletons to their own components to keep the render method clean and readable.
> 
> If not possible / feasible let me know

It is definitely possible. We can have a folder with skeletons for each page. The reason why I didn't was because I had to use some arbitrary values to minimize the content shifting its position up and down, and I opted to leave the skeleton code beside the element that it is covering up in case we would need to make changes to the structure of the page. Also, it felt easier not worrying about positioning the skeletons since the base structure is done already 😛 

I'll work on refactoring the skeletons to their own components. Unfortunately, I don't think they'll be reusable since most our pages have different layouts.

ajhollid:
> > Scared me for a minute!
> > 
> > It looks fine to me, but I think it would be nice to extract those skeletons to their own components to keep the render method clean and readable.
> > 
> > If not possible / feasible let me know
> 
> It is definitely possible. We can have a folder with skeletons for each page. The reason why I didn't was because I had to use some arbitrary values to minimize the content shifting its position up and down, and I opted to leave the skeleton code beside the element that it is covering up in case we would need to make changes to the structure of the page. Also, it felt easier not worrying about positioning the skeletons since the base structure is done already 😛 
> 
> I'll work on refactoring the skeletons to their own components. Unfortunately, I don't think they'll be reusable since most our pages have different layouts.

No problem, I wasn't expecting they be reusable, just refactored for visibility.

Thanks for doing that!

ajhollid (approved):

PR #573, opened 2024-08-08
Configure pagespeed page

danielcj2:
The code is much of the same as the `create pagespeed page`. Updated redux to include thunks for updating and deleting pagespeed monitors.


ajhollid (approved):
Looks good to me!

Skorpios604 (approved):

Issue #566, opened 2024-08-08
Emails are case sensitive for logging into the app


Skorpios604:
Must be changed to not be case sensitive.

PR #578, opened 2024-08-09
Fixed login page email to not be case sensitive anymore. Closes issue #566

Skorpios604:


ajhollid (commented):
Not sure if a users email is converted to lowercase when they register, I don't recall doing that so we should double check that

PR #585, opened 2024-08-09
Feat/login security, resolves #583

### LINKED ISSUE #583: FE & BE - Double login prevention
Users should not be able to login more than once concurrently

Since our API is stateless we should probably handle this by using the JWT.

I think we could do something like this:
 
1. Store the JWT in the user object when logging in
2. When logging in, check if the user has a JWT
3. When checking JWT in middleware, remove JWT from user it if it has expired
4. Remove JWT from user on logout

ajhollid:
This PR should prevent users from being able to login in mulitple clients

- [x] Store the JWT in the user object when logging in
- [x] When logging in, check if the user has a JWT
- [x] When checking JWT in middleware, remove JWT from user it if it has expired
- [x] Remove JWT from user on logout

### DIFF
--- a/Client/src/Components/NavBar/index.jsx
+++ b/Client/src/Components/NavBar/index.jsx
@@ -99,9 +99,10 @@ function NavBar() {
     // Clear auth state
     dispatch(clearAuthState());
     dispatch(clearUptimeMonitorState());
+    // Make request to BE to remove JWT from user
     await axiosIntance.post(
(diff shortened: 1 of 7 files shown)


Skorpios604 (approved):
Everything seems to be in order.

ajhollid:
Removing this feature as it is not particularly useful and is causing headaches for some users.

Issue #586, opened 2024-08-09
FE - Table population delay


danielcj2:
It seems there is a slight delay when the table populates. The skeleton component displays as expected while data is being fetched.  However, after that operation concludes, there is an additional brief delay before the rows are displayed.

PR #589, opened 2024-08-10
Stop double rendering, add monitors to useEffect dependency array, #586

ajhollid:
There was a double rendering happening in `IncidentTable`, I've added a check condition to stop the extra rendering and fixed the depndency array as well.

I believe this resovles the skeleton issue, can you comfirm @danielcj2 ?


danielcj2 (approved):
Not sure if this removes all the delay 🤔, but there's definitely less now.

ajhollid:
> Not sure if this removes all the delay 🤔, but there's definitely less now.

Lets keep an eye on it then and see if it requires further attention.  We can leave the issue open until we are satisfied with it.

PR #600, opened 2024-08-12
Added fallback UI when there are no incidents recorded

danielcj2:
- [x] Added fallback UI when there are no incidents recorded.
- [x] Added fallback UI when a specific monitor has no incidents recorded.
- [x] Removed redundant divs.


ajhollid (approved):
:+1:

PR #602, opened 2024-08-12
Forgot password page

### LINKED ISSUE #593: FE - Password reset email should autocomplete
Since email needs to be entered before you can start the password reset process the email it would be nice to have the email filled in for the password request field so the user doesn't need to enter it again.

Currently email needs to be entered twice, once here:

image

And again here immediately after:

image

danielcj2:
- [x] Revamped layout.
- [x] Added BlueWave Uptime icon.
- [x] Implemented autocomplete email on navigate. 

Looks much the same as the other auth pages.

Resolves #593


ajhollid (approved):
Always nice to see divs go away.  Looks good :+1:

PR #608, opened 2024-08-12
Refactor pagespeed details to use stats endpoint

ajhollid:
- [x] Refactored stats endpoint to return appropriate stats for different moniotr types
- [x] Refactored Pagespeed page to take advantage of stats endpoint

### DIFF
--- a/Client/src/Pages/Monitors/Details/index.jsx
+++ b/Client/src/Pages/Monitors/Details/index.jsx
@@ -5,7 +5,6 @@ import { useSelector } from "react-redux";
 import { useNavigate, useParams } from "react-router-dom";
 import axiosInstance from "../../../Utils/axiosConfig";
 import MonitorDetailsAreaChart from "../../../Components/Charts/MonitorDetailsAreaChart";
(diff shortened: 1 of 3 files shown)


danielcj2 (approved):

Client/src/Pages/Monitors/Details/index.jsx
```diff
@@ -267,7 +266,7 @@ const DetailsPage = () => {
                 </ButtonGroup>
               </Stack>
               <Box sx={{ height: "200px" }}>
-                <MonitorDetailsAreaChart checks={monitor.checks} />
+                <MonitorDetailsAreaChart checks={monitor.checks.reverse()} />
```
Skorpios604, line 269: Are we wanting to modify the original array here? 

I saw that this const was used before 

```javascript
const reversedChecks = monitor.checks.slice().reverse();
```

ajhollid: Good catch, I don't think we want to mutate the array

PR #612, opened 2024-08-12
Responsive design

danielcj2:
- [x] Added responsive design to authentication pages.
- [x] Changed pattern image to `div` background image for more flexibility.

Didn't feel like it was necessary to account for multiple breakpoints since the authentication pages are relatively simple and don't display much content. Let me know what you all think.


alvarotorrestx (approved):

ajhollid (approved):
Auth screens look good to me, agreed I don't feel like we need many breakpoints here.

PR #617, opened 2024-08-13
Login and register emails changed to lowercase now so accounts can be recognized if user enters uppercase. 

Skorpios604:
- [x] updated login and register slices. 
- [x] updated validation.

### DIFF
--- a/Client/src/Features/Auth/authSlice.js
+++ b/Client/src/Features/Auth/authSlice.js
@@ -14,7 +14,9 @@ export const register = createAsyncThunk(
   "auth/register",
   async (form, thunkApi) => {
     try {
-      const res = await axiosInstance.post("/auth/register", form);
+      // Convert email to lowercase before sending the request
+      const modifiedForm = { ...form, email: form.email.toLowerCase() };
+      const res = await axiosInstance.post("/auth/register", modifiedForm);
       return res.data;
     } catch (error) {
       if (error.response.data) {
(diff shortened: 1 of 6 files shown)


Client/src/Features/Auth/authSlice.js
```diff
@@ -14,7 +14,9 @@ export const register = createAsyncThunk(
   "auth/register",
   async (form, thunkApi) => {
     try {
-      const res = await axiosInstance.post("/auth/register", form);
+      // Convert email to lowercase before sending the request
+      const modifiedForm = { ...form, email: form.email.toLowerCase() };
```
ajhollid, line 18: Can we move the `toLowerCase()` function to the register page?  Ideally we should pefrorm all operations we need to before handing things off to redux, redux should only be kept clean for state management only

Client/src/Features/Auth/authSlice.js
```diff
@@ -27,7 +29,9 @@ export const register = createAsyncThunk(
 
 export const login = createAsyncThunk("auth/login", async (form, thunkApi) => {
   try {
-    const res = await axiosInstance.post(`/auth/login`, form);
+    // Convert email to lowercase before sending the request
+    const modifiedForm = { ...form, email: form.email.toLowerCase() };
```
ajhollid, line 33: Same with login, this would ideally be done on the login page before data is dispatched to the thunk.

Best to keep all data modification in one place, otherwise we have to look in two places when figuring out how the data is maniuplated as a whole

Server/validation/joi.js
```diff
@@ -35,7 +35,7 @@ const registerValidation = joi.object({
     .string()
     .required()
     .pattern(/^[A-Za-z]+$/),
-  email: joi.string().email().required(),
+  email: joi.string().email().lowercase().required(),
```
ajhollid, line 38: solid :+1:

Server/package.json
```diff
@@ -11,8 +11,8 @@
   "author": "",
   "license": "ISC",
   "dependencies": {
-    "@sendgrid/mail": "^8.1.3",
-    "axios": "^1.7.2",
+    "@sendgrid/mail": "^7.7.0",
```
ajhollid, line 14: Wee can drop sendgrid entirely

ajhollid (changes requested):
It would be ideal to handle all input manipulations in their respective pages in order to 

1.  Keep the reducers/thunks pure
2.  Keep all data operations in one place so we're not hopping around looking for changes

PR #625, opened 2024-08-14
App layout redesign

danielcj2:
- [x] Replaced `grid` with `flexbox`.
- [x] Removed `navBar` files.
- [x] Moved profile menu to `sidebar`.
- [x] Limited main content to `1400px`.
- [x] Set different profile background color based on the user's name.   

Padding and margins may be off in some pages, something to look out for. I might've missed some.

image
(would recommend opening the screenshot in another page for a more accurate view)


Client/src/Components/Sidebar/index.jsx
```diff
@@ -32,10 +34,87 @@ const menu = [
...
+  /**
+   * Handles logging out the user
+   *
+   */
+  const logout = async () => {
+    // Clear auth state
+    dispatch(clearAuthState());
+    dispatch(clearUptimeMonitorState());
+    // Make request to BE to remove JWT from user
+    await axiosIntance.post(
```
ajhollid, line 83: Axios import is missing, so logout is failing with an uncaught error right now

Client/src/Components/Sidebar/index.jsx
```diff
@@ -32,10 +34,87 @@ const menu = [
...
+      "/auth/logout",
+      { email: authState.user.email },
+      {
+        headers: {
+          Authorization: `Bearer ${authState.authToken}`,
+          "Content-Type": "application/json",
+        },
+      }
+    );
```
ajhollid, line 92: There's no error handling for this Axios request, it should be surrounded in a `try catch` block so we don't have uncaught errors

ajhollid (changes requested):
Really nice looking refactor!  

Logout function is broken due to missing import and the axios request should be in a try/catch block.

Mobile support is also pretty broken at the moment, this is TODO?

image


Nice work overall, thanks for the hard work and many refactors!

gorkem-bwl:
> Mobile support is also pretty broken at the moment, this is TODO?

Your/Daniel's but my preference would be to push back mobile support as much as we can, until we have our first stable release :)

alvarotorrestx:
Also found myself in a bug, after testing the logout, I successfully reached the login page. But the system still recognizes me as logged in and gives the error of "User logged in."

Client/src/Components/Sidebar/index.jsx
danielcj2: Woops 😬 
Added the missing import.

Client/src/Components/Sidebar/index.jsx
danielcj2: Added `try catch` block.

ajhollid:
> > Mobile support is also pretty broken at the moment, this is TODO?
> 
> Your/Daniel's take but my preference would be to push back mobile support as much as we can, until we have our first stable release :)

Right you are boss!  Daniel and I discussed and we agree 👌

ajhollid (approved):

PR #628, opened 2024-08-14
Refactored notifications from Check model to NetworkService, resolves #627

### LINKED ISSUE #627: BE - Notifications could be better handled outside of the Check model
Notifications are currently issued from the `Check` model which tightly couples our database layer to our application logic.

This functionality can be refactored to a less tightly coupled implementation, probably in the `NetworkService` class.

ajhollid:
- [x] Inject `emailService` into `NetworkService`
- [x] Refacored `getMonitorById` to just return a monitor
- [x] Moved notificiaton logic from `Check` model to `NetworkService` to decouple functionality from DB layer


Skorpios604 (approved):
Looks good 👍👍

PR #636, opened 2024-08-14
fixed login email and register email case sensitivity issue.

Skorpios604:
- [x] Case normalization.


Server/validation/joi.js
```diff
@@ -35,7 +35,7 @@ const registerValidation = joi.object({
     .string()
     .required()
     .pattern(/^[A-Za-z]+$/),
-  email: joi.string().email().required(),
+  email: joi.string().email().lowercase().required(),
```
ajhollid, line 38: This validation isn't working, not sure why.

I can still create a user with uppercase chars in the email.

We'll have to investigate this further

Issue #498, opened 2024-07-30
FE - Monitor Sort Order


ajhollid:
Currently the monitors cannot be sorted:

![image](https://github.com/user-attachments/assets/faf16320-5238-48fd-90fa-0611622e1a29)

If we want to enable we can sort on either the client or the backend, with the usual tradeoffs.

**Client side**
- Downside: Potentially slow if many monitors
- Upside: No need for API call, easy to implement

**Server side** 
- Downside: Requires an API call, marginally more complicated to implement
- Upside: Client will not be blocked while waiting for data

Skorpios604:
Seems like a good idea. Some people might want to sort by which monitors are up or down. Like where that arrow is next to status. Users might intuitively click that thinking that it will sort the monitors for them. It is a very wide spread feature.

gorkem-bwl:
My take: let's go with the client side and if things got slow with hundreds of servers, we can easily switch (I hope that happens lol). Eventually we'll add filters on top of the table, helping users to view only the up or down monitors

gorkem-bwl:
Another take: The incidents page consists of more than 2500 entries and client side filtering seems to be fast enough. I am not in a position to compare the amount of data on the table (dashboard) and the table (incidents) but I tend to believe it shouldn't be an issue to go client side here.

ajhollid:
Alright! Let's do it client side and see how it works, always possible to change later if performance degrades.

PR #637, opened 2024-08-15
Added sorting ability by status

alvarotorrestx:
- Set default sorting of monitors in ascending order (Up status)
- Created handling of sorting when status clicked with image notifier (arrow)
- Tested with no monitor and pagination sorting (all success)
- Default monitor render will be by Up status
- This could potentially allow room for other sortings

Default
image

Descending
image

Pagination 1
image

Pagination 2
image

Reference to #498


ajhollid (changes requested):
Really nice PR, it's great to see that you've kept the size small and the scope very narrow and focused.  Fine example for all of us to follow!

I have a couple of suggestions that I left in my review, have a look and see what you think.  As always they are open to discussion so if anything doesn't make sense or I've missed the point somewhere please do point it out.

Please do update set up your linter/formatter to use double quotes for strings or there will be a _lot_ of changes happening to files as we go forward :joy:

alvarotorrestx:
Please see updated changes:

Default on load:
image

Ascending:
image

Descending:
image

ajhollid (approved):
Thanks for making requested changes :+1:

PR #667, opened 2024-08-17
Breadcrumbs

danielcj2:
Added breadcrumbs to:
- [x] Monitor
   - [x] Create
   - [x] Details
   - [x] Configure  
 - [x] Pagespeed
   - [x] Create
   - [x] Details
   - [x] Configure  
   
   
I removed the bar chart from monitor details page for now since the feature is not fully functional and we're likely going to move its placement and use a different chart library for it.


ajhollid (approved):
Great work!

PR #677, opened 2024-08-19
Sidebar collapse

danielcj2:
- [x] Collapsible sidebar
- [x] Mini nav menu opens on button click 
- [x] Tooltips

Expand transition isn't very smooth since the BWU logo enlarges instantly. Will revisit once we have an updated logo.


danielcj2:
Will likely need to refactor some of the code in the future since it got a bit out of hand.

ajhollid (commented):
Looks great!

Is it possible to have a smoother transition while collapsing?

image

It seems to pause in this state for a little while rather than smootly closing :thinking:

ajhollid:
Nice, tooltips look good :+1:

danielcj2:
> Is it possible to have a smoother transition while collapsing?
>
> It seems to pause in this state for a little while rather than smootly closing 🤔

Ideally, the sidebar would slide to the left as the text fades, without altering the position of the icons. But since we are using collapsible select menus, we would need to close the menus first (if open), and wait on that animation to finish before collapsing the sidebar which will cause a small delay. It's definitely doable just wasn't sure if that was something we'd want. If it's not a major issue right now I can revisit the animations and smoothen them out once we get a new app logo.

ajhollid:
> > Is it possible to have a smoother transition while collapsing?
> > It seems to pause in this state for a little while rather than smootly closing 🤔
> 
> Ideally, the sidebar would slide to the left as the text fades, without altering the position of the icons. But since we are using collapsible select menus, we would need to close the menus first (if open), and wait on that animation to finish before collapsing the sidebar which will cause a small delay. It's definitely doable just wasn't sure if that was something we'd want. If it's not a major issue right now I can revisit the animations and smoothen them out once we get a new app logo.

Sounds good to me, let's leave that for a separate PR

ajhollid (approved):

PR #687, opened 2024-08-20
Add Clone Monitor Feature, resolves #512

### LINKED ISSUE #512: Clone a monitor
Sometimes it is easier to clone a monitor than to add one from scratch. Just add a "clone" entry to the context menu of a monitor and then when it is selected, a new monitor is added by cloning the current one. 

This is a good first issue for new contributors.

geekananya:
### Description
Implemented cloning feature for monitors.

Closes #512. 

### Changes Made
- Add 'clone' option in monitor context menu
- Add clone route which redirects to Create Monitor page
- Fetch details of current monitor and set default values for the monitor


Client/src/App.jsx
```diff
@@ -47,6 +47,10 @@ function App() {
             path="/monitors/create"
             element={<ProtectedRoute Component={CreateMonitor} />}
           />
+          <Route
+            path="/monitors/clone/:monitorId"
+            element={<ProtectedRoute Component={CreateMonitor} />}
+          />
```
ajhollid, line 53: I don't think there is a need for a new route here, I think we can simply add an optional path param to `create`

`path="monitors/create/:monitorId?`

and resuse the create path.

Client/src/Pages/Monitors/CreateMonitor/index.jsx
```diff
@@ -40,6 +42,21 @@ const CreateMonitor = () => {
...
   const [errors, setErrors] = useState({});
 
+  useEffect(() => {
+    if(monitorId){
+      const data = monitors.find((monitor) => monitor._id === monitorId);
+      if (!data) {
+        console.error("Error fetching monitor of id: " + monitorId);
+        navigate("/not-found");
+      }
+      const {url, name, ...rest} = data;
```
ajhollid, line 52: I think we would want the `url` and `name` included in the new monitor as we are creating a copy, no need to exclude these.

Perhaps consider appending `" (Clone)"` or somethig similar to the name to signify that this is a copy

Client/src/Pages/Monitors/CreateMonitor/index.jsx
```diff
@@ -40,6 +42,21 @@ const CreateMonitor = () => {
   const [https, setHttps] = useState(true);
   const [errors, setErrors] = useState({});
 
+  useEffect(() => {
+    if(monitorId){
+      const data = monitors.find((monitor) => monitor._id === monitorId);
+      if (!data) {
+        console.error("Error fetching monitor of id: " + monitorId);
+        navigate("/not-found");
```
ajhollid, line 50: Please add the `{replace: true}` config object to `navigate` here so we don't get stuck in a loop if the user presses back form the 404 page.

ajhollid (changes requested):
Thanks for your contribution here, this is a nice solution to the issue!

A few minor changes and I think we can merge this in :+1:

ajhollid:
You can add "resolves #<issue_number> to your PR title to automatically close the issue that the PR resolves.

geekananya:
> You can add "resolves #<issue_number>" to your PR title to automatically close the issue that the PR resolves.

Got it, thanks! 
I'll apply the changes shortly.

ajhollid (approved):
Thanks for making the changes requested, looks like everything is functioning as expected!

Client/src/Pages/Monitors/CreateMonitor/index.jsx
```diff
@@ -40,6 +42,21 @@ const CreateMonitor = () => {
...
+      }
+      const {url, name, ...rest} = data;
+      rest.interval = rest.interval/60000;
```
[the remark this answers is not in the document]
geekananya: Understood!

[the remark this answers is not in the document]
ajhollid: No problem I snuck the change in :P

ajhollid:
Thanks again for your hard work and contribution, as soon as the other team members have a chance to review we should be able to merge this in :+1:

Skorpios604 (approved):

Client/src/Pages/Monitors/CreateMonitor/index.jsx
```diff
@@ -40,6 +42,23 @@ const CreateMonitor = () => {
...
 
+  useEffect(() => {
+    if(monitorId){
+      const data = monitors.find((monitor) => monitor._id === monitorId);
+      if (!data) {
+        console.error("Error fetching monitor of id: " + monitorId);
+        navigate("/not-found", {replace: true});
+      }
+      const {name, ...rest} = data;   //data.name is read-only
+      if(rest.type == 'http')  rest.url = rest.url.substring(8, rest.url.length)
```
ajhollid, line 53: I think a better approach here would be to create a [URL instance](https://developer.mozilla.org/en-US/docs/Web/API/URL/URL) and parse the host  from it. 

This current code will not parse the URL correctly if the monitor is http procotcol instead of https:

`https://www.google.com` -> `www.google.com`
`http://www.google.com` -> `ww.google.com`

ajhollid (changes requested):
Apologies, I just noticed there was a small issue with the parsing of the URL that needs to be addressed. Thank you!

Client/src/Pages/Monitors/CreateMonitor/index.jsx
geekananya: Thanks for identifying the issue! Creating a URL instance does seem like a better option, I'll make the required changes.

ajhollid:
Nice looks good!

We have moved the `/Monitors/index.jsx` to `/Monitors/Home/index.jsx` however, can you merge in our latest code to resolve the conflict?

You can copy your code from the old `index.jsx` to the new `index.jsx` and then just remove the old `index.jsx` from your commit.

Thank you!

ajhollid:
This PR has gone stale, I've created a new PR incorporating these changes and resolving merge conflicts

PR #688, opened 2024-08-20
Marshal data, logging it, creating post req, send to BE

MuhammadKhalilzadeh:
- Marshal data and log it
- Create post request
- Send post request to BE

### DIFF
--- a/Client/src/Components/Inputs/Field/index.jsx
+++ b/Client/src/Components/Inputs/Field/index.jsx
@@ -177,7 +177,14 @@ const Field = forwardRef(
 );
 
 Field.propTypes = {
-  type: PropTypes.oneOf(["text", "password", "url", "email", "description"]),
+  type: PropTypes.oneOf([
+    "text",
+    "password",
+    "url",
+    "email",
+    "description",
+    "number",
+  ]),
   id: PropTypes.string.isRequired,
   label: PropTypes.string,
   https: PropTypes.bool,
--- a/Client/src/Pages/Maintenance/CreateMaintenanceWindow/index.jsx
+++ b/Client/src/Pages/Maintenance/CreateMaintenanceWindow/index.jsx
@@ -10,6 +10,7 @@ import dayjs from "dayjs";
 import { DatePicker } from "@mui/x-date-pickers/DatePicker";
 import { MobileTimePicker } from "@mui/x-date-pickers/MobileTimePicker";
 import Field from "../../../Components/Inputs/Field";
+import axios from "axios";
 
 const directory = {
   title: "Create a maintenance window",
@@ -69,6 +70,29 @@ const CreateNewMaintenanceWindow = () => {
       ...prev,
       [name]: value,
     }));
+  };
+
+  const handleSubmit = async () => {
+    const data = {
+      repeat: values.repeat,
+      date: values.date.format("YYYY-MM-DD"),
+      startTime: values.startTime.format("HH:mm"),
+      duration: values.duration,
+      unit: values.unit,
+      friendlyName: values.friendlyName,
+      addMonitors: values.AddMonitors,
+    };
+
+    console.log("Submitting data: ", data);
+
+    await axios
+      .post("/address-of-the-route", data)
+      .then((res) => {
+        console.log("Response: ", res.data);
+      })
+      .catch((err) => {
+        console.log("Error: ", err);
+      });
   };
 
   const configOptions = [
@@ -246,6 +270,7 @@ const CreateNewMaintenanceWindow = () => {
             }}
             level="primary"
             label="Create"
+            onClick={handleSubmit}
           />
         </Stack>
       </Stack>
--- a/Client/src/Pages/Maintenance/CreateMaintenanceWindow/index.jsx
+++ b/Client/src/Pages/Maintenance/CreateMaintenanceWindow/index.jsx
@@ -163,6 +163,7 @@ const CreateNewMaintenanceWindow = () => {
             placeholder="60"
             onChange={(e) => handleChange(e, "duration")}
             value={values.duration}
+            error={errors.duration}
           />
           <Select
             onChange={(e) => handleChange(e, "unit")}
@@ -181,6 +182,7 @@ const CreateNewMaintenanceWindow = () => {
           placeholder="Maintanence at __ : __ for ___ minutes"
           value={values.friendlyName}
           onChange={(e) => handleChange(e, "friendlyName")}
+          error={errors.friendlyName}
         />
       ),
     },
@@ -197,6 +199,7 @@ const CreateNewMaintenanceWindow = () => {
             placeholder="Start typing to search for current monitors"
             value={values.AddMonitors}
             onChange={(e) => handleChange(e, "AddMonitors")}
+            error={errors.addMonitors}
           />
           <Typography
             sx={{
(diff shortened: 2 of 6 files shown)


Client/src/Pages/Maintenance/CreateMaintenanceWindow/index.jsx
```diff
@@ -10,6 +10,7 @@ import dayjs from "dayjs";
 import { DatePicker } from "@mui/x-date-pickers/DatePicker";
 import { MobileTimePicker } from "@mui/x-date-pickers/MobileTimePicker";
 import Field from "../../../Components/Inputs/Field";
+import axios from "axios";
```
ajhollid, line 13: We have a `NetworkService` class that provides an axios connection with the app base URL set up and methods for connecting to the backend.

We can remove this and import network service instead.  See other pages that make network requests for examples :+1:

Client/src/Pages/Maintenance/CreateMaintenanceWindow/index.jsx
```diff
@@ -71,6 +72,29 @@ const CreateNewMaintenanceWindow = () => {
...
+
+    await axios
+      .post("/address-of-the-route", data)
+      .then((res) => {
+        console.log("Response: ", res.data);
+      })
+      .catch((err) => {
+        console.log("Error: ", err);
+      });
+  };
```
ajhollid, line 96: We can remove this for now, let's just marshal the data correctly first and make sure it is validated

Client/src/Pages/Maintenance/CreateMaintenanceWindow/index.jsx
```diff
@@ -71,6 +75,39 @@ const CreateNewMaintenanceWindow = () => {
...
+    const { error } = maintenanceWindowValidation.validate(data, {
+      abortEarly: false,
+    });
+    logger.log("error: ", error);
+    if (error) {
+      const newErrors = {};
+      error.details.forEach((err) => {
+        newErrors[err.path[0]] = err.message;
+      });
+      setErrors(newErrors);
```
ajhollid, line 98: Each `Field` should be updated with errors so that it's error state is shown to the user:

image

The error message should also be a user friendly error message, ie instead of `friendlyName cannot be empty`  it should be something like

"Friendly Name is required"

Please see other components and validation for examples.

[Login Page](https://github.com/bluewave-labs/bluewave-uptime/blob/develop/Client/src/Pages/Auth/Login.jsx) and it's associated validation is a good straightforward example.

Client/src/Pages/Maintenance/CreateMaintenanceWindow/index.jsx
```diff
@@ -71,6 +75,39 @@ const CreateNewMaintenanceWindow = () => {
     }));
   };
 
+  const handleSubmit = async () => {
+    const data = {
+      repeat: values.repeat,
+      date: values.date.format("YYYY-MM-DD"),
+      startTime: values.startTime.format("HH:mm"),
+      duration: values.duration,
```
ajhollid, line 83: If the input here can only be numbers we should restrict the input type to that?  Doesn't make sense to allow the user to enter non-numerical characters if they're going to be rejected by vaildation anyway :thinking:

ajhollid (changes requested):
The

Client/src/Pages/Maintenance/CreateMaintenanceWindow/index.jsx
MuhammadKhalilzadeh: Done ✅

Screenshot 2024-08-21 224312

Client/src/Pages/Maintenance/CreateMaintenanceWindow/index.jsx
MuhammadKhalilzadeh: Done ✅

Screenshot (115)

MuhammadKhalilzadeh: Screenshot 2024-08-21 224843

Skorpios604: Maybe for this part say something like A monitor is required. Add monitors is required sounds kind of robotic lol.

ajhollid (changes requested):
image


I can still enter letters for duration?

And the last field doesn't clear the error state even if something is entered :thinking:

MuhammadKhalilzadeh:
@ajhollid 

Both bugs are fixed now, but, now that I've changed the field type for duration to "number" to just get numeric values, the width has changed, and styles I add for it width does not change the width. Should I even consider that a bug?

Screenshot 2024-08-22 100015

ajhollid:
> @ajhollid 
> 
> Both bugs are fixed now, but, now that I've changed the field type for duration to "number" to just get numeric values, the width has changed, and styles I add for it width does not change the width. Should I even consider that a bug?
> 
> Screenshot 2024-08-22 100015
> 

I'm not sure about that one, I suggest consulting with the other FE guys, they'll know more about that than me 👍

MuhammadKhalilzadeh:
@Skorpios604 @danielcj2 
What do you think guys?

danielcj2:
> @Skorpios604 @danielcj2 What do you think guys?

Seems like the width you're trying to set is being overwritten by another of your css selectors.

image

To fix it, all you need to do is prepend the parent class `.maintenance-options`. Like so: 

```
.duration-config .field-text .MuiTextField-root, 
.maintenance-options .duration-config .field {
    width: 70px;
    min-width: 70px;
    max-width: 70px;
}
```

MuhammadKhalilzadeh:
@danielcj2 
Done. Fixed 👍🏻✅

ajhollid (changes requested):
image

Page is throwing an error, it looks like the proptypes for the `Field` component need to be updated.

Please check for all errors before requesting PR reviews, there should be no errors in the console and everything should be working as expected before requesting review :+1:

MuhammadKhalilzadeh:
The reason for this error was that in the implementation of the Field component, the type number, was not specified or added to the list of the types Field component supports. I added it, and now its working.

Screenshot 2024-08-24 100947

Client/src/Components/Inputs/Field/index.jsx
```diff
@@ -177,7 +177,14 @@ const Field = forwardRef(
...
 
 Field.propTypes = {
-  type: PropTypes.oneOf(["text", "password", "url", "email", "description"]),
+  type: PropTypes.oneOf([
+    "text",
+    "password",
+    "url",
+    "email",
+    "description",
+    "number",
```
ajhollid, line 186: Great, thanks for adding this

Client/src/Pages/Maintenance/CreateMaintenanceWindow/index.jsx
```diff
@@ -71,6 +75,41 @@ const CreateNewMaintenanceWindow = () => {
...
+      createToast({
+        body:
+          error.details && error.details.length > 0
+            ? error.details[0].message
+            : "Error validating data",
+      });
+      logger.error("Validation errors:", error.details);
+    }
+
+    logger.log("Submitting data: ", data);
```
ajhollid, line 110: When we're ready to submit interval value has to be convereted to miliseconds, we can do that in the next PR though

ajhollid (approved):
It looks like everything is working now :+1: 

When it's time to submit the data the duration needs to be converted to miliseconds, but we can do that in the next PR.

Client/src/Pages/Maintenance/CreateMaintenanceWindow/index.jsx
MuhammadKhalilzadeh: great, then let's go to the next level

Skorpios604 (approved):

PR #692, opened 2024-08-20
UI slice

### LINKED ISSUE #685: FE - Create utility redux slice
Will manage various application states that are essential for UI interactions and UX. 
The slice will include:

- Table states (more specifically `rowsPerPage`)
- Sidebar (collapsed state)
- Light / Dark Mode (when we eventually get to it)
etc. 

Will add more as we go.

danielcj2:
Added a redux slice to manage UI preferences, such as light/dark mode, table row settings, sidebar collapse state, and other similar features. Currently there's only 2 actions implemented. Will add more as we go.

- [x] Implemented UI slice
- [x] Added slice to `redux-persist` 
- [x] Implemented UI actions
  - Change rows per page
  - Toggle sidebar collapsed state     

Resolves #685

### DIFF
--- a/Client/src/Features/UI/uiSlice.js
+++ b/Client/src/Features/UI/uiSlice.js
@@ -1,0 +1,22 @@
+import { createSlice } from "@reduxjs/toolkit";
+
+const initialState = {
+  table: {
+    rowsPerPage: 5,
+  },
+  sidebar: {
+    collapsed: false,
+  },
+};
+
+const uiSlice = createSlice({
+  name: "ui",
+  initialState,
+  reducers: {
+    setRowsPerPage: (state, action) => {},
+    setCollapsed: (state, action) => {},
+  },
+});
+
+export default uiSlice.reducer;
+export const { setRowsPerPage, setCollapsed } = uiSlice.actions;
--- a/Client/src/Features/UI/uiSlice.js
+++ b/Client/src/Features/UI/uiSlice.js
@@ -3,7 +3,10 @@ import { createSlice } from "@reduxjs/toolkit";
 // Initial state for UI settings.
 // Add more settings as needed (e.g., theme preferences, user settings)
 const initialState = {
-  table: {
+  monitors: {
+    rowsPerPage: 5,
+  },
+  team: {
     rowsPerPage: 5,
   },
   sidebar: {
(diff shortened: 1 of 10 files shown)


Client/src/Features/UI/uiSlice.js
```diff
@@ -0,0 +1,28 @@
...
+// Initial state for UI settings.
+// Add more settings as needed (e.g., theme preferences, user settings)
+const initialState = {
+  table: {
+    rowsPerPage: 5,
+  },
+  sidebar: {
+    collapsed: false,
+  },
+};
```
ajhollid, line 12: Is there only ever going to be one table whose state we need to keep track of?

Maybe we should make this more extensible by having a state object?

```
{
   montiorsTable: { rowsPerPage: 5}
   otherTable: {rowsPerPage: 10}
}
```

What do you think?

Either that or separate slices for each?

ajhollid (commented):
Question about future tables :thinking:

Client/src/Features/UI/uiSlice.js
danielcj2: > Is there only ever going to be one table whose state we need to keep track of?

Not sure, it depends if we want to add the rows functionality to any other tables. 

>Maybe we should make this more extensible by having a state object?

Yea I agree.

>Either that or separate slices for each?

I don't think we need separate slices for each table, we won't be tracking too many variables. All I can think of right now is `rowsPerPage` and `sortOrder`.

ajhollid: OK lets assume we'll have other tables to track in the future, seems more likely than not that it will come up.

Looks like that state object is the simplest way to go

danielcj2:
The fix is a little hacky and will not work for tables where the route is ends with `monitor id`. Will need to figure out a better implementation.

ajhollid (approved):
Nice, looking good on my end, thanks for revisions :+1:

Client/src/store.js
```diff
@@ -14,7 +15,7 @@ const authTransform = createTransform(
   // No transformation on rehydration
   null,
   // Only applies to auth
-  { whitelist: ["auth"] }
+  { whitelist: ["auth", "ui"] }
 );
```
ajhollid, line 19: Just noticed this, the whitelist should go under the persistconfig below I believe, this transform was for the image

ajhollid (changes requested):
Whitelist

Client/src/store.js
danielcj2: Good catch

ajhollid (approved):

PR #695, opened 2024-08-21
Status Sorting w/ localStorage

alvarotorrestx:
Reference #498 

**Changes:**
- Recreated sort feature for Status button
- Added icon to show Status is sortable + Update icon depending on sorting
- Used localStorage to save sorted order if changed
- Updates/saves sort order on adding monitor or removing

**Comments:**
I apologize this took longer than expected, I was trying to achieve a different approach without using localStorage but this worked best for me. I'm open to suggestions!

Additionally, I think adding a button or some way to clear the sorting may be an option as well. Thoughts?

**Pictures:**
Default:
image

Ascending:
image
localStorage:
image

Descending:
image
localStorage:
image


ajhollid (commented):
First off thanks for getting the feature in!

What was your approach previous to using local storage?

I'm not sure we need to persist the sort order, which it looks like local storage is being used for here.

I don't think it's a big deal if the sort order resets when the page is refreshed. I would say that could be considered expected behavior 🤔

gorkem-bwl:
> First off thanks for getting the feature in!
> 
> What was your approach previous to using local storage?
> 
> I'm not sure we need to persist the sort order, which it looks like local storage is being used for here.
> 
> I don't think it's a big deal if the sort order resets when the page is refreshed. I would say that could be considered expected behavior 🤔

My personal take would be “Oh, when I refresh the page my favorite sorting behavior still persists and that’s cool from a UX standpoint”. So for me: If it’s there then it’s convenient, if not, no very big deal.

alvarotorrestx:
Right, I was looking at it for a UX point of view. Constantly having to re-sort can be annoying.

As for my original approach, the only issue I had was going from Monitor/Create and back to /Monitor. It would cause a re-render on the display monitors and wasn't saving the previous state of sorting. I tried used useLocation from react-router-dom but either I wasn't using it properly or it just wasn't the right approach.

@gorkem-bwl @ajhollid

Client/src/Pages/Monitors/index.jsx
```diff
@@ -152,17 +154,52 @@ const Monitors = () => {
...
+      setSortOrder(parsedSortOrder);
+      const monitors = [...monitorState.monitors].sort(a => {
+        return a.status ? -1 * parsedSortOrder : 1 * parsedSortOrder;
+      });
+      setSortedMonitors(monitors);
+      setIsSorted(true);
+    } else {
+      setSortedMonitors(monitorState.monitors);
+    }
+  }, [monitorState.monitors]);
```
ajhollid, line 175: I think we can just drop this entirely, I don't think we need to persist the sort order through reloads of the page.

If we _did_ decide we wanted to do that we should use the Redux state for that in any case.

Fow now, I think safe to just drop this

Client/src/Pages/Monitors/index.jsx
```diff
@@ -152,17 +154,52 @@ const Monitors = () => {
...
+  
+  
+  const handleSort = () => {
+    const newSortOrder = sortOrder * -1;
+    setSortOrder(newSortOrder);
+  
+    const monitors = [...monitorState.monitors].sort(a => {
+      return a.status ? -1 * newSortOrder : 1 * newSortOrder;
+    });
```
ajhollid, line 184: I think this can be simplified somewhat to

```
    setSortOrder(-sortOrder);

    const monitors = [...monitorState.monitors].sort((a) => {
        return a.status ? -sortOrder : sortOrder;
    });
    ```
I don't think the `newSortOrder` var is necessary here.  If I've missed something let me know :thinking: 
    
We can save a bit of verbosity by directly inverting the sort order with the `-` operator too instead of multiplying by `-1`

May as well simplify wherever we can :joy:

ajhollid (changes requested):
Nice work, everything is working as expected!

I think it's safe to drop the `localStorage` implementation for now, I don't think persisting the sort order is necessary.  If we do decide we want to do that we should use the Redux state anyway since that's where we are handling persistence.

We'll have to test this out again with a large dataset and see how the sorting goes.  Sorting is a reasonable O(n log n) but even that can be noticeable with large numbers.

Thanks for you hard work here, just a couple minor changes and we're good to go!

Client/src/Pages/Monitors/index.jsx
alvarotorrestx: Thanks for this, you taught me something! I went ahead and used this logic and it works flawlessly. Went ahead and removed the localStorage as well.

ajhollid (approved):
Looks good to me, thanks for revisions!

ajhollid:
FE has changed conisderably since this PR, we can revisit later

Issue #725, opened 2024-08-26
BE - Teams - Creating admin account, team, and creating monitors


ajhollid:
1. Create a team table
2. When an admin account is created, create a team and associate it with the admin
3. When a monitor is created, associate it with the team

PR #728, opened 2024-08-26
Feat/team table, #725

ajhollid:
This is quite a large PR as it is the basis of implementing teams

- [x] Add a team table
- [x] Create a new superadmin role
- [x] Add superadmin role to first user created
- [x] Add teamId to users

### DIFF
--- a/Server/db/mongo/MongoDB.js
+++ b/Server/db/mongo/MongoDB.js
@@ -15,9 +15,9 @@ const connect = async () => {
   }
 };
 
-const checkAdmin = async (req, res) => {
+const checkSuperadmin = async (req, res) => {
   try {
-    const admin = await UserModel.findOne({ role: "admin" });
+    const admin = await UserModel.findOne({ role: "superadmin" });
     if (admin !== null) {
       return true;
     }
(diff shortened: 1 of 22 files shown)


Server/models/Team.js
```diff
@@ -0,0 +1,12 @@
+const mongoose = require("mongoose");
+
+const TeamSchema = mongoose.Schema({
+  userId: {
+    type: mongoose.Schema.Types.ObjectId,
+    ref: "User",
+    immutable: true,
+  },
+  timestamps: true,
```
Skorpios604: I believe this should be 
```javascript
{ timestamps: true }
```
to automatically create the 'createdAt' and 'updatedAt' fields for your documents.

The way it is now I believe it just stores the value true in each document which is not the intended behavior I believe.

Server/db/mongo/MongoDB.js
```diff
@@ -15,9 +15,9 @@ const connect = async () => {
   }
 };
 
-const checkAdmin = async (req, res) => {
+const checkSuperadmin = async (req, res) => {
   try {
-    const admin = await UserModel.findOne({ role: "admin" });
+    const admin = await UserModel.findOne({ role: "superadmin" });
```
Skorpios604, line 20: Should we rename this const to Superadmin so that it is more clear?

Server/models/Team.js
ajhollid: I think I got this one in a later commit should be good now

Skorpios604 (approved):

MuhammadKhalilzadeh (approved):

PR #740, opened 2024-08-28
autocomplete implementation

MuhammadKhalilzadeh:
Implementation for the Autocomplete Component

- [x] Takes an array of data in a standardized format
- [x] Displays error status if this is a required field

Screenshot 2024-08-28 132521
Screenshot 2024-08-28 132532

### DIFF
--- a/Client/package.json
+++ b/Client/package.json
@@ -20,6 +20,7 @@
     "@mui/x-data-grid": "7.3.2",
     "@mui/x-date-pickers": "7.3.2",
     "@reduxjs/toolkit": "2.2.5",
+    "autosuggest-highlight": "^3.3.4",
     "axios": "^1.7.4",
     "chart.js": "^4.4.3",
     "dayjs": "1.11.11",
--- a/Client/src/App.jsx
+++ b/Client/src/App.jsx
@@ -27,7 +27,6 @@ import CreatePageSpeed from "./Pages/PageSpeed/CreatePageSpeed";
 import CreateNewMaintenanceWindow from "./Pages/Maintenance/CreateMaintenanceWindow";
 import PageSpeedDetails from "./Pages/PageSpeed/Details";
 import PageSpeedConfigure from "./Pages/PageSpeed/Configure";
-import ComponentText from "./Pages/ComponentText";
 
 function App() {
   const AdminCheckedRegister = withAdminCheck(Register);
--- a/Client/src/App.jsx
+++ b/Client/src/App.jsx
@@ -27,6 +27,7 @@ import CreatePageSpeed from "./Pages/PageSpeed/CreatePageSpeed";
 import CreateNewMaintenanceWindow from "./Pages/Maintenance/CreateMaintenanceWindow";
 import PageSpeedDetails from "./Pages/PageSpeed/Details";
 import PageSpeedConfigure from "./Pages/PageSpeed/Configure";
+import ComponentText from "./Pages/ComponentText";
 
 function App() {
   const AdminCheckedRegister = withAdminCheck(Register);
--- a/Client/src/Components/Inputs/Autocomplete/index.jsx
+++ b/Client/src/Components/Inputs/Autocomplete/index.jsx
@@ -1,0 +1,139 @@
+import { Autocomplete, Stack, Typography } from "@mui/material";
+import "./index.css";
+import React from "react";
+import PropTypes from "prop-types";
+import parse from "autosuggest-highlight/parse";
+import match from "autosuggest-highlight/match";
+
+/**
+ * AutoCompleteField component.
+ *
+ * @component
+ * @param {Object} props - The component props.
+ * @param {string} props.id - The ID of the autocomplete field.
+ * @param {string} props.label - The label of the autocomplete field.
+ * @param {React.Element} props.icon - The icon element to display.
+ * @param {string} props.type - The type of the input field (text or number).
+ * @param {Array} props.options - The options for the autocomplete field.
+ * @param {string} props.error - The error message to display.
+ * @param {string} props.placeholder - The placeholder text for the input field.
+ * @param {boolean} props.isRequired - Indicates if the field is required.
+ * @param {function} props.onChange - The function to handle onChange event.
+ * @param {string} props.value - The value of the input field.
+ * @param {boolean} props.disabled - Indicates if the field is disabled.
+ * @param {string} props.setWidth - The width of the input field.
+ * @returns {JSX.Element} The AutoCompleteField component.
+ */
+
+const AutoCompleteField = ({
+  id,
+  label,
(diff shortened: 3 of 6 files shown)


gorkem-bwl:
Can you please change the dropdown design so it matches our UI style?

Client/package.json
```diff
@@ -20,6 +20,7 @@
     "@mui/x-data-grid": "7.3.2",
     "@mui/x-date-pickers": "7.3.2",
     "@reduxjs/toolkit": "2.2.5",
+    "autosuggest-highlight": "^3.3.4",
```
ajhollid, line 23: What is this package?  We shouldn't need to add any libraries this component, MUI should have everything we need.

Introducing more packages makes the app larger and more vulnerable so we should do so only if we absolutely need to

MuhammadKhalilzadeh:
@gorkem-bwl Could you please give me some more details for the parts I need to modify? since first I tried to follow what's been done in the Text Fields.
Which parts and which details you want me to change or add to?

Client/src/App.jsx
```diff
@@ -27,6 +27,7 @@ import CreatePageSpeed from "./Pages/PageSpeed/CreatePageSpeed";
 import CreateNewMaintenanceWindow from "./Pages/Maintenance/CreateMaintenanceWindow";
 import PageSpeedDetails from "./Pages/PageSpeed/Details";
 import PageSpeedConfigure from "./Pages/PageSpeed/Configure";
+import ComponentText from "./Pages/ComponentText";
```
ajhollid, line 30: This doesn't resolve and app won't run as a result, I assume this is from testing?  It can be removed

Client/src/Components/Inputs/Autocomplete/index.jsx
```diff
@@ -0,0 +1,139 @@
...
+          width: "fit-content",
+          padding: "8px",
+          borderRadius: "var( --env-var-radius-1)",
+          border: isRequired ? "1px solid #d32f2f" : "1px solid #D0D5DD",
+          alignItems: "center",
+        }}
+        gap={1}
+        direction="row"
+        alignItems="center"
+      >
```
ajhollid, line 66: This component should just be a simple wrapper for the Autocomplete component, it shouldn't contain other structural elements like `Stack`.  Those should be implemented by the user of `Autocomplete`

All that should happen in this component is that it should take an `Array<data>` as a Prop and return a styled MUI Autocomplete component with that data.

ajhollid (changes requested):
It's a good start, but I think this component has more strucutre than it should.

It should be a simple wrapper for an MUI Autocomplete component.

The wrapper should take an `Array<data>` as a prop and return a styled MUI autocomplete component.

Structure like `Stacks` can be left to the user of the component to implement at a higher level

Client/package.json
MuhammadKhalilzadeh: This package what in the doc itself! without this, the dropdown on the autocomplete does not work!

gorkem-bwl:
> @gorkem-bwl Could you please give me some more details for the parts I need to modify? since first I tried to follow what's been done in the Text Fields. Which parts and which details you want me to change or add to?

Sure. For example check the dropdown component here: 

Screenshot 2024-08-28 at 1 43 00 PM

Basically, the look and feel of your dropdown component should be matching this component. This is where you need to change. I am just looking at it from the UI point of view by the way.

Client/package.json
MuhammadKhalilzadeh: The `Autocomplete` Component is a wrapper itself. What lets you type and to get your options is simply a Text Field that is passed to the `renderInput`. The doc used the MUI's `TextField` for in its example which made it unlikely to change in its styles, neither our Field component could be passed to it! So, I passed an `input` element to it to make it as light and editable as possible.

I had to wrap it inside Stacks to add validations `errors`, `helperText`, and `title`, in case we want to have such things added in the future.

Client/src/App.jsx
MuhammadKhalilzadeh: Oh, it was for the purpose of test!
Forgot to remove it!

MuhammadKhalilzadeh:
@gorkem-bwl 
Got it 👌

Client/src/Components/Inputs/Autocomplete/index.jsx
MuhammadKhalilzadeh: The `Autocomplete` is a wrapper itself.
[Check this example please](https://mui.com/material-ui/react-autocomplete/#customization)

Client/package.json
ajhollid: > This package what in the doc itself! without this, the dropdown on the autocomplete does not work!

The example in the docs for a controlled component doesn't seem to require that?

```
import * as React from 'react';
import TextField from '@mui/material/TextField';
import Autocomplete from '@mui/material/Autocomplete';

const options = ['Option 1', 'Option 2'];

export default function ControllableStates() {
  const [value, setValue] = React.useState(options[0]);
  const [inputValue, setInputValue] = React.useState('');

  return (
    <div>
      <div>{`value: ${value !== null ? `'${value}'` : 'null'}`}</div>
      <div>{`inputValue: '${inputValue}'`}</div>
      <br />
      <Autocomplete
        value={value}
        onChange={(event, newValue) => {
          setValue(newValue);
        }}
        inputValue={inputValue}
        onInputChange={(event, newInputValue) => {
          setInputValue(newInputValue);
        }}
        id="controllable-states-demo"
        options={options}
        sx={{ width: 300 }}
        renderInput={(params) => <TextField {...params} label="Controllable" />}
      />
    </div>
  );
}
```

ajhollid: > I had to wrap it inside Stacks to add validations `errors`, `helperText`, and `title`, in case we want to have such things added in the future.

I don't think we need any of those things here :thinking:   I think we can safely remove them and simplify here

MuhammadKhalilzadeh:
@gorkem-bwl 
@ajhollid 

I just made some changes to both code and the looks of the the component:

Here's the overall look
Screenshot 2024-08-28 232210

Here are parts that I could not change. 
The `color`, `fontFamily`, `fontSize`, `border` styles are all applied same as other parts of or UI components.
Whenever the dropdown opens and I try to inspect it to see what classes to target, the dropdown closes.

Screenshot 2024-08-28 232224

Client/src/Components/Inputs/Autocomplete/index.jsx
ajhollid: That's fine, we still need a wrapper for the `Autocomplete` component so that we can pass it data and style it as we would like.

gorkem-bwl:
I think it could use some additional styling to ensure consistency throughout. Could you please reach out to Daniel for assistance?

PR #742, opened 2024-08-28
Add status codes and messages to checks

ajhollid:
This PR adds status codes and messages to all checks

- [x] Parse Axios responses for status codes
- [x] Grab messages from HTTP library based on status codes
- [x] Add messages to HTTP, Ping, and PageSpeed monitors
- [x] Add status codes to HTTP and PageSpeed monitors


MuhammadKhalilzadeh (approved):
👍🏻

PR #745, opened 2024-08-28
Add check for monitor type to abort certificate check if not http type, resolves #720

### LINKED ISSUE #720: BE - Don't perform certificate check if monitor is ping type
Only perform certificate checks for HTTP type monitors

ajhollid:
- [x] Add a check to abort certificate check early if not http type monitor


Skorpios604 (approved):

alvarotorrestx (approved):
Seems to work on my end 🫡

PR #749, opened 2024-08-28
Add pause routes and controller, implement functionality

ajhollid:
This PR implements pause functionality for monitors

- [x] Add route for pause
- [x] Add controller for pause
- [x] Implement logic for adding and removing jobs from the queue
- [x] Implement FE logic for properly displaying pause/resume
- [x] Implement FE request to /api/v1/monitor/pause/:monitorId to pause/resume monitor

### DIFF
--- a/Client/src/Pages/Monitors/Home/index.jsx
+++ b/Client/src/Pages/Monitors/Home/index.jsx
@@ -24,11 +24,20 @@ const Monitors = ({ isAdmin }) => {
     dispatch(getUptimeMonitorsByTeamId(authState.authToken));
   }, [authState.authToken, dispatch]);
 
-  const up = monitorState.monitors.reduce((acc, cur) => {
-    return cur.status === true ? acc + 1 : acc;
-  }, 0);
+  const monitorStats = monitorState.monitors.reduce(
+    (acc, monitor) => {
+      if (monitor.isActive === false) {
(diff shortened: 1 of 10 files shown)


Client/src/Pages/Monitors/Home/index.jsx
```diff
@@ -24,11 +24,20 @@ const Monitors = ({ isAdmin }) => {
...
   }, [authState.authToken, dispatch]);
 
-  const up = monitorState.monitors.reduce((acc, cur) => {
-    return cur.status === true ? acc + 1 : acc;
-  }, 0);
+  const monitorStats = monitorState.monitors.reduce(
+    (acc, monitor) => {
+      if (monitor.isActive === false) {
+        acc["paused"] += 1;
+      } else if (monitor.status === true) {
```
Skorpios604, line 31: should we rename this to isUp instead of status?

Skorpios604 (approved):

Client/src/Pages/Monitors/Home/index.jsx
ajhollid: Yes something along those lines would make more sense, but that's a fairly large refactor.

We should open a separate issue and PR for that

Skorpios604: Sounds good to me!

PR #757, opened 2024-08-30
fixed skeleton from bad merge conflict resolution

ajhollid:
PR resolves an issue caused by a mistake in a precious conflict resolution.  Incoming change was accepted instead of current change, now corrected.

- [x] Use correct skeleton


ajhollid:
Resolved in another PR

PR #770, opened 2024-08-31
Feat/theme button styles

danielcj2:
- [x] Refactored button components
- [x] Defined button variants in the theme file  
- [x] Fixed minor theme issues


ajhollid (changes requested):
Looks good to me, but the create monitor button's color isn't correct anymore:

image

That's the only one I've found

danielcj2:
> Looks good to me, but the create monitor button's color isn't correct anymore:
> 
> image
> 
> That's the only one I've found

Woops, added the wrong variant prop. Fixed now.

Skorpios604 (approved):

ajhollid (approved):
Thanks for making changes!

PR #774, opened 2024-09-01
Autocomplete

MuhammadKhalilzadeh:
I closed the pr and created this one since some conflicts appeared after other branches got merged.
Now, after fetching upstream for a couple of times and making sure everything is up-to-date, here's what I have for the dropdown part:

Screenshot 2024-09-01 091258


ajhollid (changes requested):
I think it would be best to consult further with @danielcj2  on the styles here, we need to use the Theme variables and not the global CSS vars anymore.

As for the logic of the component, it's deeply tied to the dummy data in the file, so it won't work for other data sets.

I assume this is a work in progress, so we can take a closer look when the component is ready to accept a more generalized data set.

MuhammadKhalilzadeh:
New looks everyone!

Screenshot 2024-09-02 230649

Screenshot 2024-09-02 233423

Screenshot 2024-09-02 233355

Screenshot 2024-09-02 231201

Screenshot 2024-09-02 230704

Screenshot 2024-09-02 230656

Screenshot 2024-09-02 230710

Screenshot 2024-09-02 230718

MuhammadKhalilzadeh:
What do you think guys?
@gorkem-bwl  
@ajhollid 
@danielcj2 
@Skorpios604

gorkem-bwl:
That looks good to me. Can you remove the scrollbar and add 'Type to search' inside the search bar?

Client/src/Components/Inputs/Autocomplete/index.css
```diff
@@ -0,0 +1,9 @@
+.auto-complete-field div {
+  width: var(--env-var-width-2);
+  height: 34px;
+  border-radius: 4px;
+}
+
+.auto-complete-field div div .MuiOutlinedInput-root {
+  padding-left: 0;
+}
```
ajhollid, line 9: I believe we should be using the theme for most if not all of these values?

Client/src/Components/Inputs/Autocomplete/index.jsx
```diff
@@ -0,0 +1,104 @@
...
+          sx: {
+            marginTop: theme.spacing(2),
+            border: 1,
+            borderColor: theme.palette.border.light,
+            borderRadius: theme.shape.borderRadius,
+            boxShadow: theme.shape.boxShadow,
+            backgroundColor: theme.palette.background.main,
+            paddingY: 0,
+            paddingX: 2,
+            fontSize: "var(--env-var-font-size-medium)",
```
ajhollid, line 85: Font sizes should be specified in the theme as well

Client/src/Components/Inputs/Autocomplete/index.jsx
```diff
@@ -0,0 +1,104 @@
+import { Autocomplete, Stack, TextField, Typography } from "@mui/material";
+import "./index.css";
+import React from "react";
+import PropTypes from "prop-types";
+import parse from "autosuggest-highlight/parse";
+import match from "autosuggest-highlight/match";
```
ajhollid, line 6: I'm not sure highlighting the text as the user types is worth adding a dependency and importing another library, small as it may be.

I think filtering out options that don't match the user's text is probably sufficicent.

Client/src/Components/Inputs/Autocomplete/index.jsx
```diff
@@ -0,0 +1,104 @@
...
+        const matches = match(option.name, inputValue, {
+          insideWords: true,
+        });
+        const parts = parse(option.name, matches);
+
+        return (
+          <li key={key} {...optionProps}>
+            <div>
+              {parts.map((part, index) => (
+                <span key={index}>{part.text}</span>
```
ajhollid, line 68: Index [shouldn't be used as a key](https://react.dev/learn/rendering-lists#rules-of-keys) when rendering lists:

>You might be tempted to use an item’s index in the array as its key. In fact, that’s what React will use if you don’t specify a key at all. But the order in which you render items will change over time if an item is inserted, deleted, or if the array gets reordered. Index as a key often leads to subtle and confusing bugs.

ajhollid:
Looks pretty nice to me visually! Code-wise I left a review with some suggestions to think about

MuhammadKhalilzadeh:
@gorkem-bwl 

'Type to search' text for placeholder, Done ✅

Now, about the scroller, similar to what we saw on the monitors page, this scroller appears only because the dropdown is containing the whole list, is it appears by default anyway! but when you start typing, the options get more and more limited and the scroller disappears. Now, what we discussed so far as our only solution with @danielcj2  to resolve this, is to redesign the scroller, so that the scroller fits our UI anywhere else on our app.

Screenshot 2024-09-03 093137
Screenshot 2024-09-03 093102
Screenshot 2024-09-02 231201

Client/src/Components/Inputs/Autocomplete/index.css
MuhammadKhalilzadeh: @ajhollid 

This is the smallest I could write this code. Neither of these classes could be targeted from the `jsx` file.
Wherever I added the `height` and removed it from here, the effect was gone and the height of the autocomplete got bigger.

Like this:

Screenshot 2024-09-03 122610

And about the padding, this classes is a nested class inside the `TextField` component and can not be targeted by the properties provides by the component itself, as below:

Screenshot 2024-09-03 122936
Screenshot 2024-09-03 122907

Do you have any ways for fixing this?

Client/src/Components/Inputs/Autocomplete/index.jsx
MuhammadKhalilzadeh: Fixed ✅

Client/src/Components/Inputs/Autocomplete/index.jsx
MuhammadKhalilzadeh: Removed ✅

Client/src/Components/Inputs/Autocomplete/index.jsx
MuhammadKhalilzadeh: Fixed ✅

gorkem-bwl:
> @gorkem-bwl
> 
> 'Type to search' text for placeholder, Done ✅
> 
> Now, about the scroller, similar to what we saw on the monitors page, this scroller appears only because the dropdown is containing the whole list, is it appears by default anyway! but when you start typing, the options get more and more limited and the scroller disappears. Now, what we discussed so far as our only solution with @danielcj2 to resolve this, is to redesign the scroller, so that the scroller fits our UI anywhere else on our app.
> 

Sounds good! Thank you.

Client/src/Components/Inputs/Autocomplete/index.css
ajhollid: Have you discussed the issue with the rest of the FE team?  They would be the best people to ask this question to :+1:   I'm sure there's a way to deal with it.

If the rest of the FE team agrees that there's no way to do this other than using the CSS above then we'll have to live with it.

Client/src/Components/Inputs/Autocomplete/index.jsx
```diff
@@ -0,0 +1,105 @@
...
+ *
+ * @component
+ * @param {Object} props - The component props.
+ * @param {string} props.id - The ID of the autocomplete field.
+ * @param {string} props.type - The type of the input field (text or number).
+ * @param {Array} props.options - The options for the autocomplete field.
+ * @param {string} props.placeholder - The placeholder text for the input field.
+ * @param {boolean} props.disabled - Indicates if the field is disabled.
+ * @returns {JSX.Element} The AutoCompleteField component.
+ */
```
ajhollid, line 18: What sort of data does this component take?  Please include an example in the JS docs like this one from the data table that describes the data the component takes:

```
 * @example
 * const data = {
 *   cols: [
 *     { id: 1, name: "First Col" },
 *     { id: 2, name: "Second Col" },
 *     { id: 3, name: "Third Col" },
 *     { id: 4, name: "Fourth Col" },
 *   ],
 *   rows: [
 *     {
 *       id: 1,
 *       data: [
 *         { id: 1, data: <div>Data for Row 1 Col 1</div> },
 *         { id: 2, data: <div>Data for Row 1 Col 2</div> },
 *         { id: 3, data: <div>Data for Row 1 Col 3</div> },
 *         { id: 4, data: <div>Data for Row 1 Col 4</div> },
 *       ],
 *     },
 *     {
 *       id: 2,
 *       data: [
 *         { id: 5, data: <div>Data for Row 2 Col 1</div> },
 *         { id: 6, data: <div>Data for Row 2 Col 2</div> },
 *         { id: 7, data: <div>Data for Row 2 Col 3</div> },
 *         { id: 8, data: <div>Data for Row 2 Col 4</div> },
 *       ],
 *     },
 *   ],
 * };
 *
 * <BasicTable data={data} rows={rows} paginated={true} />
 */
 ```
 
 Otherwise devs have to spend time reading thorugh the component to try to figure out how to use it.

MuhammadKhalilzadeh: @danielcj2 
@Skorpios604 
What do you guys think?

MuhammadKhalilzadeh: Bugs are fixed and modifications are made ✅

ajhollid (approved):
Looks good to me :rocket:

PR #778, opened 2024-09-01
Change color of percentages according to uptime of monitor.

### LINKED ISSUE #775: Monitor percentage colors on the monitor dashboard.
**ISSUE**
Monitor percentage color is tied to whether the monitor is up or down. So when a monitor is down it is displayed as red even though the monitor might have been up 99% of the time. This is misleading because the user might understand it as being down 99% of the time because of the color. 

**SOLUTION**
Having the color not related to whether the monitor is up or down but rather having the color tied to the productivity of the monitor. 
70-100: Green 
40-70: Yellow
0-40: Red

updown2

updown

Skorpios604:
- [x] Adjusted light color in dark theme. 
- [x] Adjusted bg color for the alert to avoid conflicts. 
- [x] Adjusted color of percentages according to uptime. 
- [x] Resolves #775


Client/src/Utils/Theme/lightTheme.js
```diff
@@ -49,8 +49,8 @@ const lightTheme = createTheme({
     },
     warning: {
       text: "#DC6803",
-      main: "#fdb022",
-      light: "#fffcf5",
+      main: "#ff7a00",
+      light: "#ffb400",
```
ajhollid, line 53: Should we define new colors for the uptime percentages instead of changing the warning and success colors?

Is the intentio here to change the colors everywhere, or just for the percentages?

@Skorpios604  @danielcj2  I'll defer to you guys for this one so if you are both in agreement here I'm good to go.

Client/src/Pages/Monitors/Home/monitorData.jsx
```diff
@@ -32,21 +32,30 @@ export const buildData = (monitors, isAdmin, navigate) => {
...
+          if (monitor.uptimePercentage < 0.25) {
+            percentageColor = theme.palette.error.text;
+          } else if (monitor.uptimePercentage >= 0.25 && monitor.uptimePercentage < 0.5) {
+            percentageColor = theme.palette.warning.main;
+          } else if (monitor.uptimePercentage >= 0.5 && monitor.uptimePercentage < 0.75) {
+            percentageColor = theme.palette.warning.light;
+          } else if (monitor.uptimePercentage >= 0.75 && monitor.uptimePercentage <= 1) {
+            percentageColor = theme.palette.success.main;
+          }
+        }
```
ajhollid, line 52: What about using ternary operators instead of an if/else ladder?

```
let percentageColor = 
  monitor.uptimePercentage < 0.25 ? theme.palette.error.text :
  monitor.uptimePercentage < 0.5 ? theme.palette.warning.main :
  monitor.uptimePercentage < 0.75 ? theme.palette.warning.light :
  theme.palette.success.main;
  ```
  
seems more readable to me :thinking: 

You could also do something similar with an immediately invoked anonymous function with early returns

```
let percentageColor = (() => {
  if (monitor.uptimePercentage < 0.25) return theme.palette.error.text;
  if (monitor.uptimePercentage < 0.5) return theme.palette.warning.main;
  if (monitor.uptimePercentage < 0.75) return theme.palette.warning.light;
  return theme.palette.success.main;
})();
```

ajhollid (commented):
Looks great!  I think this is a big improvement over just red or green :+1: 

I have a concern over the changing of the warning and success colors for the theme, but I'm not sure if it is warranted or not.  If you and @danielcj2  could chime in on that and clear it up for me that'd be great! 

Aside from that I personally find the if/else ladder a little hard to read, so I proposed a couple of options for the color mapping if either of them strike your fancy!  If you perfer the if/else ladder there's nothing wrong with it, except the last block could just be an `else`, no need for the conditional.

Client/src/Utils/Theme/lightTheme.js
danielcj2: I don't think `main` needs changing tbh 🤔. But if it is needed for improved contrast then it might be better to define separate colors.

danielcj2: Yeah, now that I've had more time to think about it, that's probably the best approach. We're likely going to be using the props in `warning` and `success` for charts in the future, and it'd be better if they remained as they are.

ajhollid: Alright makes sense, @Skorpios604 sound good to you?

Skorpios604: Yup! That sounds good to me. Thanks guys for looking into that for me and for tackling finding the best approach to to go about it.

Client/src/Pages/Monitors/Home/monitorData.jsx
Skorpios604: Yup either of those options work for me. I will go ahead and implement one of them. Thank you invaluable input. I really appreciate it!

ajhollid (approved):
Nice work, looks great!

PR #791, opened 2024-09-04
Added avgResponseTime to monitor stats, changed date grouping for agg…

ajhollid:
This PR adds more data to the monitor stats and aggregate monitor stats
- [x] Refactor getMonitorStatsById to return aggregate data
- [x] Only return data for selected date range


danielcj2 (approved):
Looks good 👍

PR #795, opened 2024-09-04
modify check schema to handle optional TTL.

### LINKED ISSUE #764: BE - Optional TTL for checks
In the settings page a user can optionally set how long they want their data to persist for.

There should be no TTL by default, and if the user adds TTL an expiry should be added to all checks.

If there are chekcs that are already past the TTL they should be deleted.  Not sure if Mongo does this automatically.

Skorpios604:
- [x] Modify check schema to handle optional TTL.
- [ ] Set TTL for existing checks. 
- [ ] Ensure new checks have TTL. 
- [ ] Update API endpoint to save TTL settings. 
- [ ] Add actions and reducers to Redux slice to handle TTL settings. 
- [ ] Ensure MongoDB collection uses TTL indexes to auto delete expired documents. 
- Resolves #764

### DIFF
--- a/Server/service/networkService.js
+++ b/Server/service/networkService.js
@@ -103,13 +103,17 @@ class NetworkService {
       const response = await ping.promise.probe(job.data.url);
       return response;
     };
-
+  
     let isAlive;
-
+  
     try {
-      const { responseTime, response } =
-        await this.measureResponseTime(operation);
+      const { responseTime, response } = await this.measureResponseTime(operation);
       isAlive = response.alive;
+  
+      // Fetch TTL setting from the monitor
+      const monitor = await this.db.getMonitorById(job.data._id);
+      const ttlDays = monitor.ttlDays || 30; // Default to 30 days if not set
+  
       const checkData = {
         monitorId: job.data._id,
         status: isAlive,
@@ -117,6 +121,7 @@ class NetworkService {
         message: isAlive
           ? successMessages.PING_SUCCESS
           : errorMessages.PING_CANNOT_RESOLVE,
+        expiry: new Date(Date.now() + ttlDays * 24 * 60 * 60 * 1000) // Set TTL
       };
       return await this.logAndStoreCheck(checkData, this.db.createCheck);
     } catch (error) {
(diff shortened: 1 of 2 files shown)


Server/service/networkService.js
```diff
@@ -103,20 +103,25 @@ class NetworkService {
...
+  
     try {
-      const { responseTime, response } =
-        await this.measureResponseTime(operation);
+      const { responseTime, response } = await this.measureResponseTime(operation);
       isAlive = response.alive;
+  
+      // Fetch TTL setting from the monitor
+      const monitor = await this.db.getMonitorById(job.data._id);
+      const ttlDays = monitor.ttlDays || 30; // Default to 30 days if not set
```
ajhollid, line 115: How about we name this `checkTTL` since it's TTL for the check, not the monitor, and we don't want to lock ourselves in to using days, so we should drop the days suffix

Server/service/networkService.js
```diff
@@ -103,20 +103,25 @@ class NetworkService {
...
+  
       const checkData = {
         monitorId: job.data._id,
         status: isAlive,
         responseTime,
         message: isAlive
           ? successMessages.PING_SUCCESS
           : errorMessages.PING_CANNOT_RESOLVE,
+        expiry: new Date(Date.now() + ttlDays * 24 * 60 * 60 * 1000) // Set TTL
```
ajhollid, line 124: We only want to set `expiry` if `monitor.checkTTL` is not undefined right?  Otherwise we don't need to set it as it has a default value

Server/service/networkService.js
```diff
@@ -144,42 +151,78 @@ class NetworkService {
...
-        await this.measureResponseTime(operation);
-      // check if response is in the 200 range, if so, service is up
+      // Measure response time and get the response
+      const { responseTime, response } = await this.measureResponseTime(operation);
       isAlive = response.status >= 200 && response.status < 300;
-
-      //Create a check with relevant data
+  
+      // Fetch monitor TTL
+      const monitor = await this.db.getMonitorById(job.data._id);
```
ajhollid, line 163: We don't need to do this query, `job.data` is this `Monitor`.  You can revert this code and just add the TTL parts

Server/service/networkService.js
Skorpios604: Agreed. Thank you!

Server/service/networkService.js
Skorpios604: Thats right. Let me fix that.

Server/service/networkService.js
Skorpios604: Alrighty sounds good. Will do! Thanks!!

PR #802, opened 2024-09-05
get monitors from redux, complete submit func

MuhammadKhalilzadeh:
- [X] Getting monitors data from redux
- [X] Preparing data for sending to backend
- [X] Adding validations for Add monitors


ajhollid (changes requested):
The dark styles need to be updated:

Screenshot from 2024-09-05 09-39-09


and the logic for submitting needs to be refactrored:

```
    logger.log("error: ", error);
    if (!error || error.details.length === 0) {
      setErrors({});
    } else {
      const newErrors = {};
      error.details.forEach((err) => {
        newErrors[err.path[0]] = err.message;
      });
      setErrors(newErrors);
      createToast({
        body:
          error.details && error.details.length > 0
            ? error.details[0].message
            : "Error validating data",
      });
      logger.error("Validation errors:", error.details);
    }

    logger.log("Submitting data: ", data);
   ```
    
Errors should only be logged if there are errors and data should only be submitted if there are no errors.

I assume it is set up this way for testing purposes, but it needs to be refactored in order to prepare for actual data submission to the BE.

I believe all font sizes should be declared in the theme? 

```
          sx={{
            width: "fit-content",
            fontSize: "var(--env-var-font-size-small)",
            borderBottom: `1px dashed ${theme.palette.primary.main}`,
            paddingBottom: "4px",
          }}
```

We are eventually going to remove most if not all of the global CSS variables.

**Comment by ajhollid** [comment]
Apologies I also missed this in pervious review:

```
import dayjs from "dayjs";
```
I don't think there's any need for the dayjs module, again we should minimize the llibraries that we use unless there is a specific reuqirement.  Using the built in `Date` class is sufficient and is consistent with the rest of the app

**Comment by MuhammadKhalilzadeh** [comment]
@ajhollid 
I removed `import dayjs from "dayjs";` and this is what I got:

Screenshot 2024-09-09 093832
Screenshot 2024-09-09 094611

So this:
```
  const [values, setValues] = useState({
    repeat: 1,
    date: dayjs(),
    startTime: dayjs(),
    duration: "60",
    unit: "minutes",
    displayName: "",
    AddMonitors: "",
  });
```
Turned into this:

```
  const [values, setValues] = useState({
    repeat: 1,
    date: new Date(),
    startTime: new Date(),
    duration: "60",
    unit: "minutes",
    displayName: "",
    AddMonitors: "",
  });
```

and after searching for the cause of error, here's what I got:



The error `Uncaught TypeError: value.isValid is not a function` indicates that the value being passed to the `AdapterDayjs` is not a valid `Dayjs` object. This typically happens when a plain JavaScript **Date** object is used instead of a `Dayjs` object.

To resolve this, you need to convert the **Date** object to a `Dayjs` object before passing it to the `AdapterDayjs`.

Now, these are links for [MUI's Date Picker](https://mui.com/x/react-date-pickers/date-picker/) and [MUI's Time Picker](https://mui.com/x/react-date-pickers/time-picker/). 
What's do you suggest?

**Comment by MuhammadKhalilzadeh** [comment]
@ajhollid @danielcj2 
Those values were not in the theme yet, the time I created this pull request. Do you want me to add them in this same branch? shouldn't we add them later?

**Comment by ajhollid** [comment]
> @ajhollid 
> I removed `import dayjs from "dayjs";` and this is what I got:
> 
> Screenshot 2024-09-09 093832
> Screenshot 2024-09-09 094611
> 
> So this:
> ```
>   const [values, setValues] = useState({
>     repeat: 1,
>     date: dayjs(),
>     startTime: dayjs(),
>     duration: "60",
>     unit: "minutes",
>     displayName: "",
>     AddMonitors: "",
>   });
> ```
> Turned into this:
> 
> ```
>   const [values, setValues] = useState({
>     repeat: 1,
>     date: new Date(),
>     startTime: new Date(),
>     duration: "60",
>     unit: "minutes",
>     displayName: "",
>     AddMonitors: "",
>   });
> ```
> 
> and after searching for the cause of error, here's what I got:
> 
> 
> 
> The error `Uncaught TypeError: value.isValid is not a function` indicates that the value being passed to the `AdapterDayjs` is not a valid `Dayjs` object. This typically happens when a plain JavaScript **Date** object is used instead of a `Dayjs` object.
> 
> To resolve this, you need to convert the **Date** object to a `Dayjs` object before passing it to the `AdapterDayjs`.
> 
> Now, these are links for [MUI's Date Picker](https://mui.com/x/react-date-pickers/date-picker/) and [MUI's Time Picker](https://mui.com/x/react-date-pickers/time-picker/). 
> What's do you suggest?

Hmm, I read through the docs and we have a number of choices

1.  Use a different date picker
2.  Write our own date adapter that works with this library
3.  Use dayjs and convert the date back to the same format we use elsewhere after it is picked.

I think 3 is probably the least problematic, so we'll have to bite the bullet and add the library. 

Make sure to convert the time produced by the picker to our standard format produced by Date, we use UTC time

**Comment by ajhollid** [comment]
> @ajhollid @danielcj2 
> Those values were not in the theme yet, the time I created this pull request. Do you want me to add them in this same branch? shouldn't we add them later?

If they are in the theme now please use them, if not then you can coordinate with @danielcj2 to add them.

This should be done in this PR to avoid building technical debt.

File: Client/src/Pages/Maintenance/CreateMaintenanceWindow/index.jsx [hunk 1]
```diff
@@ -200,20 +213,23 @@ const CreateNewMaintenanceWindow = () => {
...
-            onChange={(e) => handleChange(e, "AddMonitors")}
+          <AutoCompleteField
+            sx={{ marginLeft: "auto" }}
+            options={monitorOptions}
+            width={380}
+            autoCompleteValue={autoCompleteValue}
+            setAutoCompleteValue={setAutoCompleteValue}
+            autoCompleteInputValue={autoCompleteInputValue}
+            setAutoCompleteInputValue={setAutoCompleteInputValue}
             error={errors.addMonitors}
```
ajhollid, line 227: Is it possible to restrict the input to options that exist?  Right now I can enter whatever I want as long as it begins with an option in the list:

image

And this will be accepted as acceptable input.  It would be best if this could be restricted.

Also if I enter something that is not acceptable it is not rejected, an error is thrown and is unhandled:

image

This needs to be addressed as well and will probably be resolved by fixing the above issue

Client/src/Pages/Maintenance/CreateMaintenanceWindow/index.jsx
```diff
@@ -57,6 +62,16 @@ const durationOptions = [
...
   const theme = useTheme();
+  const monitorState = useSelector((state) => state.uptimeMonitors);
+  const [monitorOptions, setMonitorOptions] = useState([]);
+
+  const [autoCompleteValue, setAutoCompleteValue] = useState("");
+  const [autoCompleteInputValue, setAutoCompleteInputValue] = useState("");
+
+  useEffect(() => {
+    setMonitorOptions(monitorState.monitors);
+    console.log("monitorState.monitors -->", monitorState.monitors);
```
ajhollid, line 73: This can be removed before merging :+1:

Client/src/Pages/Maintenance/CreateMaintenanceWindow/index.jsx
```diff
@@ -80,36 +95,34 @@ const CreateNewMaintenanceWindow = () => {
   const handleSubmit = async () => {
     const data = {
       repeat: values.repeat,
-      date: values.date.format("YYYY-MM-DD"),
-      startTime: values.startTime.format("HH:mm"),
+      date: values.date.utc().format("YYYY-MM-DD"),
+      startTime: values.startTime.utc().format("HH:mm"),
```
ajhollid, line 99: We use typical unix timestamps measured in ms throughout the app, both on the frontend and the backend, so the date needs to be in that format.  Conversion to human readable date/time will be done in the components that use the timestamp.

We don't need to submit the date and start times seperately, just the timestamp.

Duration should also be converted to ms

TLDR we need a timestamp, you [can get it from DayJS](https://day.js.org/docs/en/display/unix-timestamp-milliseconds)

ajhollid (commented):
Looks good!  Just some adjustments needed for dates and there are a couple validation issues with autocomplete that need to be resovled.

I think we're ready to go after that.

Client/src/Pages/Maintenance/CreateMaintenanceWindow/index.jsx
MuhammadKhalilzadeh: Fixed ✅

Client/src/Pages/Maintenance/CreateMaintenanceWindow/index.jsx
MuhammadKhalilzadeh: almost fixed ✅❗
There are some parts that could not be deleted, since removing them makes other parts like validations broken

Client/src/Pages/Maintenance/CreateMaintenanceWindow/index.jsx
MuhammadKhalilzadeh: I somehow fixed it.
The text field is disabled now!
the user might be able to type, but only an option from the list is selected.
This is what could be done by attributes that was provided with components.

danielcj2 (approved):

Client/src/Pages/Maintenance/CreateMaintenanceWindow/index.jsx
```diff
@@ -66,50 +77,80 @@ const CreateNewMaintenanceWindow = () => {
...
+  };
+
+  const convertDurationToMilliseconds = (duration, unit) => {
+    let milliseconds = 0;
+    if (unit === "minutes") {
+      milliseconds = duration * 60 * 1000;
+    } else if (unit === "hours") {
+      milliseconds = duration * 60 * 60 * 1000;
+    } else if (unit === "days") {
+      milliseconds = duration * 24 * 60 * 60 * 1000;
```
ajhollid, line 109: We should avoid [magic numbers](https://en.wikipedia.org/wiki/Magic_number_(programming)) in our code.

These can be refactored as
```
MS_PER_MINUTE = 60000
MS_PER_HOUR = 360000
MS_PER_DAY = 86400000
```
for clarity

ajhollid (commented):
image

There is a warning on page load, it looks like a valid value is not provided to the autocomplete for a default value perhaps?

There are other errors relating to missing prop validation as well:

image

Please make sure to add those prop types.

`date` and `startTime` do not need to be timestamps, you can leave those as human readable dates as they were before

image

We only need the timestamp of the final value of when the maintenance window should take place.

```
    if (unit === "minutes") {
      milliseconds = duration * 60000;
    } else if (unit === "hours") {
      milliseconds = duration * 360000;
    } else if (unit === "days") {
      milliseconds = duration * 86400000;
    }
    ```
    
  These are still magic numbers, they should be defined as constants:
  
  ```
  MS_PER_MINUTE = 60000;
  MS_PER_HOUR = 360000;
  MS_PER_DAY = 86400000;
```

and referenced like so:

```
    if (unit === "minutes") {
      milliseconds = duration * MS_PER_MINUTE;
    } 
 ```

**Comment by MuhammadKhalilzadeh** [comment]
Fixes:

- Warning Fixed
- date and time turned back to human readable formats
- constants for timeframes by ms added

Screenshot 2024-09-11 210748
Screenshot 2024-09-11 205518

**Comment by ajhollid** [comment]
Validation doesn't work on page load before adding anything:

image

It appears a variable is being unsafely accessed.

Thorough testing and debugging should be done prior to making pull requests, everyone's time is limited and back and forth in a github review takes a long time.  Other team members are here to verify the code and look for anything non-obvious that may have been missed, the general debugging and preparation for merging should be done on your end :+1:

**Comment by MuhammadKhalilzadeh** [comment]
@ajhollid 
In order to fix the warning that appeared on the console, I passed a null value some where. Apparently the error appeared for you but not me, but I'm not sure why because I make sure that there's absolutely no errors or warning before pushing it! 🤔

I made some changes and tried to fix the seemingly issue and prevent any potential ones. Now, There's ABSOLUTLY NOOO ERRORS on my end!

Here's what I have when page first loads:
Screenshot (131)

Here's what happens when I click Create button and validations errors appears:
Screenshot (132)

And here's what the data looks like when button is pressed:
Screenshot (133)

**Comment by ajhollid** [comment]
image

There are still errors present in this code:

```
    const combinedTimestamp = dayjs(
      `${values.date.format("YYYY-MM-DD")}T${values.startTime.format("HH:mm")}`
    ).valueOf();
```

This is not a safe way to access properties of the `values` object.  If `value.date` is `null` as in the screenshot above you will throw an error because you can't read `null.format`:

```
Cannot read properties of null (reading 'format')
```

It is dangerous to assume that an object is guaranteed to have a property.

```
const example = {
  propertyOne: {
    propertyTwo: {
      propertyThree: "Hello World",
    },
  },
};
```

Imagine you have this object, if you call `console.log(propertyOne.propertyTwo.propertyThree)` you will log "Hello World", but what if the object now looks like this:

```
const example = {
  propertyOne: {
    propertyTwo: undefined,
  },
};
```

if you call `console.log(propertyOne.propertyTwo.propertyThree)` you will throw an error as `propertyTwo` is `undefiend`.

Safe property access is a general principle of development and is _especially_ inportant when dealing with user input as we cannot really guarantee what we are going to get back from the user.

In order to write professional quality robust code it is important to make sure that your code works in all cases, not just the [happy path](https://en.wikipedia.org/wiki/Happy_path) of ideal use case.

Modern Javascript offers a lot of clean and simple solutions for dealing with safe property access, I personally like optional chaining.  Check out [this article](https://dev.to/mmainulhasan/optional-chaining-and-nullish-coalescing-safe-property-access-in-javascript-1d3d) for some ideas on how to deal with the problem.

If you take a look through the app you'll see lots of examples of optional chaining going on if you need a local example to look at.

That date field's error is also not cleared if there is an error and then it is later resolved:

image

Even with a valid date the error state persists

PR #808, opened 2024-09-05
Fixed janky bar animations

danielcj2:
- [x] Memoized chart data
- [x] Fixed bar transitions


ajhollid (approved):
Nice work here, looks great

PR #834, opened 2024-09-13
Pagespeed Details Revamp

danielcj2:
- [x] Refactored code to separate files for increased readability
- [x] Revamped UI
- [x] Removed redundant css
- [x] Removed old chart files
- [x] Implemented new stacked area chart
- [x] Added toggle dataKey checkboxes


Client/src/Pages/PageSpeed/Details/Charts/AreaChart.jsx
```diff
@@ -0,0 +1,313 @@
...
+
+  // Helper function to add a null entry
+  const addNullEntry = (timestamp) => {
+    formattedData.push({
+      accessibility: "N/A",
+      bestPractices: "N/A",
+      performance: "N/A",
+      seo: "N/A",
+      createdAt: timestamp,
+    });
```
ajhollid, line 144: What was the reason for interpolating missing data points?  Was the appearance of the graph being distorted?

It seems like users might not realize that these data points are actually missing rather than there just being no data for them if we display N/A.  Maybe it would be better not to interpolate the points, or if it is critical for appearance then maybe some indication that the check doesn't actually exist?

ajhollid (approved):
Nice job, the new graph looks pretty awesome, love the ability to turn off certain metrics :+1: 

The only thing that felt a bit weird to me was the interpolation of missing data points for the Pagespeed Area chart, some details in coments

Client/src/Pages/PageSpeed/Details/Charts/AreaChart.jsx
danielcj2: The reason for that is because the x-axis ticks did not appear to be continuous. Since the monitor will run checks while the server is active, once you shut it down and then turn it back on, it'll start displaying checks where it initially left off, so you'll have cases where the previous check might've happened let's say 8 hours ago, which didn't feel right to me. So I added null data points for a more accurate representation of time. I don't think removing the interpolation all together is ideal but what we can do is remove the `connectNulls` prop which will not connect the dots anymore, leaving a visible gap in the chart. This should make it more intuitive for users but will hurt the overall aesthetics😭 

🤔 I also think we should probably add this kind of interpolation to our other charts.

A paused monitor resumed after several days:
image

image

Issue #674, opened 2024-08-19
FE - Maintenance Window Functionality


ajhollid:
Maintanance window UI is mostly complete.  The next step should be functionality.

1.  Marshal data into the form that it should look like for submission and log it when the submit button is pressed

```
{
   startDate: <date>
   endDate: <end_date>
   ...
}
```

2. Send data to backend via Axios request

PR #861, opened 2024-09-24
feat: Maintenance Page, resovles #674

danielcj2:
- [x] Cleaned up code
- [x] Added layout
- [x] Add remaining fields
  - [x] Connect to local state
  - [x] Add error handling 
- [x] Connect to BE 
- [x] Handle action menu
  - [x] Delete
  - [x] Edit
  - [x] Pause


jennifer-gan:
I have to say overall looking good, just make sure the failed checks turn to successful regarding adding rate limit as in above details, it is good to go

ajhollid:
> I have to say overall looking good, just make sure the failed checks turn to successful regarding adding rate limit as in above details, it is good to go

Thanks for having a look at it :+1: 

Rate limiting isn't actually a major concern for us as we are not building a SaaS product and this product will be used generlly in a local environment by small teams or individuals.

Rate limiting is something we can look at in the future though if it ever becomes an issue for users, though it seems unlikely we will need to.

Issue #912, opened 2024-10-08
Add Prettierrc config


marcelluscaio:
The lack of a prettierrc file makes github diffs less precise.

Creating a prettierrc file would solve that.

PR #916, opened 2024-10-08
chore: adding prettierrc file

marcelluscaio:
Resolves Issue #912

- Created a prettierrc file
- Installed prettier on Client and Server
- Tested two folders to make sure it was working

Now all formatting will be done following project settings, instead of local settings.

### DIFF
--- a/Client/package.json
+++ b/Client/package.json
@@ -25,6 +25,7 @@
     "dayjs": "1.11.11",
     "joi": "17.13.1",
     "jwt-decode": "^4.0.0",
+    "prettier": "^3.3.3",
     "react": "^18.2.0",
     "react-dom": "^18.2.0",
     "react-redux": "9.1.2",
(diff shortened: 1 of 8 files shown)


Client/package.json
```diff
@@ -25,6 +25,7 @@
     "dayjs": "1.11.11",
     "joi": "17.13.1",
     "jwt-decode": "^4.0.0",
+    "prettier": "^3.3.3",
```
ajhollid, line 28: Prettier should probably be a dev dependency instead of a build dependency, we don't need it to be included in the build I don't think?

ajhollid (changes requested):
Thanks for putting the time prettier config together, looks pretty good to me!  We can adjust settings as we go if people gave particularly preference as to how to set up the config.

I don't think we need prettier in the build though so it should probably be a dev dependency. Let me know if I've missed something though!

Client/package.json
marcelluscaio: Agreed, Alex! I corrected it. Thanks for that!

ajhollid (approved):
Looks good to me, ready to merge in!

Issue #907, opened 2024-10-04
Fix the "current password" widget


gorkem-bwl:
The "current password" section needs to be aligned.

<img width="1050" alt="Screenshot 2024-10-04 at 12 14 55 AM" src="https://github.com/user-attachments/assets/aa4c9ae8-88fb-424e-bcb7-108d115a7778">

GaurangMapuskar:
@gorkem-bwl please assign me this issue. I will work on this immediately

gorkem-bwl:
Sure - go ahead :)

rishavd3v:
@gorkem-bwl is it still open? I would like to work on it.

om-3004:
Hello @gorkem-bwl, Please find and review my PR and let me know if there are any issues.

om-3004:
Hello @gorkem-bwl, I had some problem with the previous forked repo so I have forked a new one and created a new PR. Please find and review my PR and let me know if there are any issues.

PR #940, opened 2024-10-11
fix: add css to fix the current password widget under account>password

om-3004:
## Fixes https://github.com/bluewave-labs/bluewave-uptime/issues/907 (Fix the "current password" widget)

- Added the CSS to fix current password widget in light/dark mode.
Screenshot 2024-10-12 002318
Screenshot 2024-10-12 002329

### DIFF
--- a/Client/src/Components/Inputs/Field/index.jsx
+++ b/Client/src/Components/Inputs/Field/index.jsx
@@ -74,6 +74,7 @@ const Field = forwardRef(
           "&:has(.input-error) .MuiOutlinedInput-root fieldset": {
             borderColor: theme.palette.error.text,
           },
+          display: hidden ? "none" : "",
         }}
       >
         {label && (
@@ -142,7 +143,7 @@ const Field = forwardRef(
                   },
                 }
               : {
-                  display: hidden ? "none" : "",
+                  
                 }
           }
           InputProps={{


gorkem-bwl:
Looks good, thanks!

Client/src/Components/Sidebar/index.css
```diff
@@ -92,3 +92,7 @@ aside .MuiListSubheader-root {
     visibility: visible;
   }
 }
+
+.MuiTabPanel-root.css-8i83hh-MuiTabPanel-root .MuiStack-root.field.field-text.css-qm2p9n-MuiStack-root, .MuiTabPanel-root.css-2aha5r-MuiTabPanel-root .MuiStack-root.field.field-text.css-78xnin-MuiStack-root {
+  display: none;
+}
```
ajhollid, line 98: This is in the sidebar CSS file but it is targeting a field in the password panel component.  This is unmaintainable as future devs will not know this CSS is in this file, nor will they know what it is for if they come across this CSS in the future.

The specificity of this CSS selector is also not really maintainable as it is not reasonably possible to tell what that selector is for.  I assume it was extracted from dev tools inspector?  It works, but no one is going to know what it is for or what it means :thinking:

ajhollid (changes requested):
I understand what you are going for here and it works, however it isn't really maintainable.

I propose a different solution to this issue.

The root of the problem is that the `Field` component is meant to be hidden:

```
  <Field
    type="text"
    id="hidden-username"
    name="username"
    autoComplete="username"
    hidden={true}
    value=""
  />
```
      
however this is not behaving as expected.  The problem lies with the `Field` component and how the hidden prop is applied:
      
```
sx={
      type === "url"
        ? {
            "& .MuiInputBase-root": { padding: 0 },
            "& .MuiStack-root": {
              borderTopLeftRadius: theme.shape.borderRadius,
              borderBottomLeftRadius: theme.shape.borderRadius,
            },
          }
        : {
            display: hidden ? "none" : "",
          }
    }
```
          
The hidden prop is applied to the  `TextField` subcomponent of `Field`, while in fact it should be applied to the root `Stack` subcomponent of `Field`.

I suggest reverting `TextField`'s SX prop to

```
  sx={
    type === "url"
      ? {
          "& .MuiInputBase-root": { padding: 0 },
          "& .MuiStack-root": {
            borderTopLeftRadius: theme.shape.borderRadius,
            borderBottomLeftRadius: theme.shape.borderRadius,
          },
        }
      : {
        }
  } 
```
 and adding
 
 `display: hidden? "none" : ""`
 
 to the root `Stack` subcomponent.

om-3004:
@ajhollid Yes understood, you are right. I've committed that change so please review it.

Client/src/Components/Inputs/Field/index.jsx
```diff
@@ -74,6 +74,7 @@ const Field = forwardRef(
           "&:has(.input-error) .MuiOutlinedInput-root fieldset": {
             borderColor: theme.palette.error.text,
           },
+          display: hidden ? "none" : "",
```
ajhollid, line 77: Awesome, thanks for taking care of this :+1:

ajhollid (approved):
Nice, short and sweet!  This resovles the issue, thank you for your contribution

PR #946, opened 2024-10-12
check if the url resolves before adding the monitor

### LINKED ISSUE #513: Check if endpoint URL resolves before adding the monitor
When adding a monitor, the endpoint must be checked to move forward with adding.

If the endpoint URL doesn't resolve, it should throw a notification saying that the endpoint doesn't resolve. During the endpoint checking process, the "Create new monitor" button should read "Checking the endpoint". 

There are two options: 

1. If the endpoint doesn't resolve, it should throw a toast notification saying "The endpoint you entered doesn't resolve. Check the URL again.". The same page stays. 
2. If the endpoint resolves, then the monitor is added. There are no changes here.

om-3004:
## Fixes #513 (Check if endpoint URL resolves before adding the monitor)

- Made the button content dynamic

1. When Website Monitoring selected:
Screenshot 2024-10-13 000650

2. When Ping Monitoring selected:
Screenshot 2024-10-13 000702

- When Website Monitoring is selected, check if the entered url is resolved before adding the monitor
Screenshot 2024-10-13 000720

### DIFF
--- a/Client/src/Pages/Monitors/CreateMonitor/index.jsx
+++ b/Client/src/Pages/Monitors/CreateMonitor/index.jsx
@@ -124,46 +124,58 @@ const CreateMonitor = () => {
 
   const handleCreateMonitor = async (event) => {
     event.preventDefault();
-    //obj to submit
-    let form = {
-      url:
-        //preprending protocol for url
-        monitor.type === "http"
-          ? `http${https ? "s" : ""}://` + monitor.url
-          : monitor.url,
-      name: monitor.name === "" ? monitor.url : monitor.name,
-      type: monitor.type,
-      interval: monitor.interval * MS_PER_MINUTE,
-    };
 
-    const { error } = monitorValidation.validate(form, {
-      abortEarly: false,
-    });
+    const formattedUrl = monitor.type === "http"
+    ? `http${https ? "s" : ""}://` + monitor.url
+    : monitor.url;
 
-    if (error) {
-      const newErrors = {};
-      error.details.forEach((err) => {
-        newErrors[err.path[0]] = err.message;
-      });
-      setErrors(newErrors);
-      createToast({ body: "Error validation data." });
-    } else {
-      form = {
-        ...form,
-        description: form.name,
-        teamId: user.teamId,
-        userId: user._id,
-        notifications: monitor.notifications,
-      };
-      const action = await dispatch(
-        createUptimeMonitor({ authToken, monitor: form })
-      );
-      if (action.meta.requestStatus === "fulfilled") {
-        createToast({ body: "Monitor created successfully!" });
-        navigate("/monitors");
-      } else {
-        createToast({ body: "Failed to create monitor." });
+    try {
+      // Check if the URL resolves by sending a request
+      const response = await fetch(formattedUrl, { method: 'GET', mode: 'no-cors' });
+      if (!response) {
+        throw new Error();
       }
+
+      //obj to submit
+      let form = {
+        url:formattedUrl,
+        name: monitor.name === "" ? monitor.url : monitor.name,
+        type: monitor.type,
+        interval: monitor.interval * MS_PER_MINUTE,
+      };
+
+      const { error } = monitorValidation.validate(form, {
+        abortEarly: false,
+      });
+
+      if (error) {
+        const newErrors = {};
+        error.details.forEach((err) => {
+          newErrors[err.path[0]] = err.message;
+        });
+        setErrors(newErrors);
+        createToast({ body: "Error validation data." });
+      } else {
+        form = {
+          ...form,
+          description: form.name,
+          teamId: user.teamId,
+          userId: user._id,
+          notifications: monitor.notifications,
+        };
+        const action = await dispatch(
+          createUptimeMonitor({ authToken, monitor: form })
+        );
+        if (action.meta.requestStatus === "fulfilled") {
+          createToast({ body: "Monitor created successfully!" });
+          navigate("/monitors");
+        } else {
+          createToast({ body: "Failed to create monitor." });
+        }
+      }
+    } catch (error) {
+      createToast({ body: "The endpoint you entered doesn't resolve. Check the URL again."  });
+      setErrors({ url: "The entered URL is not reachable." });
     }
   };
 
@@ -375,7 +387,7 @@ const CreateMonitor = () => {
             onClick={handleCreateMonitor}
             disabled={Object.keys(errors).length !== 0 && true}
           >
-            Create monitor
+            {monitor.type === "http" ? "Create Monitor" : "Check Endpoint"}
           </Button>
         </Stack>
       </Stack>


Client/src/Pages/Monitors/CreateMonitor/index.jsx
```diff
@@ -124,46 +124,58 @@ const CreateMonitor = () => {
...
-    } else {
-      form = {
-        ...form,
-        description: form.name,
-        teamId: user.teamId,
-        userId: user._id,
-        notifications: monitor.notifications,
+    try {
+      // Check if the URL resolves by sending a request
+      const response = await fetch(formattedUrl, { method: 'GET', mode: 'no-cors' });
```
ajhollid, line 134: Checking if the URL resolves should be done on the back end, sending a `no-cors` request from the front end won't tell us if the server we want to reach is reachable or not.

The response is opaque, so we have no idea what is in the response, just that we got something back.  If there was a proxy inbetween us and the server for example we could still get a 200 response, but the server might not be reachable.

We should implement a more comprehensive solution which would look like this:

Front End sends GET request to Back End -> BE forwards request to Server URL -> BE returns response to FE

ajhollid (commented):
I appreciate the effort put in here, however I believe a more comprehensive approach here is to proxy this URL to the back end so that we can get a non-opaque response back from the server we are trying to connect to.

If you would like to tackle this one it's a bit more invovled than the naive solution of sending a no-cors request from the front end.  If you feel like jumping on it let us know, otherwise we can close this PR and revisit the issue when we have bandwidth to do so.

om-3004:
Hello @ajhollid, I have updated the code to follow the flow 
Front End sends GET request to Back End -> BE forwards request to Server URL -> BE returns response to FE

- On the back-end, I have used dns  to check whether the domain resolves to an IP address. 

https://github.com/user-attachments/assets/9f8ed3f6-3109-40c3-8be5-9c1234081ce9

om-3004:
Also I want to know what should happen when Ping Monitoring is selected?

Client/src/Features/UptimeMonitors/uptimeMonitorsSlice.js
```diff
@@ -31,6 +31,30 @@ export const createUptimeMonitor = createAsyncThunk(
   }
 );
 
+export const checkEndpointResolution = createAsyncThunk(
+  "monitors/checkEnpoint",
```
ajhollid, line 35: Typo here, should be `checkEndpoint`

Client/src/Pages/Monitors/CreateMonitor/index.jsx
```diff
@@ -375,12 +385,12 @@ const CreateMonitor = () => {
             onClick={handleCreateMonitor}
             disabled={Object.keys(errors).length !== 0 && true}
           >
-            Create monitor
+            Create Monitor
```
ajhollid, line 388: I believe this is intentionally lower case, this can be reverted

ajhollid (changes requested):
Looks good for the most part! 

Minor nit-pick about the route path, I think simpler would be nicer.

Major revisions required here are

1.  Testing:  Currently monitors have 100% coverage, please include comprehensive tests for your new method in `monitorController`.  Tests are located in `./Server/test/controller`

2.  Documentaiton:  Please add full documentation for the new endpoint created to the `openapi.json` file.  Our documentation follows the [OpenAPI](https://www.openapis.org/) standards.

3.  Resolution check also needs to be done for PageSpeed monitors on its respsective create monitor page

ajhollid:
> Also I want to know what should happen when Ping Monitoring is selected?

I think you can bypass the resolution check if Ping is seleted

Server/controllers/monitorController.js
```diff
@@ -262,6 +263,43 @@ const createMonitor = async (req, res, next) => {
...
+    // Remove the trailing slash if it exists
+    monitorURL = monitorURL.endsWith('/') ? monitorURL.slice(0, -1) : monitorURL;
+
+    dns.setServers(['8.8.8.8', '1.1.1.1']);// Google, Cloudfare    
+    dns.resolve(monitorURL, (error) => {
+      console.log("Inside .resolve");
+      if (error) {
+        next(handleError(error, SERVICE_NAME, "checkEndpointResolution"));
+        return ;
+      }
```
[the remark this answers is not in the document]
om-3004: By throwing the error, the server is getting crashed because 'The error thrown inside the callback function is not caught by the surrounding try-catch block because it happens asynchronously'.

Instead can we return the response directly?

if (error) {
        return res.status(400).json({
          success: false,
          msg: \`DNS resolution failed: ${error.message}\`,
        });
      }

om-3004:
@ajhollid I have made a commit that adds the test cases for monitor endpoint resolution. Can you please review it.

Server/tests/controllers/monitorController.test.js
```diff
@@ -473,6 +475,58 @@ describe("Monitor Controller - createMonitor", () => {
   });
 });
 
+describe('checkEndpointResolution', () => {
```
ajhollid, line 478: Please update to follow the naming conventions of other tests 

```
"Monitor Controllor - checkEndpointResolution"

File: Server/controllers/monitorController.js (hunk 4 above)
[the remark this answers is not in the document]
Comment (ajhollid) [reply Server/controllers/monitorController.js:292]: All error handling is done centrally via middleware, we do not want to have error handling done in multiple places.

You can promisify `dns.resolve` to throw the error:

```
await new Promise((resolve, reject) => {
	dns.resolve(monitorURL, (error) => {
		if (error) {
			reject(error);
		}
		resolve();
	});
});
return res.status(200).json({
	success: true,
	msg: `URL resolved successfully`,
});
```

File: Server/tests/controllers/monitorController.test.js [hunk 6]
```diff
@@ -473,6 +475,58 @@ describe("Monitor Controller - createMonitor", () => {
...
+      msg: 'URL resolved successfully',
+    })).to.be.true;
+    expect(next.called).to.be.false;
+  });
+  it("should call next with an error if an exception is thrown", async () => {
+    req.query = {};  // Missing monitorURL will cause an error
+    await checkEndpointResolution(req, res, next);
+    expect(next.calledOnce).to.be.true;
+    const error = next.getCall(0).args[0];
+    expect(error).to.be.an.instanceOf(Error);
```
ajhollid, line 526: It would be best to check for the specific error that you are getting rather than just to see if an error is thrown.

We should be very specific, if you had a function that did:

```
dangerousOperationOne()
dangerousOperationTwo()
```

and your test just checks to see if an error is thrown you wouldn't know if it was `dangerousOpeartionOne()` or `dangerousOperationTwo()` that failed

ajhollid (changes requested):
The tests look good for the most part, though some specificity around checking for errors would be good.

You can promisify `dns.resolve` to throw the error and let the error handling midleware handle it.  This ensures consistency in the handling of errors and keeps the code base maintainable.  If errors are being handled as they pop up all over the place it will be difficult to maintain.

There are some issues from the previous code review to resolve still as well.

Please note that the backend has recently migrated to use ESM, so you will have to resolve merge conflicts before we can merge this.  I suggest merging the `develop` branch into your feature branch and resolving on your end locally and then pushing

om-3004:
Hello @ajhollid, I've merged the develop branch into my feature branch and resolved the conflicts on my end.

### Changes made:

- Check resolution before adding monitor for Website Monitoring
- Bypass the resolution check for Ping Monitoring
- Check resolution before adding Page Speed Monitor
- Update the button content to `Checking endpoint` while the URL is being checked for resolution
- Add test cases for check endpoint resolution
- Add the API documentation for new check resolution endpoint
- Validate the URL string before resolving it


https://github.com/user-attachments/assets/b3593a7e-59de-47b4-a477-575f1bab60d5

### Please check the changes made and tried to resolve most of the reviews. If anything is missed, please mention that and I'll resolve it.

Client/src/Pages/PageSpeed/CreatePageSpeed/index.jsx
```diff
@@ -39,6 +40,7 @@ const CreatePageSpeed = () => {
   });
   const [https, setHttps] = useState(true);
   const [errors, setErrors] = useState({});
+  const [isCheckingEndpoint, setIsCheckingEndpoint] = useState(false);
```
ajhollid, line 43: Since you've added an async thunk to the redux store we may as well use the redux loading state rather than handling this manually.

```
	const { monitors, isLoading } = useSelector((state) => state.uptimeMonitors);
```

that will be true whenever any of the async operations is happening

Client/src/Pages/PageSpeed/CreatePageSpeed/index.jsx
```diff
@@ -111,6 +114,17 @@ const CreatePageSpeed = () => {
...
+      const checkEndpointAction = await dispatch(
+        checkEndpointResolution({ authToken, monitorURL: form.url })
+      )
+      if (checkEndpointAction.meta.requestStatus === "rejected") {
+        createToast({ body: "The endpoint you entered doesn't resolve. Check the URL again." });
+        setErrors({ url: "The entered URL is not reachable." });
+        setIsCheckingEndpoint(false);
+        return;
+      }
+      setIsCheckingEndpoint(false);
```
ajhollid, line 126: We can remove this local state management if we use the redux loading state

Client/src/Pages/PageSpeed/CreatePageSpeed/index.jsx
```diff
@@ -333,7 +347,7 @@ const CreatePageSpeed = () => {
             onClick={handleCreateMonitor}
             disabled={Object.keys(errors).length !== 0 && true}
           >
-            Create monitor
+            { isCheckingEndpoint ? "Checking endpoint" : "Create monitor" }
```
ajhollid, line 350: I think we can skip this, we don't need to change the text of this button.

Instead, you can change it to `LoadingButton` and use the `loading` prop to pass `isLoading` from the reudx state to it.  That will indicate an operation is ongoing.

```
import LoadingButton from "@mui/lab/LoadingButton";
```

ajhollid (changes requested):
Looking good!  A few minor chagnes regarding loading state and I think we're good to go.  Thanks for your hard work!

om-3004:
@ajhollid I've added the loading button and also merged the develop branch into my feature branch and resolved the conflicts.

ajhollid:
Will do, I'll review when I have a chance

Client/src/Pages/PageSpeed/CreatePageSpeed/index.jsx
```diff
@@ -16,11 +18,12 @@ import Breadcrumbs from "../../../Components/Breadcrumbs";
...
 
 const CreatePageSpeed = () => {
-	const MS_PER_MINUTE = 60000;
-	const { user, authToken } = useSelector((state) => state.auth);
-	const dispatch = useDispatch();
-	const navigate = useNavigate();
-	const theme = useTheme();
+  const MS_PER_MINUTE = 60000;
+  const { user, authToken } = useSelector((state) => state.auth);
+  const { isLoading } = useSelector((state) => state.uptimeMonitors);
```
ajhollid, line 23: This is potentially problematic as we are using the loading state for the uptime monitors rather than page speed monitors.

I know it's kind of redundant, but I think it would be better to copy the action from the uptimeMonitorSlice to the pageSpeedMonitorSlice so that we're dealing with the correct loading state here.

ajhollid (changes requested):
Looks pretty good, but we should use the pagespeed monitor state in the pagespeed page instead of the uptime monitor state.  That will help us avoid problems down the line.

Also please make sure to use the included `prettierrc` file to format your files according to our standards or you will have conflicts to resolve.

Client/src/Pages/PageSpeed/CreatePageSpeed/index.jsx
om-3004: @ajhollid, If I understood correctly, instead of using state.uptimeMonitors I should use state.pageSpeedMonitors for Creating PageSpeed Monitors? Please correct me if I've misunderstood.

ajhollid (approved):
Looks good to me, thanks for the revisions

Issue #942, opened 2024-10-12
Add "Docs" and "Changelog" to the main menu


gorkem-bwl:
We don't have any links to our documentation on Gitbook and changelog (which is, basically, our release page). 

Here is a sample screen: 

<img width="226" alt="Screenshot 2024-10-11 at 8 40 11 PM" src="https://github.com/user-attachments/assets/c04e72e5-9bdc-4b01-9297-9536088c409a">

Here are the SVG files: 

![changelog](https://github.com/user-attachments/assets/fdcaf04a-1ff5-4de3-9b9a-c3ce599228a3)
![documents](https://github.com/user-attachments/assets/5bafd5b7-bb95-47d0-b777-b384a1a5e5df)

The design (stroke, line weight, size etc) of the SVGs must match the icons above, and none of those properties should be hard coded. 

Docs should go to https://bluewavelabs.gitbook.io/uptime-manager
Changelog should go to https://github.com/bluewave-labs/bluewave-uptime/releases

Both links should open in a new tab.

Abhi-m-anue:
I'm ready to work on it! Can you assign it to me?

gorkem-bwl:
Sure. Thanks for your interest.

gorkem-bwl:
Thank you @Abhi-m-anue for the video. Looks good to me UI-wise.

Abhi-m-anue:
I think this issue can be closed

PR #947, opened 2024-10-12
Added docs and changelog to main menu

Abhi-m-anue:
https://github.com/user-attachments/assets/93e8b3f9-ab48-4a97-8d89-578f0b86e8fd

#942


Client/src/Components/Sidebar/index.jsx
```diff
@@ -498,14 +502,33 @@ function Sidebar() {
...
+                  }
+                  else if(item.path === "changelog"){
+                    window.open(
+                            "https://github.com/bluewave-labs/bluewave-uptime/releases",
+                            "_blank",
+                            "noreferrer"
+                          )
+                  }
+                  else{
+                    navigate(`/${item.path}`)
```
ajhollid, line 529: While this works, I'm not a big fan of `if/else` ladders, especially if it involves repeated logic.  Here we have the `window.open` logic repeated three times, not ideal from a maintainability perspective.

How about the following?  Declare a module scoped URL_MAP that we can use:

```
const URL_MAP = {
  support: "https://github.com/bluewave-labs/bluewave-uptime/issues",
  docs: "https://bluewavelabs.gitbook.io/uptime-manager",
  changelog: "https://github.com/bluewave-labs/bluewave-uptime/releases",
};
```

Now we can get refactor the `onClick` logic:

```
onClick={() => {
  const url = URL_MAP[item.path];
  if (url) {
    window.open(url, "_blank", "noreferrer");
  } else {
    navigate(`/${item.path}`);
  }
}}
```

That's cleaner and more maintanable if we ever decide to change the behavior when one of the links is selected.

Client/src/Components/Sidebar/index.jsx
```diff
@@ -71,6 +73,8 @@ const menu = [
 const other = [
   { name: "Support", path: "support", icon: <Support /> },
   { name: "Settings", path: "settings", icon: <Settings /> },
```
ajhollid, line 75: Could you move `Settings` to the first position here?  I think that makes more sense since users will probably use Settings more than any of the other links.

ajhollid (changes requested):
Nice work :+1:   I think we can do some minor refactoring to make the code a bit more maintainable and readable, and it would be nice to have the Settings link be the first in the list.  Please see comments for details.

Thanks for your hard work and contribution, much appreciated!

Abhi-m-anue:
Thanks for the feedback. That looks cleaner. 
I have updated the code please check

ajhollid (approved):
Awesome, looks good now :+1:  Thanks for making the changes!

ajhollid:
> Thanks for the feedback. That looks cleaner. I have updated the code please check

My pleasure, looks good to me now :rocket:

PR #952, opened 2024-10-13
Fix: Error handling for team invite - Issue#950

### LINKED ISSUE #950: Bug: Error handling for null email in team invite
**Description**
When the 'Send Invite' button is clicked to invite a member, the application correctly checks if the email is valid and displays an error message if it is not. However, when the email is null, the function to invite a member is called and loading animation is shown infinitely, which should not happen.

**To Reproduce**
- Navigate to Account > Team
- Click 'Invite a team member'
- Click 'Send Invite' without entering any value in email.

**Expected behavior**
If the email is null the invite function should not be triggered, and no invite should be sent.

**Screenshots**
Screenshot (72)

**Solution**
Implement a proper validation check to ensure the email is not only in the correct format but also not empty (null). The invite function should only be triggered if the email passes validation.

rishavd3v:
Added conditional check for null email before calling the invite function.

Recording2024-10-14020519

closes #950

### DIFF
--- a/Client/src/Components/TabPanels/Account/TeamPanel.jsx
+++ b/Client/src/Components/TabPanels/Account/TeamPanel.jsx
@@ -49,6 +49,7 @@ const TeamPanel = () => {
   const [tableData, setTableData] = useState({});
   const [members, setMembers] = useState([]);
   const [filter, setFilter] = useState("all");
+  const [isDisabled, setIsDisabled] = useState(true);
   const [errors, setErrors] = useState({});
   const [isSendingInvite, setIsSendingInvite] = useState(false);
 
@@ -137,6 +138,9 @@ const TeamPanel = () => {
 
     setTableData(data);
   }, [members, filter]);
+  useEffect(() => {
+    setIsDisabled(Object.keys(errors).length !== 0 || toInvite.email === "");
+  }, [errors, toInvite.email]); 
 
   // RENAME ORGANIZATION
   const toggleEdit = () => {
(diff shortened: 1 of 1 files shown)


gorkem-bwl:
Looks good to me UI-wise.

Client/src/Components/TabPanels/Account/TeamPanel.jsx
```diff
@@ -172,6 +172,10 @@ const TeamPanel = () => {
   };
 
   const handleInviteMember = async () => {
+    if (!toInvite.email) {
+      setErrors((prev) => ({ ...prev, email: "Email is required." }));
+      return;
+    }
```
ajhollid, line 178: This works, however I think from a UX standpoint it makes more sense to have the `send invite` button disabled until the user actually has inputted something.  Otherwise the user can potentially have thier time wasted by clicking submit with a form that is guaranteed to fail.

How about tracking the state of the input and errors?

```
  const [isDisabled, setIsDisabled] = useState(true);

  useEffect(() => {
    setIsDisabled(Object.keys(errors).length !== 0 || toInvite.email === "");
  }, [errors, toInvite.email]);
```

and setting the disabled state of the loading button based on that?

```
<LoadingButton
  variant="contained"
  color="primary"
  onClick={handleInviteMember}
  loading={isSendingInvite}
  disabled={isDisabled}
>
  Send invite
</LoadingButton>
```

ajhollid (changes requested):
Thanks for catching this issue, totally slipped under the radar.

I think it might make more sense to actually have the button just be disabled until the user enters valid input so they don't waste their time clicking a button that is guaranteed to fail.

Have a look at my comments in the code review for details

ajhollid (approved):
Looks good to me, I think we can go with that.

PR #956, opened 2024-10-14
revamp side-menu

Abhi-m-anue:
https://github.com/user-attachments/assets/e5eeafae-35fa-4e89-80e6-1fa72f51b8a0


ajhollid (changes requested):
Nice work!  Couple readability/maintainability changes that I think could improve things, other than that looks great.

Thanks again for your contribution!

ajhollid (approved):
Nice work, thanks for incorporating those changes :+1:

PR #959, opened 2024-10-15
Migrate backend to ESM

ajhollid:
This PR migrates the project to use ESM.  This is the offical standard of JavaScript and will afford us better compatability.  It will also allow us to eventually do things like tree shaking and static analysis which we can use for optimization.  

- [x] Migrate BE to use ESM
  - [x] Update all imports and exports
  - [x] Resolve minor errors
- [x] Migrate tests   
  - [x] Find solution for mocking sslchecker
    - [x] Rewrote fetchMonitorCertificate to skip using sslchecker


jennifer-gan:
This looks like lots of code refactoring and formatting, tests are all running on my end as well,
please fix the conflict and it should be good to merge

Issue #915, opened 2024-10-08
FE / BE - Refresh tokens


ajhollid:
Currently if you have a valid JWT token that has expired you are logged out and booted to the login page.

It would be preferrable to have use long lived refresh tokens to renew the JWT rather than logout completely.  The final logout would happen when the long lived refresh token expires. 

This would allow users to stay logged in without having long lived JWT tokens in the wild.

sankettank66:
Hello @ajhollid I would like to contribute on this issue.
Can you assign to me?

ajhollid:
> Hello @ajhollid I would like to contribute on this issue.
> Can you assign to me?
> 

Hi @sankettank66 ,

Thanks for your interest in in the issue! 

As this is a core security issue that requires work on both these front and back end I would need to see a proposal with clearly defined plan and steps to achieve the goal before assigning the issue. 

We can then go back and forth and refine the plan together. 

If that's something you you're interested in doing please go ahead and post your plan and we can discuss!

Rushi1109:
Hello @ajhollid,

Here’s a plan for implementing the refresh token system:

1) **Database Schema:** We’ll either extend the current User schema or create a new table to store refresh tokens, associated with user IDs.

2) **Token Generation:** Upon a successful login, we’ll generate two tokens:

- A short-lived JWT access token (for authenticated requests).
- A long-lived refresh token (stored securely on the server).

3) **Client-side Storage:** The Refresh token will be stored in the client side cookies.

4) **Token Renewal:** When the JWT access token expires, the frontend will send a request to a new endpoint (/auth/refresh_token) to obtain a new JWT, using the refresh token. This will ensure users remain logged in seamlessly.

5) **Logout on Expired Refresh Token:** If the refresh token has expired, the user will be logged out and redirected to the login page. All tokens (JWT and refresh) will be invalidated at server-side.

Let me know your thoughts on this approach, and if you’d like any further ajdustments!

ajhollid:
Hey @Rushi1109,

Thanks for taking the time to put a proposal together.  Here's my thoughts:

1.  Database Schema - I don't believe there's any need to store the refresh token.  It should function in the same way as a regular JWT, all we need to do is verify that it is correctly signed and has not expired.

2.  Token generation - This sounds good to me, I don't see any issues here, aside from the fact that we don't need to store the refresh token

3.  Client Side Storage -We use Redux on the client side, so we can store the refersh token in the user's redux store along with the JWT.

[...]

PR #960, opened 2024-10-15
Generate refresh token & add it to response

Rushi1109:
**Changes**

- Created a new issueRefreshToken function to generate refresh tokens with a longer TTL.
- The login and register response now contains both jwtToken and refreshToken.

Implements First Step of #915


Rushi1109:
Hey @ajhollid,

Currently, I have created a new function for generating the refresh token. Can we pass an additional parameter to the issueToken function to determine which token to generate, and use the corresponding secret?

ajhollid:
> Hey @ajhollid,
> 
> Currently, I have created a new function for generating the refresh token. Can we pass an additional parameter to the issueToken function to determine which token to generate, and use the corresponding secret?

I think that's a good idea, there's no reason to have a separate function to create refresh tokens vs access tokens.

Why don't we have `issueToken` take a config object that we can then use in the function to create the tokens.

The config could look like:

```
{
   type: "access" | "refresh"
   expiry: "date"
   roles: ["role_1", "role_2"]
 }
 ```
 
I can't remember exactly how the function works off the top of my head, but it shouldn't be difficult to refactor.

Please make sure tests still pass after refactoring :+1:

File: Server/controllers/authController.js [hunk 1]
```diff
@@ -33,6 +33,24 @@ const issueToken = (payload, appSettings) => {
...
+ */
+const issueRefreshToken = (payload, appSettings) => {
+  try {
+    const tokenTTL = appSettings.refreshTokenTTL ? appSettings.refreshTokenTTL : "7d";
+    return jwt.sign(payload, appSettings.refreshTokenSecret, { expiresIn: tokenTTL });
+  } catch (error) {
+    throw handleError(error, SERVICE_NAME, "issueRefreshToken");
+  }
+};
+
```
ajhollid, line 52: As we discussed, this method is nearly identical to `issueToken`, they shoudl be refactored and combined into one method for issuing tokens.

Also a refresh token should not have a payload, it should consist only of an expiry date, no other information is necessary.

Server/controllers/authController.js
```diff
@@ -81,6 +99,7 @@ const registerUser = async (req, res, next) => {
     const appSettings = await req.settingsService.getSettings();
 
     const token = issueToken(userForToken, appSettings);
+    const refreshToken = issueRefreshToken(userForToken, appSettings);
```
ajhollid, line 102: No payload is needed for a refresh token

ajhollid (changes requested):
Looks good for the most part!

We can refactor `issueToken` as discussed, there's no need for a second token issuance method.  

Refresh token should not have a payload, it should have no data other than an expiry time.

Please make sure to update required tests if you make any changes to the controllers.  Currently test coverage is 100% and all tests are passing:

image


Please make sure this is the case after you make changes.  Currently coverage has been reduced in your PR as changes have been made:

image

Please write tests for your new code as required.  Thanks!

Server/controllers/authController.js
Rushi1109: Sure, I'll refactor the issueToken function.

If we don’t include the payload in the refresh token, how will I retrieve the payload when generating a new JWT access token? (We are including payload while generating access token.)

ajhollid: > Sure, I'll refactor the issueToken function.
> 
> If we don’t include the payload in the refresh token, how will I retrieve the payload when generating a new JWT access token? (We are including payload while generating access token.)

The expired access token will be sent along with the refresh token, so you can grab the payload from the access token

Rushi1109: Oh got it! 

@ajhollid thanks for clarity.

Rushi1109:
Hello @ajhollid, 

Could you review this pull request? I guess this pull request is ready to be merged.

Server/controllers/authController.js
```diff
@@ -12,22 +12,34 @@ require("dotenv").config();
...
+    if(typeOfToken === tokenType.REFRESH_TOKEN) {
+      tokenSecret = appSettings.refreshTokenSecret;
+      tokenTTL = appSettings.refreshTokenTTL ? appSettings.refreshTokenTTL : "7d";
+    }
+    else {
+      tokenSecret = appSettings.jwtSecret;
+      tokenTTL = appSettings.jwtTTL ? appSettings.jwtTTL : "2h";
+      payloadData = payload;
+    }
+
```
ajhollid, line 41: Rather than an `if/else`, why don't we build this up as we go?

```
const tokenTTL = tokenType.REFRESH_TOKEN? (appSettings?refreshTokenTTL?? "2h) : (appSettings?.jwtTTL?? "7d");
   ...
   ... // Set other properties
 jwt.sign(payloadData, tokenSecret, {exipresIn: tokenTTL});
```

The fewer levels of indendtation the better usually, especially when we're doing something like building up a request.

ajhollid (changes requested):
Looks good to me, I think we can replace the `if/else` statement in the `issueToken` function by building up the request to sign the token.  Last change and I think we're good to go :+1: 

Thanks for writing tests too, good coverage :+1:

Server/controllers/authController.js
Rushi1109: Replaced.

ajhollid (approved):
Great work, thank you for being patient while we worked through the process!

Apprecaite all your effort here.

Issue #967, opened 2024-10-15
About page should also show the latest version on GitHub


gorkem-bwl:
When the user goes to Settings page, the About block should show the latest version number pulled from GitHub.

This is currently how it looks:

<img width="1129" alt="image" src="https://github.com/user-attachments/assets/60933191-9b99-47df-a34c-ad7d1165ef37">

This is how it should look like (ignore the Figma design - use the current styling):

<img width="892" alt="image" src="https://github.com/user-attachments/assets/7de017c6-ce28-4c1d-8a05-7f15e42cac7e">

navishmonga:
Can I work on this issue please ?

PDGamerSG:
Can I work on this issue please ?

Celnet-hub:
I have a question to ask. is the version number just after BlueWave Uptime not supposed to be the latest version? Adding about version label seems redundant in my opinion. What do you think @gorkem-bwl

gorkem-bwl:
> I have a question to ask. is the version number just after BlueWave Uptime not supposed to be the latest version? Adding about version label seems redundant in my opinion. What do you think @gorkem-bwl

For example, an admin may be deploying the latest version, however one month later it may be an old version already if there is another version on GitHub. Am I missing something here?

PR #970, opened 2024-10-15
Added the latest repo version in about section of Settings page

navishmonga:
Fixes issue #967 

I have added the tag which will fetch the latest version of the release from the github repo and show it in the about section of settings page.

### DIFF
--- a/Client/src/Pages/Settings/index.jsx
+++ b/Client/src/Pages/Settings/index.jsx
@@ -37,7 +37,7 @@ const Settings = ({ isAdmin }) => {
   const [form, setForm] = useState({
     ttl: checkTTL ? (checkTTL / SECONDS_PER_DAY).toString() : 0,
   });
-  const [version,setVersion]=useState("");
+  const [version,setVersion]=useState("unknown");
   const [errors, setErrors] = useState({});
   const dispatch = useDispatch();
   const navigate = useNavigate();
@@ -45,18 +45,18 @@ const Settings = ({ isAdmin }) => {
   //Fetching latest release version from github
   useEffect(() => {
 		const fetchLatestVersion = async () => {
+			let version="unknown";
 			try {
-				const response = await fetch(
-					"https://api.github.com/repos/bluewave-labs/bluewave-uptime/releases/latest"
-				);
-				if (!response.ok) {
+				const response = await networkService.fetchGithubLatestRelease();
+				if (!response.status===200) {
 					throw new Error("Failed to fetch latest version");
 				}
-				const data = await response.json();
-				setVersion(data.tag_name); // Set the latest version number
+				version=response.data.tag_name;
 			} catch (error) {
 				createToast({ body: error.message || "Error fetching latest version" }); // Set error message
-			}
+			} finally{
+        setVersion(version);
+      }
 		};
 		fetchLatestVersion();
 	},[]);
--- a/Client/src/Pages/Settings/index.jsx
+++ b/Client/src/Pages/Settings/index.jsx
@@ -59,7 +59,7 @@ const Settings = ({ isAdmin }) => {
 			}
 		};
 		fetchLatestVersion();
-	});
+	},[]);
 
   const handleChange = (event) => {
     const { value, id } = event.target;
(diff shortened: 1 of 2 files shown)


gorkem-bwl:
> Fixes issue #967
> 
> I have added the tag which will fetch the latest version of the release from the github repo and show it in the about section of settings page.

Can you add a video of this?

gorkem-bwl:
Looks good to me. What happens if GitHub is unreachable (e.g if the server is behind a firewall that blocks access)?

navishmonga:
it will be blank in place of v1.1.0  and if any error occurs, it will show a notification when settings page loads

gorkem-bwl:
This is ok. Does the GitHub API require any tokens etc?

navishmonga:
No just the url of the repository of which we want to fetch the version of

gorkem-bwl:
OK, thanks for the information.

Client/src/Pages/Settings/index.jsx
```diff
@@ -37,10 +37,30 @@ const Settings = ({ isAdmin }) => {
...
   const dispatch = useDispatch();
   const navigate = useNavigate();
 
+  //Fetching latest release version from github
+  useEffect(() => {
+		const fetchLatestVersion = async () => {
+			try {
+				const response = await fetch(
+					"https://api.github.com/repos/bluewave-labs/bluewave-uptime/releases/latest"
+				);
```
ajhollid, line 51: We have a `NetworkService` class in which we use `Axios` to make all of our network requests.  Can you refactor this to make use of the `NetworkService` so that all of our network requests are in one place?

Client/src/Pages/Settings/index.jsx
```diff
@@ -37,10 +37,30 @@ const Settings = ({ isAdmin }) => {
   const [form, setForm] = useState({
     ttl: checkTTL ? (checkTTL / SECONDS_PER_DAY).toString() : 0,
   });
+  const [version,setVersion]=useState("");
```
ajhollid, line 40: Let's provide a default value here in case network request fails, perhaps "unknown" for now is good.

Client/src/Pages/Settings/index.jsx
```diff
@@ -37,10 +37,30 @@ const Settings = ({ isAdmin }) => {
...
+					"https://api.github.com/repos/bluewave-labs/bluewave-uptime/releases/latest"
+				);
+				if (!response.ok) {
+					throw new Error("Failed to fetch latest version");
+				}
+				const data = await response.json();
+				setVersion(data.tag_name); // Set the latest version number
+			} catch (error) {
+				createToast({ body: error.message || "Error fetching latest version" }); // Set error message
+			}
```
ajhollid, line 59: Why don't we do this in case our network request fails:

```
let version = "unknown"
try {
   ...
   const data = await response.json()
   version = data.tag_name
} catch (error) {
    createToast(...)
} finally {
    setVersion(version)
}
```

Now we'll always have a recent value for `version`, even if it is unknown.

ajhollid (changes requested):
Nice work, looks good for the most part!

I think we can leverage a `try/catch/finally` flow here to make sure we always have a value for `version`. 

We also use a `NetworkService` class to handle all of our network requests using Axios, please refactor the network request to make use of it.

Thanks for your contribution here, much appreciated!

navishmonga:
I have made the requested changes. Can you please review it ?

Client/src/Pages/Settings/index.jsx
```diff
@@ -300,15 +320,16 @@ const Settings = ({ isAdmin }) => {
           </Box>
           <Box>
             <Typography component="h2">BlueWave Uptime v1.0.0</Typography>
+            <Typography component="h2">Latest Version: {version}</Typography>
```
ajhollid, line 323: We don't need both of these, you can just combine them to

`BlueWave Uptime: {version}`

Client/src/Pages/Settings/index.jsx
```diff
@@ -37,10 +37,30 @@ const Settings = ({ isAdmin }) => {
...
+			try {
+				const response = await networkService.fetchGithubLatestRelease();
+				if (!response.status===200) {
+					throw new Error("Failed to fetch latest version");
+				}
+				version=response.data.tag_name;
+			} catch (error) {
+				createToast({ body: error.message || "Error fetching latest version" }); // Set error message
+			} finally{
+        setVersion(version);
```
ajhollid, line 58: Looks good to me!

ajhollid (changes requested):
Looks good to me, we can remove the redundant version line and then we're good to go :+1:

navishmonga:
Done! Can you please check

ajhollid (approved):
Looks good to me, thanks for your hard work here!

PR #972, opened 2024-10-16
Add Colored Labels for HTTP Status Codes

### LINKED ISSUE #965: Convert status code strings to labels
We have a label (chips) component where we can use to identify the status codes in the Incidents table. 

Convert status code into tables where each table has a different color for 403, 502, 5000 and all others. 

Here is the current screenshot of the table where the status codes columns will be converted to labels: 

image

rog404:
### Description:

This PR adds a generic color mapping for HTTP status codes.

Introduced a color mapping for key HTTP status codes:

```ts
const colors = {
		500: {
			textColor: theme.palette.error.main,
		},
		400: {
			textColor: theme.palette.warning.main,
		},
		unresolved: {
			textColor: theme.palette.unresolved.main,
		},
	};
```
- For example, any 400 series code (like 401, 403, 404) will receive the same color as the 400 code. 
- If no specific code is found, the status will default to the unresolved color.

### Screenshot:

Screenshot from 2024-10-16 00-48-03

Close #965


gorkem-bwl:
Just for my info - is there a reason the stroke color of the labels is different from what I see under the status column, e.g those with pink background? 

Also is there a better font color for the labels in yellow bg, as they are barely readable?

rog404:
> Just for my info - is there a reason the stroke color of the labels is different from what I see under the status column, e.g those with pink background?
 
I'm reusing the `ColoredLabel` component, and in it, you don't select the border color; you only choose the main color, and it makes those choices automatically." 

> Also is there a better font color for the labels in yellow bg, as they are barely readable?

Thank you for the feedback. I made some color changes. I hope it meets your approval!

image

rog404:
> Currently the labels look different from what the status labels (the background is ok, stroke is not). Could you let me know why that happens?

I’ll try to explain:

Basically, we have 2 components in the app: BaseLabel and ColoredLabel.

The BaseLabel allows you to pass the style and specify the color for each piece of information: background, stroke, and text. (This is used in the Status Label).

This way, you need to provide all 3 colors to the BaseLabel.

On the other hand, the ColoredLabel only allows you to pass a single color, and it calculates the background and stroke colors. (This is the one I chose to use).

Why did I opt for the ColoredLabel? It was already being exported, and I wanted to reuse it with minimal changes.

I can use the BaseLabel and make it identical to the Status Label if you prefer. Would you like that change?

gorkem-bwl:
Understood. Thanks for the detailed response. 

> I can use the BaseLabel and make it identical to the Status Label if you prefer. Would you like that change?

Yes, please. That way it would look more consistent.

rog404:
Screenshot from 2024-10-16 13-02-58

All done!
Sorry for the confusion! I hope you like the result!

Client/src/Components/HttpStatusLabel/index.jsx
```diff
@@ -0,0 +1,65 @@
...
+            backgroundColor: bgColor,
+            borderColor: borderColor,
+          }}
+      />
+	);
+};
+
+HttpStatusLabel.propTypes = {
+	status: PropTypes.oneOfType([PropTypes.number]).isRequired,
+};
```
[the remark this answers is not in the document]
rog404: `unresolved` is not a valid prop; it simply represents a default color. It should not be derived from `status.code`.

Client/src/Components/HttpStatusLabel/index.jsx
```diff
@@ -0,0 +1,65 @@
...
+      textColor: theme.palette.warning.text,
+      bgColor: theme.palette.warning.bg,
+      borderColor: theme.palette.warning.light,
+	},
+    unresolved: {
+      textColor: theme.palette.unresolved.text,
+      bgColor: theme.palette.unresolved.bg,
+      borderColor: theme.palette.unresolved.light,
+    },
+});
```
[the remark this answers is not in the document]
rog404: I believe this complexity is unnecessary at this stage, particularly given that we already have similar functions within the app. `Client/src/Components/Label/index.jsx`

gorkem-bwl:
Looks great! Thank you.

Client/src/Components/HttpStatusLabel/index.jsx
```diff
@@ -0,0 +1,65 @@
...
+    if (typeof status === 'number' && status >= 100 && status < 600) {
+        return Math.floor(status / 100) * 100;
+    }
+    return "unresolved";
+};
+
+const HttpStatusLabel = ({ status }) => {
+	const theme = useTheme();
+	const colors = getColors(theme);
+	// Get the generic status code
```
marcelluscaio, line 43: Maybe we can remove the comment on the code, since the code itself is very well written and understandable

Client/src/Components/HttpStatusLabel/index.jsx
```diff
@@ -0,0 +1,65 @@
...
+	const genericStatus = getGenericStatus(status);
+	// Look up the color for the status
```
marcelluscaio, line 45: Same here

ajhollid:
@rog404 `HttpStatusLabel` and `StatusLabel` are nearly identical aside from the indicator dot, I don't see the need for an entirely new component here?

You should be able to just add an additional prop to enable or disable the dot and reuse the same component.

rog404:
@ajhollid 

You could be right.

My only concern is not to make this component too large since I will have to include the color codes for HTTP status within the Label.

In this case, I would need to understand the best way to know when the status will be taken into consideration. 

I want to understand, actually, if it makes sense to take this complexity into the component and make it less generic for other cases.

Another alternative would be to remove the 'colors from the component' and let the person who wants the labels know how to work with them.

What do you think?

ajhollid:
> @ajhollid
> 
> You could be right.
> 
> My only concern is not to make this component too large since I will have to include the color codes for HTTP status within the Label.
> 
> In this case, I would need to understand the best way to know when the status will be taken into consideration.
> 
> I want to understand, actually, if it makes sense to take this complexity into the component and make it less generic for other cases.
> 
> Another alternative would be to remove the 'colors from the component' and let the person who wants the labels know how to work with them.
> 
> What do you think?

I think for now we can just use the `StatusLabel` component and move the color codes for the HTTP status codes there.  I don't think we gain much by creating a new component specifically for HTTP stauts codes as it's use case will be limited to just HTTP status codes anyways.

It may make sense to refactor this afterwards to allow the user to select what colors they want and pass that in to the component via props.  We can explore that after in a new issue.

Client/src/Components/Label/index.jsx
```diff
@@ -116,70 +116,93 @@ ColoredLabel.propTypes = {
...
+    400: {
+      dotColor: theme.palette.warning.main,
+      bgColor: theme.palette.warning.bg,
+      borderColor: theme.palette.warning.light,
+    },
+    500: {
+      dotColor: theme.palette.error.text,
+      bgColor: theme.palette.error.bg,
+      borderColor: theme.palette.error.light,
+    },
```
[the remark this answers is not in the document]
rog404: At this moment, I think it’s better to generalize the color codes to 400 and 500 to avoid dealing with a large number of codes and colors. In the future, we can consider making adjustments.

Client/src/Components/Label/index.jsx
```diff
@@ -116,70 +116,93 @@ ColoredLabel.propTypes = {
...
+    "paused",
+    "pending",
+    "cannot resolve",
+    "400",
+    "500",
+    "5000"
+  ]),
+  text: PropTypes.string,
+  customStyles: PropTypes.object,
+  hasDot: PropTypes.bool,
```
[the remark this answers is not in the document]
rog404: We are handling the codes that come from the _backend_, and since code 5000 can be included, we are treating it like the others.

Client/src/Components/Label/index.jsx
```diff
@@ -116,70 +116,93 @@ ColoredLabel.propTypes = {
  * @param {Object} props
  * @param { 'up' | 'down' | 'cannot resolve'} props.status - The status for the label
  * @param {string} props.text - The text of the label
+ * @param {boolean} props.hasDot - Whether to show a dot or not
```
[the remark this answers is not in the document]
rog404: To avoid excessive file edits, we set the default to true, ensuring that any label remains unchanged.

Client/src/Pages/Incidents/IncidentTable/index.jsx
```diff
@@ -129,84 +129,84 @@ const IncidentTable = ({ monitors, selectedMonitor, filter }) => {
...
+                    check.createdAt,
+                    "YYYY-MM-DD HH:mm:ss A",
+                    uiTimezone
+                  );
+                  const getGenericStatus = (status) => {
+                      if (typeof status === 'number' && status >= 100 && status < 600) {
+                          return Math.floor(status / 100) * 100;
+                      }
+                      return "N/A";
+                  };
```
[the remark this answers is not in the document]
rog404: We have already changed this code a bit. So I will let it to a next PR Correction.

rog404:
@ajhollid Can you help me with this conflict?

Every time I resolve this conflict, the code seems to lose its proper indentation. When I use git diff before committing, the code looks correct.

This is making it difficult to keep the PR clear for reviewing the changed files.

ajhollid:
> @ajhollid Can you help me with this conflict?
> 
> Every time I resolve this conflict, the code seems to lose its proper indentation. When I use git diff before committing, the code looks correct.
> 
> This is making it difficult to keep the PR clear for reviewing the changed files.

We use prettier to format, you should be able to just call `prettier` and format the file correctly.

What IDE are you using?  Maybe you've got something overriding our formatting?

ajhollid:
I've merged develop in for you, I think that's what you want.  You can pull that and check.

@marcelluscaio has some ideas he'd like to share with you about this PR however, so please touch base with him and we'll go from there :+1:

rog404:
I'll wait for @marcelluscaio ideas then. He can contact me if he needs anything!" If you need help with anything else, just let me know!

Client/src/Components/HttpStatusLabel/index.jsx
rog404: This file does not exist anymore. So I've resolved this Thread!

ajhollid:
> I'll wait for @marcelluscaio ideas then. He can contact me if he needs anything!" If you need help with anything else, just let me know!

Great, appreciate your patience, the team should get back to you after the weekend.

marcelluscaio:
Hi, @rog404 ! Sorry it took me a while to get back. I really liked your approach in the lighten function!
Right now I am having trouble understanding the change in the IncidentTable. It seems a whole table and logic was erased, and just the label component is there. Is that right?

Client/src/Pages/Incidents/IncidentTable/index.jsx
```diff
@@ -1,218 +1,207 @@
...
-	Paper,
-	Typography,
-	Box,
-} from "@mui/material";
-
-import ArrowBackRoundedIcon from "@mui/icons-material/ArrowBackRounded";
-import ArrowForwardRoundedIcon from "@mui/icons-material/ArrowForwardRounded";
-import { useState, useEffect } from "react";
-import { useSelector } from "react-redux";
-import { networkService } from "../../../main";
```
marcelluscaio, line 20: I mean, this whole component seems to have been deleted, maybe there was a problem when merging?

rog404: You are right! It happens in the @ajhollid merge. I think.

If you go to my first commit. You will see the changes correctly.  https://github.com/bluewave-labs/bluewave-uptime/pull/972/commits/15c4d2b40058372b266f8a1220e3b2144afea155

I will fix it. Thanks.

Client/src/Components/Label/index.jsx
```diff
@@ -119,13 +119,14 @@ ColoredLabel.propTypes = {
...
```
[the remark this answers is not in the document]
rog404: Since I wasn't the one who added customStyles, I chose not to make any changes related to it. There might be a reason for its implementation that was discussed in a previous PR.

Client/src/Components/Label/index.jsx
```diff
@@ -145,6 +146,16 @@ const StatusLabel = ({ status, text, customStyles }) => {
...
+    400: {
+      dotColor: theme.palette.warning.main,
+      bgColor: theme.palette.warning.bg,
+      borderColor: theme.palette.warning.light,
+    },
+    500: {
+      dotColor: theme.palette.error.main,
+      bgColor: theme.palette.error.bg,
+      borderColor: theme.palette.error.light,
+    },
```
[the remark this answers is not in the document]
rog404: Since handling the status code shouldn't be the responsibility of the `label` component, I left it as the other file did.

Client/src/Components/Label/index.jsx
```diff
@@ -165,21 +176,24 @@ const StatusLabel = ({ status, text, customStyles }) => {
...
 		</BaseLabel>
 	);
 };
 
 StatusLabel.propTypes = {
-	status: PropTypes.oneOf(["up", "down", "paused", "pending", "cannot resolve"]),
+	status: PropTypes.oneOf(["up", "down", "paused", "pending", "400", "500", "cannot resolve"]),
 	text: PropTypes.string,
 	customStyles: PropTypes.object,
+  hasDot: PropTypes.bool,
```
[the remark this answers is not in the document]
rog404: In this situation, the only status codes that should come through are related to 500 or 400 errors (I believe). We don't need to use that kind of number formatting at the moment.

Client/src/Components/Label/index.jsx
```diff
@@ -185,10 +198,13 @@ StatusLabel.propTypes = {
     "down",
     "paused",
     "pending",
+    "400",
+    "500",
```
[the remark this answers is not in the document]
rog404: In this situation, the only status codes that should come through are related to 500 or 400 errors (I believe). We don't need to use that kind of number formatting.

Client/src/Pages/Incidents/IncidentTable/index.jsx
rog404: I'm not exactly understanding what's happening, but every time after a merge, my code seems to stop working.

Now I'm getting a 500 error from the backend: 
`{"level":"error","message":"checksQuery is not defined","method":"getTeamChecks","service":"checkController","timestamp":"2024-10-20T15:42:48.177Z"}`

That's strange; I haven't changed anything related to the request, and there were no conflicts with other files during the merge.

ajhollid: > You're right! The issue seems to occur in the merge by @ajhollid.
> 
> If you check my first commit (15c4d2b), you'll see the changes reflected correctly.
> 
> I'll fix this. Thanks!

I don't recall merging any code that removes a component from, could you link to the diff so I can check it out?

ajhollid: > I'm not exactly understanding what's happening, but every time after a merge, my code seems to stop working.
> 
> Now I'm getting a 500 error from the backend: 
> `{"level":"error","message":"checksQuery is not defined","method":"getTeamChecks","service":"checkController","timestamp":"2024-10-20T15:42:48.177Z"}`
> 
> That's strange; I haven't changed anything related to the request, and there were no conflicts with other files during the merge.

Were you able to figure out what was going on here? If not I'll look into it today

ajhollid: > I'm not exactly understanding what's happening, but every time after a merge, my code seems to stop working.
> 
> Now I'm getting a 500 error from the backend: `{"level":"error","message":"checksQuery is not defined","method":"getTeamChecks","service":"checkController","timestamp":"2024-10-20T15:42:48.177Z"}`
> 
> That's strange; I haven't changed anything related to the request, and there were no conflicts with other files during the merge.

@rog404 fixed the issue, there was a missing variable declaration in one of the modules.  If you merge develop in to you r branch you should be good to go.

rog404:
@ajhollid  and @marcelluscaio 
All the changes have been finalized, and everything should be in order now. Could you please review and confirm that everything looks good?

Thanks again for all the help! :love_you_gesture:

Client/src/Pages/Incidents/IncidentTable/index.jsx
rog404: Thanks!

Client/src/Components/Label/index.jsx
```diff
@@ -116,13 +116,14 @@ ColoredLabel.propTypes = {
...
  * @param {string} props.text - The text of the label
+ * @param {boolean} props.hasDot - Whether to show a dot or not
  * @returns {JSX.Element}
  * @example
  * // Render an active label
  * <StatusLabel status="up" text="Active" />
  */
 
-const StatusLabel = ({ status, text, customStyles }) => {
+const StatusLabel = ({ status, text, hasDot = true, customStyles }) => {
```
marcelluscaio, line 126: Hey, Rogerio! Thanks for all you patience and effort to contribute to this project. 
Happy everything is working now. 
I worry about the hasDot property, though. It would allow someone to pass a dot to a number status, which does not exist in the design system. 

I see two ways out of this:
1) allowing status or httpStatus as props. I think something like this would type the props accordingly
const baseProps = {
	text: PropTypes.string.isRequired,
	customStyles: PropTypes.object,
};

const statusShape = PropTypes.shape({
	status: PropTypes.oneOf(["up", "down", "paused", "pending"]).isRequired,
	...baseProps,
});

const httpStatusShape = PropTypes.shape({
	httpStatus: PropTypes.oneOf(["400", "500", "cannot resolve"]).isRequired,
	...baseProps,
});

StatusLabel.propTypes = {
	statusProps: PropTypes.oneOfType([statusShape, httpStatusShape]).isRequired,
};

This way we us httpStatus code as proxy to know if we are dealing with status or httpStatus, and apply the dot accordingly. 

2) Get back to using two components. StatusLabel and HttpStatusLabel would both implement the BaseLabel.
I believe this approach is better.

In either approach, I suggest keeping the getGenericStatus function inside of the component.

ajhollid: @rog404 @marcelluscaio Agreed, why don't we refactor out `BaseLabel` from the `Label` component and have both `StatusLabel` and `HttpStatusLabel` extend that.

That should keep confusion to a minumum and make use of the components very straightforward.

Appreciate you going back and forth on this one @rog404 , sometimes it takes some discussion to tease out the best approach :joy:

rog404: Don't worry! I believe a good PR strategy should be just like this—well thought-out and full of great ideas. As soon as I have some time, I'll start working on this refactoring.

ajhollid: > Don't worry! I believe a good PR strategy should be just like this—well thought-out and full of great ideas. As soon as I have some time, I'll start working on this refactoring.
> 

Absolutely, good results come from good discussions 👌Looking forward to seeing what you build!

rog404:
@ajhollid  and @marcelluscaio 

Could you guys help me? Currently, there are 2 status codes in the code that are not valid:

"Network Error (code: 5000)"
"No Response (code: N/A)"
My suggestion is to keep a single value to express cases where we don't have a valid HTTP status code. What do you suggest we can do in these cases?

image

ajhollid:
> @ajhollid  and @marcelluscaio 
> 
> Could you guys help me? Currently, there are 2 status codes in the code that are not valid:
> 
> "Network Error (code: 5000)"
> "No Response (code: N/A)"
> My suggestion is to keep a single value to express cases where we don't have a valid HTTP status code. What do you suggest we can do in these cases?
> 
> image
> 

@rog404 5000 should be internal errors, I'll look into the response with no code.

What are the conditions to replicate a response with no code?

rog404:
@ajhollid 
I'm not sure.

But I believe is when you add a wrong url.
In my case: I add `localhost:3000 `(without http)

image

ajhollid:
> @ajhollid 
> I'm not sure.
> 
> But I believe is when you add a wrong url.
> In my case: I add `localhost:3000 `(without http)
> 
> image
> 

OK ill do some testing as soon as I can. 

In the meantime you could just display a default color for incidents where a code doesn't exist.

Client/src/Components/Label/index.jsx
```diff
@@ -145,6 +145,16 @@ const StatusLabel = ({ status, text, customStyles }) => {
...
+		400: {
+			dotColor: theme.palette.warning.main,
+			bgColor: theme.palette.warning.bg,
+			borderColor: theme.palette.warning.light,
+		},
+		500: {
+			dotColor: theme.palette.warning.main,
+			bgColor: theme.palette.warning.bg,
+			borderColor: theme.palette.warning.light,
+		},
```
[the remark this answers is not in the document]
rog404: As we do not have defined correct color for it. We will add this temp colors. It can be changed in a near future.

Client/src/Components/HttpStatusLabel/index.jsx
```diff
@@ -0,0 +1,65 @@
...
+ * // Render a http status label
+ * <HttpStatusLabel status="404" />
+ */
+
+const getRoundedStatusCode = (status) => {
+	if (typeof status === "number" && status >= 100 && status < 600) {
+		return Math.floor(status / 100) * 100;
+	}
+	return status;
+};
```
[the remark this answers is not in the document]
rog404: IMO, The function cost is so low. We do not need to worry about performance here.

ajhollid:
> @ajhollid 
> 
> What I did here was create only the new `HttpStatusLabel` component separately, focusing on addressing the specific issue mentioned in this ticket. I chose not to modify StatusLabel to avoid broader changes since the original issue didn’t call for a full refactor of the Label file. Given that Label also contains the `ColoredLabel` component in addition to `StatusLabel`, it might be worth opening a new issue to explore a more comprehensive refactor.
> 
> I aimed to keep modifications in unrelated files to a minimum, making the PR concise for faster review. Do you think prioritizing these refactors is necessary right now, or would it be more beneficial to address other issues that have a direct impact on the app?
> Thanks.

Hey @rog404 ,

In general I'd prefer to avoid adding technical debt to the project by adding a feature that we already know is going to need a refactor.

That said this is a relatively minor feature and the refactor won't be especially difficult or time consuming.

The front end is the domain of @marcelluscaio and @jennifer-gan and the tech debt ends up on their plate at the end of the day, so I'll leave the final call up to them.

I'll follow up once they get their reviews in and we'll go from there, thanks!

Client/src/Pages/Incidents/IncidentTable/index.jsx
```diff
@@ -196,7 +197,9 @@ const IncidentTable = ({ monitors, selectedMonitor, filter }) => {
 												/>
 											</TableCell>
 											<TableCell>{formattedDate}</TableCell>
-											<TableCell>{check.statusCode ? check.statusCode : "N/A"}</TableCell>
+											<TableCell>
+												<HttpStatusLabel status={check.statusCode || "N/A"} />
```
marcelluscaio, line 201: The implementation is great. Something I was thinking, what if status code is somthing other than undefined, and it isn't a number between 100 and 600? We sould be sending an uncovered scenario to the HttpStatusLabel component. 

What I suggest is we just pass down chueck.statusCode, and inside of http status label we deal with that

Client/src/Components/HttpStatusLabel/index.jsx
```diff
@@ -0,0 +1,65 @@
...
+ * // Render a http status label
+ * <HttpStatusLabel status={404} />
+ */
+
+const getRoundedStatusCode = (status) => {
+	if (typeof status === "number" && status >= 100 && status < 600) {
+		return Math.floor(status / 100) * 100;
+	}
+	return status;
+};
```
marcelluscaio, line 21: This comment continues the one made in IncidentTable. We are going to deal with different scenarios here. if we don't have a number between 100 and 600, we are going to return  "default" (I suggest changing N/A to default).

marcelluscaio (commented):
I like the implementation, and I asked for minor changes, to prevent some edge cases. 
Just to clarify, what is the change that would be need on BaseLabel in the future?

ajhollid:
> I like the implementation, and I asked for minor changes, to prevent some edge cases. 
> Just to clarify, what is the change that would be need on BaseLabel in the future?

I was under the impression the idea was to refactor out BaseLabel to an separate component and have StatusLabel and HttpStatusLabel extend that component, if that's not the case then please disregard.

Aside from that the implementation looks good to me too 👌

Client/src/Components/HttpStatusLabel/index.jsx
ajhollid: As an addendum to this comment, I would say that type checking here isn't necessary, we should guarantee that this function only receives a number upstream anyways.

This function also has a mixed return type which is dangerous.  It will either return a `number` or `undefined`.

JavaScript's type coercion allows us to get away with this, but now we must always do a check whenever we use this function upstream to see if we have a value or not.

As @marcelluscaio mentioned we can avoid this by returning a default value.

```
if(status >= 100 && status < 600){
   // return roudned status code
}
return defaultCode
```

I'd say that is a safer and more reliable function.  Otherwise we must always do

```
if ( getRoundedStatusCode(status) !== undefined ) {
   // do stuff here
}
```

or at the very least something like 

```
const roundedStatusCode = getRoundedStatusCode(status) || 5000
```

whenever we use the function.  This adds extra verbosity and complexity to the code which is easily avoided.

 One of the reasons I'm an advocate for strongly typed languages.. or TypeScript at least :joy:

Client/src/Components/HttpStatusLabel/index.jsx
```diff
@@ -0,0 +1,65 @@
...
+		},
+		"N/A": {
+			dotColor: theme.palette.unresolved.main,
+			bgColor: theme.palette.unresolved.bg,
+			borderColor: theme.palette.unresolved.light,
+		},
+	};
+
+	const { borderColor, bgColor, dotColor } =
+		colors[getRoundedStatusCode(status)] || colors["N/A"];
```
ajhollid, line 44: We can avoid the nullish coalescing here by slightly refacotring the `getRoudnedStatusCode` funcition mentioned above

Client/src/Pages/Incidents/IncidentTable/index.jsx
ajhollid: Agreed, let's handle the logic inside of `HttpStatusLabel` since we're dealing with it there for colors anyway

ajhollid (changes requested):
Good implementation, some minor changes RE the status code and we're good to go :+1:

rog404:
Hello,
I just wanted to inform you that I recently underwent surgery, so it may take me a few extra days to complete this task. I apologize for the delay

marcelluscaio:
There is no problem about that @rog404 . I hope you recover soon!

gorkem-bwl:
> Hello, I just wanted to inform you that I recently underwent surgery, so it may take me a few extra days to complete this task. I apologize for the delay

Hope you get well soon!

ajhollid:
@rog404 get well soon friend, no rush on the work, it will still be there when you're healed :+1:

rog404:
Hi @marcelluscaio and @ajhollid,

I’ve made some changes to this PR. Could you take a moment to review them and let me know if everything looks good?

ajhollid:
> Hi @marcelluscaio and @ajhollid,
> 
> I’ve made some changes to this PR. Could you take a moment to review them and let me know if everything looks good?

Hey @rog404,

We're in a release cycle so we've got our hands full for the next couple days, but we'll get on it as soon as possible :smile:   Thanks for your contributions again!

ajhollid (approved):
Looks good to me, thanks for your contribution!

PR #973, opened 2024-10-16
Add middleware to verify refresh token

Rushi1109:
**Changes**
Add a middleware that check for authenticity of refresh token.

Implements middleware for #915

### DIFF
--- a/Server/middleware/verifyJWT.js
+++ b/Server/middleware/verifyJWT.js
@@ -2,6 +2,9 @@ const jwt = require("jsonwebtoken");
 const SERVICE_NAME = "verifyJWT";
 const TOKEN_PREFIX = "Bearer ";
 const { errorMessages } = require("../utils/messages");
+const { getTokenFromHeaders } = require("../utils/utils");
+const { handleError } = require("../controllers/controllerUtils");
+
 /**
  * Verifies the JWT token
  * @function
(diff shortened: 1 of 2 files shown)


ajhollid (changes requested):
I don't think this is exactly what we need here.  I think what should happen in the verifyJWT middleware now is the following:

1.  Check for valid **access token**
2.  If **access token** cannot be verified, call `next()` with an error
3. If **access token** is verified but expired, check for valid and unexpired refresh token
   3a. If **refersh token** invalid or expired, call `next()` with a 401 error triggering logout
   3b. If **referesh token** is valid and not expired, call `next()` with a 403 error, FE can 
         use 403 code to trigger request for new **access token**
4.  If **access token** is valid and not expried, carry on with request, and decoded user to 
`req` as usual and call `next()`

I believe this is the process to follow here, let me know if you want to discuss any changes to it.

Rushi1109:
Hey @ajhollid, Now I understand the process of verifyJWT. I'll make changes to that middleware.

Let me know if the verifyRefreshToken middleware requires any changes.

ajhollid:
> Hey @ajhollid, Now I understand the process of verifyJWT. I'll make changes to that middleware.
> 
> Let me know if the verifyRefreshToken middleware requires any changes.

I don't think there is any need for `verifyRefreshToken` middleware, there should just be one token verification middleware as outlined above

Rushi1109:
When FE will receive 403, it will send a request on newly created endpoint. Won't I need a middleware to verify the refresh token at that point?

ajhollid:
> When FE will receive 403, it will send a request on newly created endpoint. Won't I need a middleware to verify the refresh token at that point?
> 
> And that means refresh token will also be sent in the body of request in every single request?? Am I Right?

Hey @Rushi1109 ,

Yes, when the FE receives 403 it will make a request to an endpoint on the BE to request a new access token.  You can verify the refersh token in the controller for this endpoint.  This is the only place a refresh token is going verified, so I don't see a need for middleware for it.  

Yes, refresh token will have to be sent with all requests along with the access token.

Server/middleware/verifyJWT.js
```diff
@@ -1,7 +1,10 @@
 import jwt from "jsonwebtoken";
+import { errorMessages } from "../utils/messages.js";
+import { handleError } from "../controllers/controllerUtils.js";
```
ajhollid, line 3: We don't really need `handleError` here, all it does is add status and service and a default status code to the Error if they don't already exist.

We've already added that information, so we can just call `next(error)`

ajhollid (changes requested):
Looks pretty good to me.  We don't need the `handleError` function here since it doesn't add anything for us here.

Server/middleware/verifyJWT.js
Rushi1109: Sure. let me just update it.

Rushi1109:
> We don't need the `handleError` function here since it doesn't add anything for us here.

I have removed it.

ajhollid (approved):
Looks good, thanks for making requested changes!

Rushi1109:
> Looks good, thanks for making requested changes!

😀

PR #974, opened 2024-10-16
Fix: table header - issue#963

### LINKED ISSUE #963: Incidents, maintenance and team table header fix
The incidents, maintenance and team table header should look like the table header on the main dashboard. Double check the style, font size, weight and color.

This is the incidents table header: 

image

This is the maintenance table header:

image

This is the Teams table header:
image

This is the table header on the dashboard. All previous 3 table headers should look like this.

image

rishavd3v:
Fixed styling of table headers of Incidents, maintenance and team table section as discussed in issue #963. Added a common class to the parent component that renders the table.

Before:
prev4
 
Fixed:
Screenshot (80)

closes #963

### DIFF
--- a/Client/src/Components/BasicTable/index.css
+++ b/Client/src/Components/BasicTable/index.css
@@ -120,7 +120,7 @@
   opacity: 0.4;
 }
 
-.monitors .MuiTable-root .MuiTableHead-root .MuiTableCell-root {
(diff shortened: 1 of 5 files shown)


ajhollid (approved):
Looks good, thanks for taking care of that :+1:

PR #975, opened 2024-10-17
Add ssl-checker, refactor fetchMonitorCertificate

ajhollid:
This PR adds `ssl-checker` back to the project, in the end it is more reliable than using TLS to fetch certificates

- [x] Add ssl-checker to project
  - [x] refactor fetchMonitorCertificate
  - [x] rewrite tests

### DIFF
--- a/Server/controllers/controllerUtils.js
+++ b/Server/controllers/controllerUtils.js
@@ -12,32 +12,15 @@ const handleError = (error, serviceName, method, status = 500) => {
 	return error;
 };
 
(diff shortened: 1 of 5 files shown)


marcelluscaio (approved):
I don't know the technical difference between TLS and SSL, but the implementation is clean and well written

jennifer-gan:
I just saw the email request for review, I thought TLS is the replacement of SSL? all versions of SSL are deprecated,
does that mean we should use TLS certificates from then on?

PR #986, opened 2024-10-18
FE - Advanced Settings Page Validation and Error Handling

jennifer-gan:
Add validation to advanced settings page
- Extract buildErrors and hasValidationErrors to error.js file
- Add advanced settings validation rule
- Apply  advanced settings validation rule in field change and form submission

Below is the screen recording of the validation :
[Screencast from 2024-10-18 14-07-59.webm](https://github.com/user-attachments/assets/5715c760-1aee-4857-8d3a-a4beddfc7745)

### DIFF
--- a/Client/src/Pages/AdvancedSettings/index.jsx
+++ b/Client/src/Pages/AdvancedSettings/index.jsx
@@ -12,45 +12,47 @@ import { useNavigate } from "react-router";
 import { getAppSettings, updateAppSettings } from "../../Features/Settings/settingsSlice";
 import { useState, useEffect } from "react";
 import Select from "../../Components/Inputs/Select";
+import { advancedSettingsValidation } from "../../Validation/validation";
+import { buildErrors, hasValidationErrors } from "../../Validation/error";
 
 const AdvancedSettings = ({ isAdmin }) => {
-	const navigate = useNavigate();
+  const navigate = useNavigate();
+  useEffect(() => {
+    if (!isAdmin) {
+      navigate("/");
+    }
+  }, [navigate, isAdmin]);
 
-	useEffect(() => {
-		if (!isAdmin) {
-			navigate("/");
-		}
-	}, [navigate, isAdmin]);
+  const theme = useTheme();
+  const { authToken } = useSelector((state) => state.auth);
+  const dispatch = useDispatch();
+  const settings = useSelector((state) => state.settings);
+  const [localSettings, setLocalSettings] = useState({
+    apiBaseUrl: "",
+    logLevel: "debug",
+    systemEmailHost: "",
+    systemEmailPort: "",
+    systemEmailAddress: "",
+    systemEmailPassword: "",
+    jwtTTL: "",
+    dbType: "",
+    redisHost: "",
+    redisPort: "",
+    pagespeedApiKey: "",
+  });
+  const [errors, setErrors] = useState({});
 
-	const theme = useTheme();
-	const { authToken } = useSelector((state) => state.auth);
-	const dispatch = useDispatch();
-	const settings = useSelector((state) => state.settings);
-	const [localSettings, setLocalSettings] = useState({
-		apiBaseUrl: "",
-		logLevel: "debug",
-		systemEmailHost: "",
-		systemEmailPort: "",
-		systemEmailAddress: "",
-		systemEmailPassword: "",
-		jwtTTL: "",
-		dbType: "",
-		redisHost: "",
-		redisPort: "",
-		pagespeedApiKey: "",
-	});
-
-	useEffect(() => {
-		const getSettings = async () => {
-			const action = await dispatch(getAppSettings({ authToken }));
-			if (action.payload.success) {
-				setLocalSettings(action.payload.data);
-			} else {
-				createToast({ body: "Failed to get settings" });
-			}
-		};
-		getSettings();
-	}, [authToken, dispatch]);
+  useEffect(() => {
+    const getSettings = async () => {
+      const action = await dispatch(getAppSettings({ authToken }));
+      if (action.payload?.success) {
+        setLocalSettings(action.payload.data);
+      } else {
+        createToast({ body: "Failed to get settings" });
+      }
+    };
+    getSettings();
+  }, [authToken, dispatch]);
 
 	const logItems = [
 		{ _id: 1, name: "none" },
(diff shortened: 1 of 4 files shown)


Client/src/Validation/validation.js
```diff
@@ -128,10 +128,43 @@ const maintenanceWindowValidation = joi.object({
 	monitors: joi.array().min(1),
 });
 
+const advancedSettingsValidation = joi.object({
+	apiBaseUrl: joi.string().uri({ allowRelative: true }).trim().messages({
+		"string.empty": "API base url is required.",
+		"string.uri": "The URL you provided is not valid.",
+	}),
+	logLevel: joi.string(),
```
ajhollid, line 136: Acceptable values for `logLevel` are `["debug", "none", "error", "warn"]`

FYI for future FE validaiton you can also check the BE validation if it's a value that's being sent to the BE.  Validations are in the `./Server/validation/joi.js` file:

```
logLevel: joi.string().valid("debug", "none", "error", "warn").allow(""),
```

Client/src/Validation/error.js
```diff
@@ -0,0 +1,25 @@
+const buildErrors = (prev, id, error) => {
+	const updatedErrors = { ...prev };
+	if (error) {
+		updatedErrors[id] = error.details[0].message;
```
ajhollid, line 4: What about providing a default error if this nested value isn't available for whaterver reason?

```
updatedErrors[id] = error?.details[0]?.message ?? "Validation error";
```

Something like that?

Client/src/Validation/error.js
```diff
@@ -0,0 +1,25 @@
...
+};
+
+const hasValidationErrors = (form, validation, setErrors) => {
+	const { error } = validation.validate(form, {
+		abortEarly: false,
+	});
+	if (error) {
+		const newErrors = {};
+		error.details.forEach((err) => {
+			newErrors[err.path[0]] = err.message;
```
ajhollid, line 18: Same for this one, always good to fail gracefully if possible

Client/src/Validation/validation.js
```diff
@@ -128,10 +128,43 @@ const maintenanceWindowValidation = joi.object({
...
+			"string.empty": "JWT TTL is required."			
+		}),
+	dbType: joi.string().trim().messages({
+		"string.empty": "DB type is required",
+	}),
+	redisHost: joi.string().trim().messages({
+		"string.empty": "Redis host is required",
+	}),
+	redisPort: joi.string().allow('').custom((value, helpers) => {
+		if(value && isNaN(parseInt(value))){
```
ajhollid, line 155: `isNan` coerces its operand to into a number, so it doesn't always behave the way you want it to.

[Recommended approach](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/isNaN#description) is to use `Number.isNan()` instead.

ajhollid (changes requested):
Looks good to me!

Couple minor changes and we should be good to go, please see my comments in the code review!

Thanks for your hard work here!

Client/src/Validation/validation.js
jennifer-gan: fixed in https://github.com/bluewave-labs/bluewave-uptime/pull/986/commits/8ec2f2f16b916eb15cd5d092da768e29b55e6da2, thanks!

Client/src/Validation/error.js
jennifer-gan: fixed in https://github.com/bluewave-labs/bluewave-uptime/pull/986/commits/8ec2f2f16b916eb15cd5d092da768e29b55e6da2, thanks!

Client/src/Validation/validation.js
jennifer-gan: fixed in https://github.com/bluewave-labs/bluewave-uptime/pull/986/commits/8ec2f2f16b916eb15cd5d092da768e29b55e6da2, thanks!

ajhollid: @jennifer-gan looks like `isNan` is still present in the latest commit?  Maybe it didn't get added?

Client/src/Validation/validation.js
```diff
@@ -128,10 +128,43 @@ const maintenanceWindowValidation = joi.object({
...
+			return helpers.message("Redit port must be a number")
```
ajhollid, line 156: This should be "Redis", much as I love "reddit" :stuck_out_tongue:

ajhollid (approved):
Looks like `isNan` is still pressent, and there's a small typo: "Redit" instead of "Redis

I'll mark this approved as I'm good to go otherwise, if you can just resolve those last two we can merge in.

Client/src/Validation/validation.js
```diff
@@ -128,10 +128,43 @@ const maintenanceWindowValidation = joi.object({
...
+			"string.empty": "JWT TTL is required."			
+		}),
+	dbType: joi.string().trim().messages({
+		"string.empty": "DB type is required",
+	}),
+	redisHost: joi.string().trim().messages({
+		"string.empty": "Redis host is required",
+	}),
+	redisPort: joi.string().allow('').custom((value, helpers) => {
+		if(value && isNaN(parseInt(value))){
```
marcelluscaio, line 155: Can we add custom messages to the other cases as well?

Client/src/Validation/error.js
```diff
@@ -0,0 +1,25 @@
+const buildErrors = (prev, id, error) => {
+	const updatedErrors = { ...prev };
+	if (error) {
+		updatedErrors[id] = error.details[0].message?? "Validation error";
+	} else {
+		delete updatedErrors[id];
+	}
```
marcelluscaio, line 7: I don't get the use case of this leg of the if statement

Client/src/Pages/AdvancedSettings/index.jsx
```diff
@@ -72,189 +74,210 @@ const AdvancedSettings = ({ isAdmin }) => {
...
+    const { error } = advancedSettingsValidation.validate(
+      { [id]: value },
+      {
+        abortEarly: false,
+      }
+    );
+    setErrors((prev) => {
+      return buildErrors(prev, id, error);
+    });    
+  };
```
marcelluscaio, line 89: Validating on change can be bad for performance, because we are running validation and state changes even if the user is filling correctly. 
I would suggest we validate on submit.

jennifer-gan: That can be a  design decision made before, I've seen both

Client/src/Validation/validation.js
jennifer-gan: Resolved in : https://github.com/bluewave-labs/bluewave-uptime/commit/4966f3fd38e43960e2e6d3fbea78afb205c08883

Client/src/Validation/validation.js
jennifer-gan: jwt ttl custom validation is added :Resolved in : https://github.com/bluewave-labs/bluewave-uptime/commit/4966f3fd38e43960e2e6d3fbea78afb205c08883

PR #1020, opened 2024-10-22
913 fe advanced settings page validation and error handling

jennifer-gan:
- [x] Use OnBlur to check validation errors for Field and Select components
- [x] Add select dropdown "jwtTTLUnits" for hours and days with placeholder "Select time"
- [x] When populate,
  - [x] fetch the jwtTTL and split into 2 fields (jwtTTLNum and jwtTTLUnits)
- [x] When save
  - [x] validate number and selected item ( both required)
  - [x] After successful validation, combined the jwtTTLNum and jwtTTLUnits into jwtTTL to persist


Client/src/Validation/validation.js
```diff
@@ -128,10 +128,57 @@ const maintenanceWindowValidation = joi.object({
...
+		})
+		.custom((value, helpers) => {
+			// test against regx start with a number followed by number or .,
+			//ends with d or h
+			const regex = /^[1-9]+\d*(.)*(d|h)$/;
+			const found = value.match(regex);
+			if (!found) {
+				return helpers.message(
+					"JWT TTL should start with a non zero number and ends with 'd' or 'h'."
+				);
```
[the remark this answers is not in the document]
ajhollid: I guess the validation can be done on the combined string with nunber and unit?

The regex is an issue though, I think this will allow for strings like "123abcd" which is not what is intended I don't think.  If you drop the `(.)` I think you're good to go.

You can test the [regex here](https://regex101.com/)

Client/src/Validation/validation.js
```diff
@@ -128,10 +128,57 @@ const maintenanceWindowValidation = joi.object({
...
+		"string.empty": "DB type is required.",
+	}),
+	redisHost: joi.string().trim().messages({
+		"string.empty": "Redis host is required.",
+	}),
+	redisPort: joi
+		.string()
+		.allow("")
+		.custom((value, helpers) => {
+			if (value && Number.isNaN(parseInt(value))) {
```
ajhollid, line 169: :+1:

Client/src/Pages/AdvancedSettings/index.jsx
```diff
@@ -75,9 +77,20 @@ const AdvancedSettings = ({ isAdmin }) => {
...
 		setLocalSettings({ ...localSettings, [id]: value });
+		const { error } = advancedSettingsValidation.validate(
+			{ [id]: value },
+			{
+				abortEarly: false,
+			}
+		);
+		setErrors((prev) => {
+			return buildErrors(prev, id, error);
+		});
```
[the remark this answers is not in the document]
ajhollid: @jennifer-gan @marcelluscaio just a heads up this change will require modification to any components that do not expose `onBlur`.  

Off the top of my head `Field` does not currently expose `onBlur`, but it's an easy fix.  Just add an `onBlur` prop to the `Field` component and pass it down to the `TextField` subcomponent.  

Same is probably true of `Select` and any other components created on our end, I don't think any of the previous devs used `onBlur` for anything.

Don't forget to add a proptype and update any JSdocs affected :+1:

ajhollid (changes requested):
Thanks for making changes here!

There's a slight issue with the regex which allows for unintended valus like `123abcd`, but easily fixed.

To use `onBlur` instead of `onChange` will require exposing `onBlur` for any components that currently do not, which should be all of the input components in the project I think

Client/src/Validation/validation.js
jennifer-gan: Just to confirm , 
1. we will need to add a dropdown for day and hour ?
2. Add onBlur for input components
3. Fix the regex to allow only numbers ending with a 'd' or 'h' and no fraction numbers(remove the dot)

marcelluscaio: 1 - Yes, and validate on that step instead of onChange
2- Since we are splitting the input into a number and a time unit, we could split validation as well. This way you could validate if is a number, and the other would use the same sort of validation we used for the other select.
3 - Yes, I think so. Gorkem sent a message on discord with the figma image

Client/src/Pages/AdvancedSettings/index.jsx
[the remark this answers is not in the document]
jennifer-gan: I will update Field and Select, also looking into proptype and JSdocs etc

ajhollid (changes requested):
@jennifer-gan here is where you will have to split the vercel/ms time format into jwtTTL and jwtTTLUnits:

```
	useEffect(() => {
		const getSettings = async () => {
			const action = await dispatch(getAppSettings({ authToken }));
			if (action.payload.success) {
				// format action.payload.data into correct format
				setLocalSettings(action.payload.data);
			} else {
				createToast({ body: "Failed to get settings" });
			}
		};
		getSettings();
	}, [authToken, dispatch]);
```

And then you will have to do the opposite in `handleSave` to prepare for submission to the BE as it requries the vercel/ms format

Client/src/Pages/AdvancedSettings/index.jsx
```diff
@@ -180,21 +239,37 @@ const AdvancedSettings = ({ isAdmin }) => {
...
 							onChange={handleChange}
+							onBlur={handleBlur}
+							error={errors.jwtTTLNum}
 						/>
+						<Select
+							id="jwtTTLUnits"
+							label="JWT TTL Units"
+							name="jwtTTLUnits"
+							placeholder="Select time"
+							isHidden={true}
```
marcelluscaio, line 257: Why are we declaring isHidden true?

Client/src/Pages/AdvancedSettings/index.jsx
```diff
@@ -66,24 +81,57 @@ const AdvancedSettings = ({ isAdmin }) => {
 		warn: 4,
 	};
 
+	const unitItems = [
+		{ _id: "d", name: "days" },
+		{ _id: "h", name: "hours" }
+	];
```
marcelluscaio, line 87: Maybe we don't need this array. We can do something like:
const unitItems = Object.keys(unitItemLookup).map((key, index) => ({_id: index, name: key})).

Like that we only have one place to update if we need to add or remove items

Client/src/Validation/validation.js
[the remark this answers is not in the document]
jennifer-gan: resolved in :https://github.com/bluewave-labs/bluewave-uptime/pull/1020/commits/57edff36c9920fc569f24c7c11dab21f81704b12

Client/src/Pages/AdvancedSettings/index.jsx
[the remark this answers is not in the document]
jennifer-gan: resolved in :https://github.com/bluewave-labs/bluewave-uptime/pull/1020/commits/57edff36c9920fc569f24c7c11dab21f81704b12
https://github.com/bluewave-labs/bluewave-uptime/pull/1020/commits/410f3b7ab1e8abf8bb01f723f6dfcf45b8383c8c

Client/src/Validation/validation.js
jennifer-gan: resoved in : https://github.com/bluewave-labs/bluewave-uptime/pull/1020/commits/a4344b89a23692d4d7cb2a534b90d99044d5ae24
https://github.com/bluewave-labs/bluewave-uptime/pull/1020/commits/57edff36c9920fc569f24c7c11dab21f81704b12

Client/src/Pages/AdvancedSettings/index.jsx
jennifer-gan: The placeholder "Select time" will be a hidden value

Client/src/Pages/AdvancedSettings/index.jsx
jennifer-gan: Sounds good, changed here: https://github.com/bluewave-labs/bluewave-uptime/pull/1020/commits/1526b50fd2044ca2df035c6f0a41eaab0e3c56c2

Client/src/Validation/validation.js
```diff
@@ -128,10 +128,57 @@ const maintenanceWindowValidation = joi.object({
...
 });
 
+const advancedSettingsValidation = joi.object({
+	apiBaseUrl: joi.string().uri({ allowRelative: true }).trim().messages({
+		"string.empty": "API base url is required.",
+		"string.uri": "The URL you provided is not valid.",
+	}),
+	logLevel: joi.string().valid("debug", "none", "error", "warn").allow(""),
+	systemEmailHost: joi.string().allow(""),
+	systemEmailPort: joi.string().allow(""),
```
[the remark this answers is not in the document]
ajhollid: @jennifer-gan this should indeed be number

Client/src/Validation/validation.js
```diff
@@ -128,10 +128,51 @@ const maintenanceWindowValidation = joi.object({
...
+			return value;
+		}),
+	dbType: joi.string().trim().messages({
+		"string.empty": "DB type is required.",
+	}),
+	redisHost: joi.string().trim().messages({
+		"string.empty": "Redis host is required.",
+	}),
+	redisPort: joi
+		.string()
```
ajhollid, line 160: This should be number as well

ajhollid (changes requested):
Looks good to me, just need those port vaildations to be numbers.  Otherwise validation will be triggered when you try to save settings after loading initial settings:

image

Client/src/Validation/validation.js
jennifer-gan: You seems not having the latest code pulled to your local? the email port and redis port are all optional, to be consistent with BE validation, so if you left empty, there is no error message; (see below screenshot);  if you try to enter anything, only a number can be entered;   I change the label for system email port, see below
Screenshot from 2024-10-25 10-01-24

jennifer-gan: Resolved in : https://github.com/bluewave-labs/bluewave-uptime/pull/1020/commits/2427bdb78804732071a288e4d69c54f43bae804f

ajhollid:
Alright looking good!  Thanks for making changes :smile:

PR #1028, opened 2024-10-22
chore: changing rule on eslint

marcelluscaio:
This PR just corrects a rule that was bringing more trouble than good. 
This is the rule, and we do not need it because the workaround is much easier and readable.
https://kamrutkar.hashnode.dev/unescaped-html-entities-in-jsx-reactno-unescaped-entities

### DIFF
--- a/Server/index.js
+++ b/Server/index.js
@@ -36,7 +36,7 @@ import mjml2html from "mjml";
 
 // Settings Service and dependencies
 import SettingsService from "./service/settingsService.js";
-import AppSettings from "../db/models/AppSettings.js";
+import AppSettings from "./db/models/AppSettings.js";
 
 import db from "./db/mongo/MongoDB.js";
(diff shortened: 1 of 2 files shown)


Server/index.js
```diff
@@ -36,7 +36,7 @@ import mjml2html from "mjml";
 
 // Settings Service and dependencies
 import SettingsService from "./service/settingsService.js";
-import AppSettings from "../db/models/AppSettings.js";
+import AppSettings from "./db/models/AppSettings.js";
```
marcelluscaio, line 39: @ajhollid for some reason the two dots were still here, even though it has already been merged

ajhollid: Wierd, it's in the codebase on `develop` branch now, perhaps this branch was created from develop before the commit was merged?  Anyway no big deal!

ajhollid (approved):

PR #1036, opened 2024-10-23
Add threshold fields to hardware monitor schemas

ajhollid:
This PR adds a threshold field to all hardware monitoring schemas.  This will allow a user to specify at what usage threshold they want to be notified

- [x] Add threshold to CPU, Disk, and Memory schemas


ajhollid:
Thresholds should be in Monitor, not Check

PR #1037, opened 2024-10-23
feat/be/hardware monitor notifications

ajhollid:
This PR implements sending notifcations when thresholds are exceeded for hardware monitors

- [x] Add hardware threshold fields to `Monitor` model
  - [x] Updated BE validation to allow for threshold fields when creating a `Monitor`
- [x] Implement sending email on threshold violation
  - [x] Added an email template for threshold violations
  - [x] Added templates to lookup object in `EmailService` 
  - [x] Added a method to handle hardware updates
  - [x] Refactored the `handleNotifications` method to take a config object instead of multiple parameters
    - [x] `handleNotifications` can now handle arbirtary templates


Server/service/networkService.js
```diff
@@ -110,6 +121,55 @@ class NetworkService {
...
+			context.disk = `DISK USAGE: ${diskUsage * 100}% > ${thresholds.usage_disk * 100}%`;
+		}
+
+		const config = {
+			monitor: monitor,
+			template: "thresholdViolatedTemplate",
+			context: context,
+			subject: `Threshold Violated for ${monitor.name}`,
+		};
+		this.handleNotification(config);
```
marcelluscaio, line 170: Do we need await here?

ajhollid: We don't need to await here as the result of `handleNotification` isn't needed for anything, we just want to fire off the notification and then keep executing code.  If the notification fails for whatever reason we want to keep executing

PR #1059, opened 2024-10-24
Fixed the modal redirect bug

### LINKED ISSUE #1027: FE - Modal dismiss does not prevent clicks outside
**Describe the bug**
When clicking on remove monitor, on the monitor page, a modal opens. If you cancel it by clicking outside of the modal, it navigates to another page.

**To Reproduce**
1. Go to the monitors page
2. Click on the gear icon, and then remove monitor
3. A modal will open
4. Click outside of the modal
5. It should just close the modal, but it navigates to another page

**Expected behavior**
It should just close the modal

navishmonga:
Fixes #1027 

The problem was that the default behaviour of the Material UI Modal Component (used in GenericDialog component) was leading the propogation of the event . So I explicitly defined that on click, the event should not propogate.

### DIFF
--- a/Client/src/Components/Dialog/genericDialog.jsx
+++ b/Client/src/Components/Dialog/genericDialog.jsx
@@ -12,6 +12,7 @@ const GenericDialog = ({ title, description, open, onClose, theme, children }) =
 			aria-describedby={ariaDescribedBy}
 			open={open}
 			onClose={onClose}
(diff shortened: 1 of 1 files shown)


marcelluscaio (approved):
That was a good addition. Thanks for your contribution

ajhollid (approved):

PR #1070, opened 2024-10-25
feat/fe/env var injection

ajhollid:
This PR updates the docker configurations to inject a base API url at runtime.

This is needed for ease of set up for quick start users.

- [x] Update add an env.sh script to replace env vars at run time
- [x] Update the docker dev config to use dev mode
- [x] Add an `env.production` file with target var to be repalced at runtime.  This is OK to commit
- [x] Update client dockerfiles to copy the env.sh script to nginx image
  - [x] Nginx image will run var replacement script on startup
- [x] Update docker-compose files to pass API env var at run time     
- [x] Update quickstart docs


marcelluscaio:
This is to enable quick setup, right?

ajhollid:
> This is to enable quick setup, right?

Yep, this will allow end users to run docker-compose with an env var for their API server location

marcelluscaio (approved):

Issue #1026, opened 2024-10-22
FE  - Light/Dark mode cannot detect user's settings


marcelluscaio:
**Describe the bug**
 Light/Dark mode cannot detect user's settings

**To Reproduce**
Change system and browser settings to Dark, and access the project. You will see that the it defaults to light.

**Expected behavior**
Page should detect user's settings for light/dark mode.

**Additional context**
I believe we could follow a native approach using prefers-color-scheme's media query

gorkem-bwl:
Yep, let's do this one. Setting the label as "good-first-issue" as it's relatively an easy feature for contributors.

aarya-16:
@marcelluscaio I've looked at it and seen what needs to be done. Can I work on this ?

marcelluscaio:
Yes, @aarya-16 ! Thank you for your interest

aarya-16:
@marcelluscaio I have made PR #1072 addressing this issue.

marcelluscaio:
Thanks for your contribution, @aarya-16 . I will check it out

PR #1072, opened 2024-10-25
Added User theme(Light/Dark) preference detection

aarya-16:
-> Used window.matchMedia API to get the user's theme preference to solve issue #1026

### DIFF
--- a/Client/src/App.jsx
+++ b/Client/src/App.jsx
@@ -39,6 +39,14 @@ import { useDispatch } from "react-redux";
 import { getAppSettings, updateAppSettings } from "./Features/Settings/settingsSlice";
 import { logger } from "./Utils/Logger"; // Import the logger
 import { networkService } from "./main";
+import { setMode } from "./Features/UI/uiSlice";
+
+const getPreferredTheme = () => {
+	if(window.matchMedia && window.matchMedia('(prefers-color-scheme: dark)').matches) {
+		return 'dark';
+	}
+	return 'light';
+}
 function App() {
 	const AdminCheckedRegister = withAdminCheck(Register);
 	const MonitorsWithAdminProp = withAdminProp(Monitors);
@@ -57,6 +65,11 @@ function App() {
 			dispatch(getAppSettings({ authToken }));
 		}
 	}, [dispatch, authToken]);
+
+	useEffect(() => {
+		const theme = getPreferredTheme();
+		dispatch(setMode(theme));
+	},[dispatch]);
 
 	// Cleanup
 	useEffect(() => {
--- a/Client/src/App.jsx
+++ b/Client/src/App.jsx
@@ -42,11 +42,14 @@ import { networkService } from "./main";
 import { setMode } from "./Features/UI/uiSlice";
 
 const getPreferredTheme = () => {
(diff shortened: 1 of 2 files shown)


aarya-16:
@ajhollid can you take a look at this PR and check if any changes are necessary ?

ajhollid:
> @ajhollid can you take a look at this PR and check if any changes are necessary ?
> 

Yep, we will take a look at it as soon as we can, it's the weekend so it might be a little while until someone is available to review it. 

Appreciate your patience, you can also request reviews from @marcelluscaio and @jennifer-gan, they are the front end maintainers so this is their domain.

Client/src/App.jsx
```diff
@@ -39,6 +39,17 @@ import { useDispatch } from "react-redux";
...
+import { setMode } from "./Features/UI/uiSlice";
+
+const getPreferredTheme = () => {
+	try {
+		return window?.matchMedia?.('(prefers-color-scheme: dark)')?.matches? 'dark': 'light';
+	}
+	catch {
+		return 'light';
+	}
+}
```
marcelluscaio, line 51: Thank you for your time working on this issue, @aarya-16 !

This is a good path, but doing this in the app directory can bring some performance issues. I suggest you do the check using a ternary inside of the UI slice. so instead of defining the inital theme as light, we will be getting the inital one dynamically at the store level. With that we won't need to call a state update with the use effect.

marcelluscaio (changes requested):

Client/src/App.jsx
aarya-16: I'll implement those changes, @marcelluscaio. I do have one question though:
Do you want me to add the event listeners to handle dynamic user system theme changes in the UI Slice as well ?

marcelluscaio: Good point! I don't think we need the event listener. All we want is to assess user's initial preference. The rest is being dealt with through the settings menu. Do you see a scenario in which that would be necessary?

aarya-16: The only scenario I see is if someone changes their browser or system theme settings while the website is active. But even in that case, a page reload would do the trick. So I guess I'll just make the changes without the event listener.

marcelluscaio: Great! I agree with you, I don't think users expect a live reload according to system settings, and we would get loose of an event listener.

aarya-16:
@marcelluscaio I have moved the theme preference detection to the UISlice file. Please review it.

Client/src/Features/UI/uiSlice.js
```diff
@@ -15,7 +15,7 @@ const initialState = {
 	sidebar: {
 		collapsed: false,
 	},
-	mode: "light",
+	mode: window?.matchMedia?.('(prefers-color-scheme: dark)')?.matches? "dark": "light",
```
marcelluscaio, line 18: Not sure what happened, but the formatting is not that clear without the space between the first part of the ternary and the question mark. Would you mind extracting the ternary from the object to increase legibility? I suggest saving it to a variable with a name like initialMode or something like that

aarya-16: @marcelluscaio I have made that change

ajhollid (approved):
Looks good to me!

PR #1094, opened 2024-10-30
feat/be/refactor-base-check

ajhollid:
This PR refactors the various `Check` models to extend a base check schema as they should all have the same basic fields.

- [x] Refactor basic `Check` schema into a base check schema and export it
  - [x] All `Check` models implement the base schema


jennifer-gan (approved):
Changes look good and I can create new monitor as usual

PR #1125, opened 2024-11-08
Add tests for hardwareCheckModule

ajhollid:
This PR adds tests for the `hardwareCheck` module


ajhollid:
This is a dead module, no need to test

PR #1129, opened 2024-11-08
feat/be/remove-dead-modules

ajhollid:
This PR remoes some unsued DB modules.  These DB modules were used for testing purposes, they are no longer needed

- [x] Remove hardwareCheck db module
- [x] Remove pagespeedCheck db module

These are both actually just instances of `check` and their DB operations are handled by the check module.


ajhollid:
Modules are not fully dead

Issue #1131, opened 2024-11-08
Implement public status page


gorkem-bwl:
A public status page provides information about the current operational status of the servers. Main aim here is to keep end users informed during the incidents.

Here is the current design of the public status page, and the Figma link can be found [here](https://www.figma.com/design/RPSfaw66HjzSwzntKcgDUV/Uptime?node-id=483-21688&t=Osqg88AGlQ1UPFfB-1). Please ensure the settings page design leverages our components wherever possible and maintains a consistent look and feel with other sections of the dashboard.

<img width="901" alt="image" src="https://github.com/user-attachments/assets/9efe5920-237a-4aa5-8edb-b177d8ef60bd">

If all the servers are up at the time when the user views this page, then this bar is shown: 

<img width="1007" alt="image" src="https://github.com/user-attachments/assets/f602d828-666e-40bb-8a9f-99b308d4286c">

If any of the publicly tracked servers are down, this bar is shown: 

<img width="1005" alt="image" src="https://github.com/user-attachments/assets/2919d10d-e9d1-4b06-a391-8bac6bcc42c0">

This feature will have its own sidebar "Status pages", and have a settings to generate a public status page: 

1. **General settings:** Manage visibility of the public status page, configure the company name, and set the sub-URL.

<img width="661" alt="image" src="https://github.com/user-attachments/assets/fc19b943-bb5f-489b-8f25-f87a6221561c">

2. **Server list:** Add the servers visible to the public

<img width="683" alt="image" src="https://github.com/user-attachments/assets/aec3a6bd-364b-45ef-a669-ac96807c92fd">

All the designs required for the public status page settings are [here](https://www.figma.com/design/RPSfaw66HjzSwzntKcgDUV/Uptime?node-id=1-39381&t=Osqg88AGlQ1UPFfB-1).

jennifer-gan:
The creation/configuration of public page, namely status page  were done in below PRs, they simply need to be merged in:
https://github.com/bluewave-labs/Checkmate/pull/1493, https://github.com/bluewave-labs/Checkmate/pull/1494

I am working on the public page itself 

I had a question, if the public page is in this checkmate project, it will need to fetch the status page and all the monitors with 
checks, filter the monitors based on the status page configured servers.  How would we retrieve all the monitors without a login/access token?

gorkem-bwl:
@ajhollid closing this as majority of the work has been completed and we can start testing. I can reopen if you think it should stay open for a while.

[...]

PR #1138, opened 2024-11-11
feat/be/status page implementation

ajhollid:
This PR lays the groundwork for the stauts page described in #1131 

- [x] Add BE routes for stauts pages at `api/v1/status-page`
- [x] Add Controller and methods for status page
  - [x] createStatusPage
  - [x] getStatusPageByUrl
 - [x] Add DB methods for status page
 - [x] Add tests for controller and DB methods
 - [x] Update swagger API docs

### DIFF
--- a/Server/db/models/StatusPage.js
+++ b/Server/db/models/StatusPage.js
@@ -1,0 +1,41 @@
+import mongoose from "mongoose";
+
+const StatusPageSchema = mongoose.Schema(
+	{
+		companyName: {
+			type: String,
+			required: true,
+			default: "",
+		},
+		url: {
+			type: String,
+			required: true,
+			default: "",
+		},
+		timezone: {
+			type: String,
+			required: true,
+			default: "America/Toronto",
+		},
+		color: {
+			type: String,
+			required: true,
+			default: "#4169E1",
+		},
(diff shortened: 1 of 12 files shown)


Server/db/models/StatusPage.js
```diff
@@ -0,0 +1,42 @@
...
+		color: {
+			type: String,
+			required: true,
+			default: "#4169E1",
+		},
+		theme: {
+			type: String,
+			required: true,
+			default: "light",
+		},
```
marcelluscaio, line 30: Do we need to store the theme as well? I thought that was dealt with locally

marcelluscaio (approved):
The code is really well structured. I just had one question, I put it as a comment

Server/db/models/StatusPage.js
ajhollid: @marcelluscaio yes I'm not quite sure of the usage of this one either, I just followed the spec and there was a `theme` field in it :stuck_out_tongue:  It could be that the spec was from an earlier iteration before we had theme defined at all actually.

If it turns out we don't need this we can remove it when we're clear on its usage :+1:   Thanks for being vigilant!

PR #1142, opened 2024-11-11
Update tests for JobQueue and StatusService after chanages to hardwar…

ajhollid:
This PR udpates a few tests for JobQueue and StatusService to reflect changes to the JobQueue and hardware monitor schema

### DIFF
--- a/Server/tests/services/jobQueue.test.js
+++ b/Server/tests/services/jobQueue.test.js
@@ -30,6 +30,14 @@ class QueueStub {
 	async getJobs() {
 		return this.jobs;
 	}
+
+	async pause() {
+		return true;
+	}
+
+	async obliterate() {
+		return true;
+	}
 }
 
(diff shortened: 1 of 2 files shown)


Server/tests/services/jobQueue.test.js
```diff
@@ -30,6 +30,14 @@ class QueueStub {
...
 		return this.jobs;
 	}
+
+	async pause() {
+		return true;
+	}
+
+	async obliterate() {
+		return true;
+	}
```
marcelluscaio, line 40: Out of curiosity: what do these functions do?

ajhollid: > Out of curiosity: what do these functions do?

They are stubs for the actual functions in the JobQueue, we don't actually want to call the methods, we just want to mock them for testing purposes

PR #1149, opened 2024-11-12
1067 fe hardware monitoring create hardware monitor

jennifer-gan:
- [x] Created custom threshold component for threshold alert section 
- [x] Add validation for create infrastructure monitor page
- [x] Add feature slice for infrastructure monitor that include create, get, update and delete monitors
- [x] Add create infrastructure monitor page
- [x] Fixed a css issue in CheckBox component that makes the lable and checkbox vertically off when in a row with other component
- [x] Generate payload to create infrastructure monitor and able to save to BE with success
- [x] Connect Infrastructure create page to global routes and hook up to "create monitor" button on infrastructure home page
- [ ]  system load, cpu temperature and swap usage, log retention period and max tries are to be done , this will need backend support also, will be in another PR
[Screencast from 2024-11-19 09-58-57.webm](https://github.com/user-attachments/assets/12f9e873-3bfa-4f65-b269-37a94641be27)

### DIFF
--- a/Client/src/Components/Inputs/Checkbox/index.jsx
+++ b/Client/src/Components/Inputs/Checkbox/index.jsx
@@ -89,7 +89,7 @@ const Checkbox = ({
 
 Checkbox.propTypes = {
 	id: PropTypes.string.isRequired,
-	label: PropTypes.oneOfType([PropTypes.string, PropTypes.object]).isRequired,
+	label: PropTypes.oneOfType([PropTypes.string, PropTypes.node]).isRequired,
 	size: PropTypes.oneOf(["small", "medium", "large"]),
 	isChecked: PropTypes.bool.isRequired,
 	value: PropTypes.string,


gorkem-bwl:
Can you fix alignments here? 

SCR-20241112-qyyr
SCR-20241112-qyzi

Client/src/Utils/stringUtils.js
```diff
@@ -0,0 +1,3 @@
+export const capitalizeFirstLetter = (str) =>{
+  return  str.charAt(0).toUpperCase() + str.slice(1);
+}
```
ajhollid, line 3: I suggest adding some safetry checks to this utility function.  If we've gone through the trouble to extract it we should make it safe to use.

On visual inspection It will throw errors right now if given the wrong type, empty string, `null`, or `undefined`.  Could be other edge cases too, I would explore and make this robust.

Client/src/Components/Inputs/Field/index.css
```diff
@@ -1,6 +1,15 @@
...
 	min-width: 250px;
 }
+
+.field-infrastructure-alert {
+	max-width: 100px;
+}
+
+.field-infrastructure-alert .MuiInputBase-root:has(input) {
+	height: 34px;
+}
```
ajhollid, line 11: I'm wary of hardcoding width and height into a component, I would prefer these be optional props with default values :thinking:

Client/src/Components/Inputs/Checkbox/index.jsx
```diff
@@ -65,7 +65,7 @@ const Checkbox = ({
 			sx={{
 				borderRadius: theme.shape.borderRadius,
 				p: theme.spacing(2.5),
-				m: theme.spacing(-2.5),
+				//m: theme.spacing(-2.5),
```
ajhollid, line 68: Is this commented out temporarily?

If it's a permanent change then it should be removed not just commented out.

Client/src/Components/CustomThreshold/index.jsx
```diff
@@ -0,0 +1,72 @@
+import { Box, Stack, Typography } from "@mui/material";
+import Field from "../Inputs/Field";
+import Checkbox from "../Inputs/Checkbox";
+import { useTheme } from "@emotion/react";
+import PropTypes from "prop-types";
+
+export const CustomThreshold = ({
```
ajhollid, line 7: Be sure to include JSdoc for all functions/components before opening the final pull request :+1:

ajhollid (commented):
I don't see any major issues here, looks pretty good!

Minor issues:

1.  I would make the util string funciton more robust
2.  We need JSdoc for all functions/components.

Great work!

Client/src/Components/Inputs/Checkbox/index.jsx
jennifer-gan: Yes, I will remove that since it is a fix

Client/src/Components/Inputs/Field/index.css
jennifer-gan: The class name is a prop itself, I will see if I can define  the height value as css variable if that makes more sense

Client/src/Utils/stringUtils.js
jennifer-gan: will do and some JSdoc as well

jennifer-gan:
> Can you fix alignments here?
> 
> SCR-20241112-qyyr SCR-20241112-qyzi

I will see what I can do with the notification checkbox label and child element alignment; regarding the threshold part, it will be aligned when it is full size, when you shrink the screen, due to its responsive, you will see they will not be left align 
after a certain size, but they will be right aligned always.

gorkem-bwl:
> I will see what I can do with the notification checkbox label and child element alignment; regarding the threshold part, it will be aligned when it is full size, when you shrink the screen, due to its responsive, you will see they will not be left align after a certain size, but they will be right aligned always.

Makes sense!

Client/src/Components/Inputs/Field/index.css
marcelluscaio: It is better if you set it as min-height. Fixed heights are prone to overflow

marcelluscaio:
It would be nice if customize alerts inputs would be disabled when the checkbox is not checked. For the responsiveness alignment on the left, using wrap could help with that, so all inputs would go to the next line, avoiding overflow

Client/src/Validation/error.js
```diff
@@ -22,15 +22,29 @@ const hasValidationErrors = (form, validation, setErrors) => {
...
+			// Handle conditionally usage number required cases
+			if (!form.cpu || (form.cpu && form.usage_cpu)) {
+				delete newErrors["usage_cpu"];
+			}
+			if (!form.memory || (form.memory && form.usage_memory)) {
+				delete newErrors["usage_memory"];
+			}
+			if (!form.disk || (form.disk && form.usage_disk)) {
+				delete newErrors["usage_disk"];
+			}
```
marcelluscaio, line 39: What are these checks? Not sure why the condition to deleting those keys

ajhollid: Side note as well in general it's best to set the value of the key to undefined or null rather than deleting the key value pair.

Client/src/Components/Inputs/Field/index.jsx
```diff
@@ -56,7 +57,7 @@ const Field = forwardRef(
 		return (
 			<Stack
 				gap={theme.spacing(2)}
-				className={`field field-${type}`}
+				className={`${className ?? `field field-${type}`}`}
```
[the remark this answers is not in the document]
jennifer-gan: we need the extra className not only to override, but also to hide the error , since we have changed the design to display single error msg for all the inputs errors

Client/src/Validation/error.js
jennifer-gan: these checks are for the conditional error setting for the threshold inputs, the inputs are be mandatory when the checkbox are checked;

jennifer-gan: > Side note as well in general it's best to set the value of the key to undefined or null rather than deleting the key value pair.

yes, Have changed to set them as null instead

jennifer-gan: > What are these checks? Not sure why the condition to deleting those keys

these checks are for the conditional error setting for the threshold inputs, the inputs are mandatory when the checkbox are checked;

Client/src/Components/Inputs/Field/index.css
jennifer-gan: > It is better if you set it as min-height. Fixed heights are prone to overflow

To be consistent with normal input cases, I have used variable to set the height now

Client/src/Utils/stringUtils.js
jennifer-gan: resolved in : https://github.com/bluewave-labs/bluewave-uptime/pull/1149/commits/a9aece6123eaa20ce2173c62c7f7904379a9f0dd
https://github.com/bluewave-labs/bluewave-uptime/pull/1149/commits/4e60981b0b1c487182b830991c89c117f6670006 etc

Client/src/Components/Inputs/Checkbox/index.jsx
jennifer-gan: resolved in : https://github.com/bluewave-labs/bluewave-uptime/pull/1149/commits/a9aece6123eaa20ce2173c62c7f7904379a9f0dd
https://github.com/bluewave-labs/bluewave-uptime/pull/1149/commits/4e60981b0b1c487182b830991c89c117f6670006 etc

jennifer-gan:
> > I will see what I can do with the notification checkbox label and child element alignment; regarding the threshold part, it will be aligned when it is full size, when you shrink the screen, due to its responsive, you will see they will not be left align after a certain size, but they will be right aligned always.
> 
> Makes sense!



> > I will see what I can do with the notification checkbox label and child element alignment; regarding the threshold part, it will be aligned when it is full size, when you shrink the screen, due to its responsive, you will see they will not be left align after a certain size, but they will be right aligned always.
> 
> Makes sense!

resolved in : https://github.com/bluewave-labs/bluewave-uptime/pull/1149/commits/a9aece6123eaa20ce2173c62c7f7904379a9f0dd
https://github.com/bluewave-labs/bluewave-uptime/pull/1149/commits/4e60981b0b1c487182b830991c89c117f6670006 etc

Client/src/Validation/error.js
jennifer-gan: resolved in : https://github.com/bluewave-labs/bluewave-uptime/pull/1149/commits/a9aece6123eaa20ce2173c62c7f7904379a9f0dd
https://github.com/bluewave-labs/bluewave-uptime/pull/1149/commits/4e60981b0b1c487182b830991c89c117f6670006 etc

jennifer-gan:
I believe I addressed all the points raised from all 3 people reviewed, thanks!

Client/src/Pages/InfrastructureMonitors/CreateMonitor/index.jsx
```diff
@@ -0,0 +1,432 @@
...
+		url: "",
+		name: "",
+		notifications: [],
+		interval: 15,
+		cpu: false,
+		usage_cpu: "",
+		memory: false,
+		usage_memory: "",
+		disk: false,
+		usage_disk: "",
```
ajhollid, line 34: The values for usage_cpu, usage_memory, and usage_disk should all be numbers, this will cause an error if the form is submitted with default values as validation type is number

Client/src/Components/CustomThreshold/index.jsx
```diff
@@ -0,0 +1,92 @@
+import { Box, Stack, Typography } from "@mui/material";
+import Field from "../Inputs/Field";
+import Checkbox from "../Inputs/Checkbox";
+import { useTheme } from "@emotion/react";
+import PropTypes from "prop-types";
+
```
ajhollid, line 6: This is not a reusable component, you can safely move it to the `createInfrastrucutreMonitor` directory.  the `Components` directory should only hold reusable components

ajhollid (changes requested):
Nice work here, the UI looks really nice :+1: 

There's a few issues I noticed while running the code, I've noted them in the comments.

Aside from that there are some organizational changes I think we can make.  Anything not reusable should be moved to the page specific directory, only reusable code/components should be refactored out to util files or the `Component` directory.  

If component specific code is added to util files it just makes it harder to track down where logic is being executed.

It seems like something from @marcelluscaio's recent them refactor has affected the showing of errors, so we'll have to track that one down.  This isn't specific to your so we can double check the error handling once we've got that one sorted out!

Client/src/Pages/InfrastructureMonitors/CreateMonitor/index.jsx
```diff
@@ -0,0 +1,432 @@
...
+	};
+	const handleCreateInfrastructureMonitor = async (event) => {
+		event.preventDefault();
+		let form = {
+			...infrastructureMonitor,
+			name:
+				infrastructureMonitor.name === ""
+					? infrastructureMonitor.url
+					: infrastructureMonitor.name,
+			interval: infrastructureMonitor.interval * MS_PER_MINUTE,
```
ajhollid, line 159: This conversion isn't correct for any values in seconds. If you select "15s" you get 900000ms, which is 15 minutes.  Since the constant is MS_PER_MINUTE, seconds should first be converted to minutes before multiplying by MS_PER_MINUTE.  ie 15 seconds is 0.25 minutes

Client/src/Components/Inputs/Checkbox/index.jsx
```diff
@@ -39,6 +39,7 @@ const Checkbox = ({
 }) => {
 	const sizes = { small: "14px", medium: "16px", large: "18px" };
 	const theme = useTheme();
+	const checkBoxPosition = typeof label === "string" ? {} : { pb: "65px" };
```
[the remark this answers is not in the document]
jennifer-gan: This is to address the alignment issue @gorkem-bwl brought up, only when label is not a string, it will need to be a bottom padding of 65px, also not sure if this should be move to any theme , or even make it into its own css file? as it is not worth to have a separate class just for this one css property ?

Client/src/Components/Inputs/Checkbox/index.jsx
```diff
@@ -85,7 +86,7 @@ const Checkbox = ({
 
 Checkbox.propTypes = {
 	id: PropTypes.string.isRequired,
-	label: PropTypes.string.isRequired,
+	label: PropTypes.oneOfType([PropTypes.string, PropTypes.object]).isRequired,
```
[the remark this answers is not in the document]
jennifer-gan: Pass in the element as a label is to solve the alignment issue @gorkem-bwl brought up,  or we can revert back to have the child element not aligned with the parent label, please let me know.

Client/src/index.css
```diff
@@ -20,6 +20,8 @@ html {
 
 	--env-var-width-1: 100vw;
 	--env-var-width-2: 360px;
+	--env-var-width-3: 250px;
+	--env-var-width-4: 100px;
```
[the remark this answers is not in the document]
jennifer-gan: I can move it to its own css file, not sure what was done to sort out the global theme thing regarding how we can address all pixel accurate layouts with the rem to be responsible etc.

Client/src/Validation/error.js
```diff
@@ -1,7 +1,7 @@
 const buildErrors = (prev, id, error) => {
 	const updatedErrors = { ...prev };
 	if (error) {
-		updatedErrors[id] = error.details[0].message?? "Validation error";
+		updatedErrors[id] = error.details[0].message ?? "Validation error";
 	} else {
 		delete updatedErrors[id];
 	}
```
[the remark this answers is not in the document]
jennifer-gan: hasValidationErrors was refactored out when i worked on advanced settings page, it is not specific to any pages,
as it overlooking all pages for validation errors including white list unhandled fields to avoid unintended error being set such as below

```	if (
				![
					"clientHost",
					"refreshTokenSecret",
					"dbConnectionString",
					"refreshTokenTTL",
					"jwtTTL",
					"notify-email-list",
				].includes(err.path[0])
			) {
```

Client/src/Components/CustomThreshold/index.jsx
jennifer-gan: sure, will do

Client/src/Components/Inputs/Checkbox/index.jsx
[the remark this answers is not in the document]
ajhollid: I'm sure we can find a way to align things properly without compromising the integrity of the component 🤔

Can you revert it to only accepting a string for label and we can brainstorm how to sort out alignment? Thanks!

Client/src/Validation/error.js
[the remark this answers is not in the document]
ajhollid: Thanks for the explanation, I'll have to take a closer look at this function and how it is used.

Client/src/Components/Inputs/Checkbox/index.jsx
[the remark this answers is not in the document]
jennifer-gan: The label can be a string or element, that is normal in a checkbox case, this is the definition on it
from MUI FormControlLabelProps.d.ts. that is why I extended the component to accept element
```
  * A text or an element to be used in an enclosing label element.
   */
  label: React.ReactNode;
```

Client/src/Utils/stringUtils.js
jennifer-gan: yes, will update

Client/src/Pages/InfrastructureMonitors/CreateMonitor/index.jsx
jennifer-gan: that has to be a string, since the component has define it as string
```
Field.propTypes = {
	type: PropTypes.oneOf(["text", "password", "url", "email", "description", "number"]),
	id: PropTypes.string.isRequired,
	name: PropTypes.string,
	label: PropTypes.string,
	https: PropTypes.bool,
	isRequired: PropTypes.bool,
	isOptional: PropTypes.bool,
	optionalLabel: PropTypes.string,
	autoComplete: PropTypes.string,
	placeholder: PropTypes.string,
	value: PropTypes.string.isRequired,
	onChange: PropTypes.func,
	onBlur: PropTypes.func,
	onInput: PropTypes.func,
	error: PropTypes.string,
	disabled: PropTypes.bool,
	hidden: PropTypes.bool,
	className: PropTypes.string
};
```
BE seems work fine with or without those values as part of payload

Client/src/Pages/InfrastructureMonitors/CreateMonitor/index.jsx
jennifer-gan: updated, the frequency already in seconds, so I need to change to MS_PER_SECOND, done.
```
	//select values
	const frequencies = [
		{ _id: 15, name: "15 seconds" },
		{ _id: 30, name: "30 seconds" },
		{ _id: 60, name: "1 minute" },
		{ _id: 120, name: "2 minutes" },
		{ _id: 300, name: "5 minutes" },
		{ _id: 600, name: "10 minutes" },
	];
```

Client/src/Validation/validation.js
```diff
@@ -173,11 +176,51 @@ const advancedSettingsValidation = joi.object({
 	pagespeedApiKey: joi.string().allow(""),
 });
 
+const infrastractureMonitorValidation = joi.object({
```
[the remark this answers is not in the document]
jennifer-gan: fixed

Client/src/Pages/InfrastructureMonitors/CreateMonitor/index.jsx
ajhollid: @jennifer-gan we use milliseconds per m in minute everywhere else in the app, can you convert the seconds to minutes instead of changing the conversion factor so that we stay consistent?

Otherwise someone is going to be confused at some point and convert the wrong values. Thanks!

Client/src/Components/Inputs/Checkbox/index.jsx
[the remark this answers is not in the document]
ajhollid: Ah I see! 

Can you update the prop type to specify a react component and not an object in that case? It's not clear it's meant to take a react component.

Client/src/Components/Inputs/Checkbox/index.jsx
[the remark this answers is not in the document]
ajhollid: If you need additional padding for a specific case like the create infrastructure monitor page it can be added with inline styles.

I think that's a better solution than hardcoding this in. This will affect all checkboxes that use strings as labels which I don't think we want.

Client/src/Pages/InfrastructureMonitors/CreateMonitor/index.jsx
ajhollid: You can convert the value to a string before passing it to whatever component requires it to be a string 🤔

This state seems misleading as the value really should be a number.

Yes the backend will accept a request with or without these values as they are optional.

Client/src/Components/Inputs/Field/index.css
marcelluscaio: Sorry, I don't understand, Jennifer. Why using height instead of min-height?
Using a fixed height can lead to overflow, like in this example
https://developer.mozilla.org/en-US/docs/Learn/CSS/Building_blocks/Overflowing_content#css_tries_to_avoid_data_loss

Client/src/Components/Inputs/Checkbox/index.jsx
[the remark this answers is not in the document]
marcelluscaio: I agree, the 65px seem hardcoded, we could consider a different approach. Can you bring examples for this in this thread? I don't understand the link with Gorkem's print

Client/src/Components/Inputs/Checkbox/index.jsx
```diff
@@ -66,11 +67,11 @@ const Checkbox = ({
 			sx={{
 				borderRadius: theme.shape.borderRadius,
 				p: theme.spacing(2.5),
```
marcelluscaio, line 69: I think we are not supposed to use floats with MUI's spacing
https://mui.com/material-ui/customization/spacing/

Client/src/Components/Inputs/Field/index.jsx
```diff
@@ -185,10 +186,11 @@ const Field = forwardRef(
 						),
 					}}
 				/>
-				{error && (
+				{ error && (
 					<Typography
 						component="span"
 						className="input-error"
+						hidden={className? true: false}
```
marcelluscaio, line 193: Why do we have error message hidden or not based on having a className?

Client/src/Utils/stringUtils.js
```diff
@@ -0,0 +1,17 @@
...
+		return "";
+	}
+	if (typeof str !== "string") {
+		throw new TypeError("Input must be a string");
+	}
+	if (str.length === 0) {
+		return "";
+	}  
+	return str.charAt(0).toUpperCase() + str.slice(1);
+};
```
marcelluscaio, line 17: What if we get a string in all caps?

Client/src/Components/Inputs/Checkbox/index.jsx
```diff
@@ -66,11 +67,11 @@ const Checkbox = ({
 			sx={{
```
marcelluscaio, line 67: To solve the padding rpoblem, delete the padding bottom and add alignSelf: "flex-start" to the sx of the MuiCheckbox (before &:hover, on line 59 or 60

Client/src/Components/Inputs/Checkbox/index.jsx
jennifer-gan: like mentioned in the chat, the left alignment issue of the notification email text and its previous child elements were already resolved, i'm not sure what we are trying to solve here

Client/src/Components/Inputs/Field/index.jsx
marcelluscaio: Sorry, that is not clear for me, can you explain?

Client/src/Components/Inputs/Checkbox/index.jsx
marcelluscaio: This is something else. The point here is we are using spacing to have a rule for the whole project. If we pass floats here, we are bypassing that rule (the idea is that all spacing will be in a certain scale, in our case that is 2. When you pass .5, we are increasing space by 1)

jennifer-gan: not sure why it was 2.5 historically, what we should be doing with this now or there will be another PR to address all floating spacing?

Client/src/Utils/stringUtils.js
jennifer-gan: caller's call

Client/src/Components/Inputs/Field/index.jsx
jennifer-gan: If you use for threshold infrastructure scenario, then you need that class name and hide the error for each field

marcelluscaio: Classname is not specified in jsdocs. The way it is, as I understand this, if we instantiate this compoennt passing a className to it, to extend styling maybe(?) we will hide the errors.

ajhollid (approved):
I think this is in a decent enough state to merge in.

There are a few issues here and there but I think we would be better off merging this in now and addressing those fixes in a separate PR.

PR #1151, opened 2024-11-13
Feat/fe/docker monitors, resolves #1139

### LINKED ISSUE #1139: [FE & BE]  Docker container monitoring
Docker container monitoring is a useful feature that has been requested.  On our end this would require

**FE**
- [ ] Updading the create uptime monitor flow to allow for a third type alongside http and ping.  This type could be called `docker`
  - [ ] There should be a dropdown for connection type when `docker` type is selected
    - [ ] tcp/http
    - [ ] unix socket
  - [ ] All other fields should be the same as other uptime monitors

**BE**
- [x] [dockerode](https://www.npmjs.com/package/dockerode/v/2.5.5) seems to be a good package to handle connection to Docker daemon
- [ ] If user is using tls they must provide certificates and a certificate authority.  The path to these certs and authority should be placed in a default directory, with optional configuration by env var or settings from SettingsService
- [x] The monitor should follow the typical JobQueue / StatusSerivce flow as the other uptime monitors
  - [x]  Add a new case to handle processing of results of a `docker` response.
  
**Documentation**

This feature requires some significant configuration on the user's end and is not suitable for non-technical users.  Clear documentation is essential for user's to make use of this functionality

- [ ] It requires exposing the docker daemon over tcp, or mounting the daemon's socket to the uptime-monitor docker image.
- [ ] If the user wants to expose the daemon over tcp/http and wants to use TLS, they must generate a signing authority and SSL certificates.  This is not difficult, but it requires explanation for those unfamiliar with the process.

ajhollid:
This PR implements the FE component of #1139, builds off #1150 

- [x] Update create monitor UI to include an option for Docker monitors
  - [x] Refactor labels and placeholders to use a map 
- [x] Update uptime monitor slice to handle docker types
- [x] Update BE to process requests for docker monitors
  - [x] Update joi validation
  - [x] Update DB methods  
- [ ] Update BE tests - Will do this in a separate PR

### DIFF
--- a/Client/src/Features/UptimeMonitors/uptimeMonitorsSlice.js
+++ b/Client/src/Features/UptimeMonitors/uptimeMonitorsSlice.js
@@ -40,7 +40,7 @@ export const checkEndpointResolution = createAsyncThunk(
 			const res = await networkService.checkEndpointResolution({
 				authToken: authToken,
 				monitorURL: monitorURL,
-			})
+			});
 			return res.data;
 		} catch (error) {
 			if (error.response && error.response.data) {
@@ -53,7 +53,7 @@ export const checkEndpointResolution = createAsyncThunk(
 			return thunkApi.rejectWithValue(payload);
 		}
 	}
-)
+);
 
 export const getUptimeMonitorById = createAsyncThunk(
 	"monitors/getMonitorById",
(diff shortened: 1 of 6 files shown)


marcelluscaio (approved):
Nice work, Alex!  All looked good to me

PR #1162, opened 2024-11-15
fix: Remove top level package-lock.json

mertssmnoglu:
Remove `/package-lock.json` that comes with pr-987. It not breaks anything because it's top level.


ajhollid (approved):
Good catch, thanks!

PR #1167, opened 2024-11-19
feat/fe/infrastructure details, resolves #1068

### LINKED ISSUE #1068: FE - Hardware Monitoring - Hardware Monitor Details page
There should be a details page for hardware monitors that is opened when a monitor in the homepage is clicked on, similar to Uptime and Pagespeed monitors.

It should

- [ ] Dispaly a row of summary stats: CPU, RAM, OS, Max Frequency and Current Frequencey
- [ ] Display a row of gauges for Memory, CPU, and Disk.  This should be the same gauge developed in #1063 
- [ ] Display line graphs for Memory, CPU, and Disk usage

Screenshot from 2024-10-25 10-27-30

ajhollid:
This PR is a rough implementation of the Infrasctrucutre Monitor page.

I am by no means a front end expert so FE team _please_ fix my mistakes :joy: I've tagged things I'm not sure about with `FE team HELP!`, pleas `ctrl-f` through the file and see if you can make things better.

Specific things I could use help with:

1.  Typography -  I don't know much about fontweights/sizes, if anyone can fix my typography I'd apprecaite it
2.  Colors - I'm not sure if I should define new colors for the charts in the theme, or use the ones already defined?  Currently I used success/warning colors, and I don't think that's a great idea.
3. Spacing - I tried to space things conistently with the Uptime Monitors page, please correct me if it can be done better.
4. Height of the AreaChart.  I've wrapped the AreaChart in rechart's responsive container and it is a child of a flexbox, but if it has any sibling components it overflows its parent if the height of the responisve container is set to 100%.  Not sure why that is, could use some help digging into this.

I also had to include the `AreaChart` and the SVG based Gauge into this PR because I needed to use them.  We can either review them here with this PR or in their own PRs, either is fine.

- [x] Implement Infracstructure Detail page
  - [x] Currently uses dummy data 
  - [x] Dummy data fetched from json file with axios to simulate network request
  - [x] Data stored in state as will be the case when we make a BE request
- [x] Add route for infractructure details at `/infrastrucructure/:monitorId`.  You can use any ID for `:monitorId` for now
- [x] Add AreaChart and SVG based Gauge component  
- [x] Monitor status/information added to top of page

Screenshot from 2024-11-20 12-42-29
Screenshot from 2024-11-20 12-42-43


jennifer-gan (approved):
Great work, looks pretty good!  I'm able to show the detail page with dummy data,  I think it's ready to be merged , one question :  I see tooltips like attached with very long float percentage, is this necessary?
[Screencast from 2024-11-19 15-26-19.webm](https://github.com/user-attachments/assets/34f60963-f0fe-4eda-a4d4-bb5247b3d3bb)

Client/src/Components/Charts/AreaChart/index.jsx
```diff
@@ -0,0 +1,154 @@
...
+	gradientStartColor,
+	gradientEndColor,
+	customTooltip,
+}) => {
+	const theme = useTheme();
+	const gradientId = `gradient-${Math.random().toString(36).slice(2, 9)}`;
+	return (
+		<ResponsiveContainer
+			width="100%"
+			height="80%" // FE team HELP!  Why does this overflow if set to 100%?
```
marcelluscaio, line 92: What sort of layout do you want, Alex?
Right now they are wrapping to the next line, do you want them all side by side like this?
Screenshot 2024-11-19 160305

ajhollid: According to the figma design they should wrap down, looks like there should be a maximum of two graphs per line

marcelluscaio (approved):

PR #1169, opened 2024-11-20
fix: rewrite BE endpoint resolution method

ajhollid:
This PR rewrites the endpoint resolution functionality.  The original implementation was insufficient and only handle DNS resolution, which is only a subset of resolution we need to check.

- [x] use axios to check for endpoint resolution
  - [x] accepts all status code as resolution, we just want to know if the server is there or not and responding
  - [x] add statusCode and statusText to success response in case we need this later
- [x] update tests


jennifer-gan (approved):
looks good

PR #1172, opened 2024-11-20
Feat/fe/infrastructure details connect to be

jennifer-gan:
- [x] Connect to BE replacing dummy data endpoint


Client/src/Components/Inputs/Field/index.jsx
```diff
@@ -57,7 +57,7 @@ const Field = forwardRef(
 		return (
 			<Stack
 				gap={theme.spacing(2)}
-				className={`${className ?? `field field-${type}`}`}
+				className={`field field-${type} ${className}`}
```
ajhollid, line 60: @jennifer-gan I'm really glad to see you picked up on this issue too!  I just submitted a PR that does basically this, could you remove this from this PR so that this one is focused on the details page please?  One feature or fix per PR :smile:

Client/src/Components/Inputs/Field/index.jsx
```diff
@@ -186,11 +186,11 @@ const Field = forwardRef(
 						),
 					}}
 				/>
-				{ error && (
+				{error && (
 					<Typography
 						component="span"
 						className="input-error"
-						hidden={className? true: false}						
+						hidden={className ? true : false}
```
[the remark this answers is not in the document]
ajhollid: @jennifer-gan I addressed this in my PR as well :+1:

Client/src/Pages/Infrastructure/Details/index.jsx
```diff
@@ -183,11 +186,14 @@ const InfrastructureDetails = () => {
...
-						"Cache-Control": "no-cache",
-					},
+				const response = await networkService.getStatsByMonitorId({
+					authToken: authToken,
+					monitorId: monitorId,
+					sortOrder: null,
+					limit: null,
+					dateRange: dateRange,
+					numToDisplay: 50,
+					normalize: true,
```
ajhollid, line 196: We don't need to normalize results here, normalizing the data is used when we are displaying response time in a chart.  Response time has an unbounded range so we need to normalize the data to have a reasonable visual representation.

Since we're not diplaying response time here this can safely be set to false.

ajhollid (changes requested):
Looks good!

We can set normalize parameter to false as we don't need to normalize data here.

Other than that please remove changes to the Field component from this PR, we should try to keep PRs target to one feature/fix per PR.

I've opened up a PR that addresses those issues too, glad you picked up on them though :rocket:  Please see #1173 and see what you think about the approach.  Nice work!

ajhollid (approved):

PR #1173, opened 2024-11-21
Fix/fe/field refactor

ajhollid:
This PR addresses a concern @marcelluscaio and I noticed in the Create Infrastrucutre Monitors page.

The original implementation used the presence or abscence of a class name in the props of the `Field` component to enable or disable the default classes for the component.  

The original implementaiton also used the presence or absecnce of a class name to disable showing errors for the `Field`.

I feel that this is dangerous and will lead to confusion if other devs want to specify class names for the `Field` component.  It is not obvious that specifying a class name will disable the default styling for the component.  Nor is it obvious that adding a class name will disable errors from showing.

I have modified the implementaitont to achieve the inteded results in a way that is explicit and obvious, which should reduce confusion and unintended outcomes.

- [x] Add the class name specified in props to the list of existing classnames rather than disabling all class names
  - [x] Override the default styling to specify the min-width of the input
- [x] Add a `hideErrorText` prop to specify that errors should be hidden
- [x] Update JSdocs

This is a bandaid, and we should make refactoring the `Field` component a fairly high priority.  It is not up to standard and there are multiple CSS files in the app modifying the component which is a nightmare for maintainability.

I will open an issue and work on this as soon as possible.


Client/src/Components/Inputs/Field/index.jsx
jennifer-gan, line 1: I think it would be good to have field-infrastructure-alert inherit everything from field like below
.field .field-infrastructure-alert {
	min-width: unset;
	max-width: var(--env-var-width-4);
}

then in index.jsx , change to className={`field field-${type} ${className}`}
so that only override in case of infrastructure monitoring when className has the value, but still have all the css the .field possess

ajhollid: @jennifer-gan it already inerits everyting from the index.css file, adding the `.field` selector only increases the specificity :thinking: That's not a bad thing here but I don't think it adds anything.  Or were you referring to setting `min-width` to unset vs 0?  Setting it to `unset` is probably a better idea, will update.

I'll update the order of classnames as suggested above though, that's a good point :+1:

ajhollid:
I'm going to merge this as it is essentially identical to @jennifer-gan 's previous PR fixing this issue which I've already merged.

The only addition here is a prop to hide the error text.  This will be refactored soon with the Field refactor anyways.

PR #1174, opened 2024-11-21
Feat/fe/refactor/input

ajhollid:
The `Field` component was a mess and causing headaches for everyone.

This PR rewrites the component in a much more rational way.  You can easily manage the width of the field by using the `maxWidth` prop.  More sizing props can be added as needed.

If you don't want to show the error text simply omit the "helperText` prop.

Please see `<root>/test` URL for a test page.

I have tried to replicate the original `Field` component as closely as possible, please let me know if there is anything I've missed.

This was a _very_ trying tasks as there were several methods of styling applied to this component as well as the theme.  

Appreciate everyone's reviews!


jennifer-gan (commented):
test page looks good to me

jennifer-gan (approved):

Issue #1178, opened 2024-11-22
Color fixes for graphs


gorkem-bwl:
The "down" colors for graphs are a bit off after the last update. 

Here are the screenshots for the upcoming release: 

![image](https://github.com/user-attachments/assets/843f8c16-2449-4a4c-9b67-5b15cfbcefbf)
![image](https://github.com/user-attachments/assets/612144a4-927a-4d59-96c1-2a4bfc64def2)

PR #1220, opened 2024-11-27
fix: adjusted theme settings for color fixes, references #1178

ajhollid:
This PR implements some color fixes after refactoring theme

- [x] `theme.palette.error.contrastText` -> `theme.palette.error.main` for bar chart and status label
  - Not sure if this is the correct approach, but the colors appear to be correct.
  
  
image

 
- [x]  Added a `border.light.disabled` and `border.dark.disabled` property to the palette
  - [x] Use `border.light.disabled` and `border.dark.disabled` for `action.disabled`. 
  - Again I'm not sure if this is the correct approach, this will apply a color meant for borders to all disabled states?  

image

@marcelluscaio you know the most about the theme structure so please let me know if there is a better way to resovle these issues!

### DIFF
--- a/Client/src/Components/Label/index.jsx
+++ b/Client/src/Components/Label/index.jsx
@@ -127,11 +127,11 @@ const StatusLabel = ({ status, text, customStyles }) => {
 	const colors = {
 		up: {
 			dotColor: theme.palette.success.main,
-			bgColor: theme.palette.success./* dark */ contrastText,
+			bgColor: theme.palette.success.contrastText /* dark */,
 			borderColor: theme.palette.success.main /* light */,
 		},
 		down: {
-			dotColor: theme.palette.error.contrastText,
+			dotColor: theme.palette.error.main,
 			bgColor: theme.palette.error.dark,
 			borderColor: theme.palette.error.light,
 		},
--- a/Client/src/Utils/Theme/darkTheme.js
+++ b/Client/src/Utils/Theme/darkTheme.js
@@ -23,6 +23,9 @@ const {
 } = colors;
 
 const palette = {
+	action: {
+		disabled: border.dark.disabled,
+	},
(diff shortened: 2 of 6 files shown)


Client/src/Components/Label/index.jsx
```diff
@@ -127,11 +127,11 @@ const StatusLabel = ({ status, text, customStyles }) => {
 	const colors = {
 		up: {
 			dotColor: theme.palette.success.main,
-			bgColor: theme.palette.success./* dark */ contrastText,
+			bgColor: theme.palette.success.contrastText /* dark */,
 			borderColor: theme.palette.success.main /* light */,
```
marcelluscaio, line 131: I had left the light and dark commented just so we know what was there before, I think we can delete these comments (not necessarily on this pr)

Client/src/Utils/Theme/darkTheme.js
```diff
@@ -23,6 +23,9 @@ const {
 } = colors;
 
 const palette = {
+	action: {
+		disabled: border.dark.disabled,
+	},
```
marcelluscaio, line 28: Is this action.disabled color being called anywhere? I didnt find it

marcelluscaio (commented):
Great, Alex! Totally agree with your fixes. 
I just had one question regarding the disabled entry you added

Client/src/Utils/Theme/darkTheme.js
ajhollid: > Is this action.disabled color being called anywhere? I didnt find it

Possible I was playing around and forgot to remove it 🤔 I'll double check it as soon as I get a chance!

ajhollid: Just had a look, this is required for the disabled borders for inputs

PR #1248, opened 2024-12-02
- Update Falback component to accept ovelStart and enlarge the maxWid…

jennifer-gan:
- [x] Update Falback component to accept ovelStart and enlarge the maxWidth to fit into one line
- [x] Add Skeleton
- [x] Add an empty page when there is no monitors
Screenshot from 2024-12-02 10-09-07

### DIFF
--- a/Client/src/Components/Fallback/index.jsx
+++ b/Client/src/Components/Fallback/index.jsx
@@ -20,7 +20,7 @@ import "./index.css";
  * @returns {JSX.Element} The rendered fallback UI.
  */
 
-const Fallback = ({ title, checks, link = "/", isAdmin }) => {
+const Fallback = ({ title, checks, link = "/", isAdmin, ovalStart }) => {
 	const theme = useTheme();
 	const navigate = useNavigate();
 	const mode = useSelector((state) => state.ui.mode);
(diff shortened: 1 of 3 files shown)


gorkem-bwl:
Looks good! Thanks.

marcelluscaio (approved):

Client/src/Components/Fallback/index.jsx
```diff
@@ -20,7 +20,7 @@ import "./index.css";
  * @returns {JSX.Element} The rendered fallback UI.
  */
 
-const Fallback = ({ title, checks, link = "/", isAdmin }) => {
+const Fallback = ({ title, checks, link = "/", isAdmin, ovalStart }) => {
```
marcelluscaio, line 23: Hey Shemy, why is the name of this variable ovalStart?

marcelluscaio: We could provid a fallback for that variable, since this component is already being called. You can set ovalStart  = false in the destructuring (I suggest changing for a more semantic name for the variable)

marcelluscaio (commented):
Great work. I made a small suggestion and asked a question

Client/src/Components/Fallback/index.jsx
ajhollid: @shemy-gan was this meant to be `vowelStart`?  That makes sense?  In the interest of expediencey I'll push that change to the branch, we can revisit if there was a special meaning for `ovalStart`

ajhollid (approved):
Looks good, thanks for your hard work here!

ajhollid (commented):
Just a note as well, in future please do update JSdocs and PropTypes when updating components :+1:

PR #1271, opened 2024-12-04
fix: disabled input

marcelluscaio:
Solves https://github.com/bluewave-labs/checkmate/issues/1257

Screenshot 2024-12-03 193427
Screenshot 2024-12-03 193403


ajhollid (approved):
Looks good!

Issue #1275, opened 2024-12-04
Current Password Field BG color


ajhollid:
While using dark theme, the "Current Password" field on the Password page has a white background on initial visit:

![image](https://github.com/user-attachments/assets/392512c4-8abb-42f0-834f-ea52929356cd)

Steps to reproduce:

1.  Go to `Monitor` page or any other top level page
2.  Expand the account menu item and navigate to "Password"

saksham-malhotra-27:
Hey ! I can do this one if you say

gorkem-bwl:
> Hey ! I can do this one if you say

Sure! Go ahead please. Make sure you deploy the app and send a screenshot showing it is fixed.

saksham-malhotra-27:
Alright !

saksham-malhotra-27:
Hi ,
I tried reproducing the issue with the "Current Password" field on the Password page when using the dark theme. However, when I visited the page, the background color appeared correct, matching the dark theme, rather than showing a white background as described.

Could you kindly confirm the following to help me understand the issue better?

1. Browser & Version: Could you let me know which browser and version you were using when you encountered the problem?
2. Steps to Reproduce: Did the issue occur immediately after visiting the page, or was there any delay in loading the theme?
3. Clearing Cache: Have you tried clearing the browser cache or testing in an incognito window to ensure the issue persists?

This will help me troubleshoot more effectively.

Additionally, I've noticed that if we use Google-suggested passwords or passwords saved in Google, we encounter the white background in the text field for a moment. I've attached two screenshots for reference:

1. My view, where the background appears correct:
![image](https://github.com/user-attachments/assets/30af5482-f3f5-4cbe-a869-e348cb2165ad)

2. White background issue, which occurs when using Google-suggested passwords:
![image](https://github.com/user-attachments/assets/92339345-8481-45dd-88f5-266770590efb)

If the bug is being reproduced in this way, please let me know, and I can investigate further.

Thank you for your help!

saksham-malhotra-27:
If the issue is due to Google’s remembered passwords, it applies here as well, as shown in the screenshot:
![image](https://github.com/user-attachments/assets/2bcb3b19-9fc3-4c8e-8072-87678b220f2a)

This seems to be related to Chrome's autofill feature, which remembers certain fields like email or password. It’s possible that whenever these remembered fields are used, this problem arises.

[...]

PR #1288, opened 2024-12-05
feat #1275: design corrected  : Current Password Field BG color

saksham-malhotra-27:
This PR addresses issue #1275, where the "Current Password" field in the Password page displayed a white background when using the dark theme. The issue was identified as being linked to Chrome's autofill feature, which applies a default white background to input fields when auto filling saved credentials.

Changes: 
Just added a few lines of code under _globalTheme.js_:
image

Testing:
1. Reproduced the issue using saved passwords in Chrome.
2. Applied the fix and tested in multiple scenarios:
3. Dark theme with autofill.
4. Dark theme without autofill.
5. Light theme.
6. Incognito mode and cleared cache.

Attached before-and-after screenshots for reference.
Before:
image
After:
image

Let me know if there are any additional checks or refinements needed!

### DIFF
--- a/Client/src/Utils/Theme/globalTheme.js
+++ b/Client/src/Utils/Theme/globalTheme.js
@@ -237,6 +237,11 @@ const baseTheme = (palette) => ({
 							WebkitTextFillColor: "unset",
 						},
 					},
+					"& .MuiInputBase-input:-webkit-autofill": {
+						transition: "background-color 5000s ease-in-out 0s", // Long transition to override autofill background
+						WebkitBoxShadow: `0 0 0px 1000px ${theme.palette.background.main} inset`, // Match your theme's background color
+						WebkitTextFillColor: theme.palette.text.primary, // Match text color
+					},
 					"& .MuiInputBase-input.MuiOutlinedInput-input": {
 						padding: "0 var(--env-var-spacing-1-minus) !important",
 					},
--- a/Client/src/Utils/Theme/globalTheme.js
+++ b/Client/src/Utils/Theme/globalTheme.js
@@ -238,9 +238,9 @@ const baseTheme = (palette) => ({
 						},
 					},
 					"& .MuiInputBase-input:-webkit-autofill": {
-						transition: "background-color 5000s ease-in-out 0s", // Long transition to override autofill background
-						WebkitBoxShadow: `0 0 0px 1000px ${theme.palette.background.main} inset`, // Match your theme's background color
-						WebkitTextFillColor: theme.palette.text.primary, // Match text color
+						transition: "background-color 5000s ease-in-out 0s", 
+						WebkitBoxShadow: `0 0 0px 1000px ${theme.palette.background.main} inset`, 
+						WebkitTextFillColor: theme.palette.text.primary,
 					},
 					"& .MuiInputBase-input.MuiOutlinedInput-input": {
 						padding: "0 var(--env-var-spacing-1-minus) !important",


Client/src/Utils/Theme/globalTheme.js
```diff
@@ -237,6 +237,11 @@ const baseTheme = (palette) => ({
 							WebkitTextFillColor: "unset",
 						},
 					},
+					"& .MuiInputBase-input:-webkit-autofill": {
+						transition: "background-color 5000s ease-in-out 0s", 
+						WebkitBoxShadow: `0 0 0px 1000px ${theme.palette.background.main} inset`,
```
ajhollid, line 242: Can you explain the use of BoxShadow here?  Is it being used to set background color?  This seems to be a very roundabout approach to resolving the issue 🤔

ajhollid (commented):
Can you expalin the approach?  I don't follow the reasoning for using box shadow and transitions.  Thanks!

Client/src/Utils/Theme/globalTheme.js
saksham-malhotra-27: Hey ! The box-shadow here (specifically WebkitBoxShadow) to set a background color directly, but rather to override the default background color applied by browsers to **auto filled** input fields.

### Reason being:
Browsers, particularly Chrome and some **WebKit**-based ones, apply a default background color to **auto filled** fields. This color is not easily overridden by standard CSS properties like background-color.
Using the **-webkit-box-shadow** property with an inset shadow of a large size allows developers to **"paint over"** the browser's default styling. By specifying the desired background color **(${theme.palette.background.main})**, the **autofilled** field effectively displays the intended color.

This workaround is especially useful when browsers don’t allow direct manipulation of the autofill background color using the standard background-color property.

saksham-malhotra-27: It works perfectly with this, as demonstrated here:
image
image

Additionally, the pale white background in the text input field was caused by Chrome's autofill feature. To override this, we can't simply use background-color, as it's ineffective once the autofill has applied its own background. Instead, we can use a roundabout approach with **-webkit-box-shadow** to cover the autofill background and apply the desired styling.

saksham-malhotra-27: Also, regarding the transition property; it is intended to animate changes to the background-color of the input. The duration of 5000s is so long that the transition is essentially "frozen," meaning it won't be noticeable. This long duration is used to prevent the browser from abruptly changing the background color when the input is auto filled. Instead, the background color change becomes so gradual (taking thousands of seconds) that it effectively doesn't change visibly at all.

The transition ensures that if the background color of the input were to change (such as during autofill), the change would be gradual. However, because the background change is **"frozen"**, it remains imperceptible. Meanwhile, the **WebkitBoxShadow** is used to hide any background color that might appear as a result of autofill, ensuring the auto filled input looks consistent with the rest of the form.

Together, both properties are used to address the nuisance of the default autofill background color, ensuring a smooth, consistent appearance for the autofilled input fields without disrupting the design.

saksham-malhotra-27: Feel free to let me know if you need any further details regarding the PR. You can also refer to the following links, which demonstrate a similar approach:
[NextUI GitHub Issue #1346](https://github.com/nextui-org/nextui/issues/1346)
[A Trick to Change the Autocomplete Background Color](https://dev.to/atndesign/a-trick-to-change-the-autocomplete-background-color-21ll)

saksham-malhotra-27: If you have the time for this sir, please review it once again !

ajhollid: Hi @saksham-malhotra-27 , thanks for your explanation here.  I'm not a front end expert, so I'll leave the detailed review to @marcelluscaio and @jennifer-gan.

saksham-malhotra-27: Hey ! No problem at all—happy to provide context.

marcelluscaio (approved):
Good workaround. Thank you for solving that, @saksham-malhotra-27

saksham-malhotra-27:
Thank you all !

Issue #1289, opened 2024-12-05
Alignment required for "JWT" section (check the text edit field and dropdown) 


gorkem-bwl:
<img width="887" alt="image" src="https://github.com/user-attachments/assets/332ad382-bd4f-464f-8b72-6201cc4f9abd">

gauravnadkarni:
Hi @gorkem-bwl, Can i take this up?
For clarification: This is about aligning the `JWT TTL units` dropdown on the same line with `JWT time to live` text box, correct?

ajhollid:
@gauravnadkarni sure, open up a PR whenever you're ready!

gorkem-bwl:
Looks good!

PR #1293, opened 2024-12-05
fixed issue with the alignment in JWT section

gauravnadkarni:
This PR addresses the issue under #1289 
Changes done under this PR:
- Fixed the alignment issue with the "JWT TTL Units" dropdown

Before the fix:
before
After the fix:
after


ajhollid (approved):
This seems like a reasonable improvement to me, thanks for making the changes

gauravnadkarni:
@ajhollid Thanks!!

PR #1303, opened 2024-12-06
fix: Refactor Create PageSpeed

ajhollid:
This PR refactors the "Create PageSpeed Monitor" page to remove complexity.

- [x] Remove unnecessary maps from Create PageSpeed
- [x] Simplify handling of notifications and submitting forms
- [x] Set `name` on all inputs, use `event.target.name` instead of passing an extra `formItemName` varialbe to `handleChange`
- [x] Move constants out of component scope


Client/src/Pages/PageSpeed/CreatePageSpeed/index.jsx
```diff
@@ -1,114 +1,73 @@
...
 
+	const SELECT_VALUES = [
+		{ _id: 3, name: "3 minutes" },
+		{ _id: 5, name: "5 minutes" },
+		{ _id: 10, name: "10 minutes" },
+		{ _id: 20, name: "20 minutes" },
+		{ _id: 60, name: "1 hour" },
+		{ _id: 1440, name: "1 day" },
+		{ _id: 10080, name: "1 week" },
+	];
```
marcelluscaio, line 48: I really like the constant approach. I suggest we put them out of the component (line 31) so they don't get recreated on rerender (since they don't depend on props. If they did, it would be a case for using useMemo)

marcelluscaio (commented):
Really like the approach. Suggested some changes, se eif you agree

Client/src/Pages/PageSpeed/CreatePageSpeed/index.jsx
```diff
@@ -126,46 +85,98 @@ const CreatePageSpeed = () => {
...
-					body: "The endpoint you entered doesn't resolve. Check the URL again.",
-				});
-				setErrors({ url: "The entered URL is not reachable." });
-				return;
+			createToast({ body: "Failed to create monitor." });
+		}
+	};
+
+	const handleChange = (event, formItemName) => {
+		const { value } = event.target;
```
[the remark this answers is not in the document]
ajhollid: That's a good idea, will implement!

Client/src/Pages/PageSpeed/CreatePageSpeed/index.jsx
ajhollid: Will do 👌

ajhollid:
> Really like the approach. Suggested some changes, se eif you agree

Agreed on all points, I'll make the revisions!

ajhollid:
@marcelluscaio changes implemented!

marcelluscaio (approved):
Great, thanks Alex!

PR #1306, opened 2024-12-06
Feat/fe/statuspage 1

jennifer-gan:
- [x] create status page when there is no configuration
- [x] Fix Checkbox component conditionally level up the checkbox position so that publish to public checkbox looks correctly in layout
- [x]  General Settings tab Initial layout - WIP


Client/src/Components/TabPanels/Status/GeneralSettingsPanel.jsx
```diff
@@ -0,0 +1,66 @@
...
+			<ConfigBox>
+				<Box>
+					<Stack gap={theme.spacing(6)}>
+						<Typography component="h2">Access</Typography>
+						<Typography component="p">
+							If your status page is ready, you can mark it as published.
+						</Typography>
+					</Stack>
+				</Box>
+				<Checkbox
```
ajhollid, line 54: Checkbox seems to not be positioned correctly:

image

Client/src/Components/TabPanels/Status/GeneralSettingsPanel.jsx
```diff
@@ -0,0 +1,66 @@
...
+					id="published-to-public"
+					label={`Published and visible to the public`}
+					onChange={(e) => handleChange(e)}
+					onBlur={handleBlur}
```
ajhollid, line 58: Required prop `isChecked` is missing, currently throwing an error

image

Client/src/Components/Inputs/Checkbox/index.jsx
```diff
@@ -40,6 +40,7 @@ const Checkbox = ({
 	/* TODO move sizes to theme */
 	const sizes = { small: "14px", medium: "16px", large: "18px" };
 	const theme = useTheme();
+	const override = typeof label == "string" ? {} : { alignSelf: "flex-start" };
```
ajhollid, line 43: This makes the use of `CheckBox` unclear, you would not know that `alignSelf` is set to `flex-start` if the label is not a string unless you had a deep understanding of the implementation of `CheckBox`. 

The point of having the reusable components is that you should not need to know how they work, merely that you can use them.

If you need to be able to set `alignSelf` then it would be better to expose a prop to set it explicitly.  Please include JSdoc as well so that InteliSense can provide information on how to use the component.

ajhollid (changes requested):
Looks like a good start!  

I have some concerns around the `CheckBox` component and there appears to be an alignment issue with the `CheckBox` as well, other than that nice work here, nice and clean!

Client/src/Components/TabPanels/Status/GeneralSettingsPanel.jsx
jennifer-gan: It is fixed in the next PR: Feat/fe/statuspage 2

Client/src/Components/TabPanels/Status/GeneralSettingsPanel.jsx
jennifer-gan: It is fixed in the next PR: Feat/fe/statuspage 2

Client/src/Components/Inputs/Checkbox/index.jsx
jennifer-gan: Fixed here: https://github.com/bluewave-labs/checkmate/pull/1306/commits/54c2826205c3aaa5413eb0394af044457c8b5fc4

Client/src/Components/Inputs/Checkbox/index.jsx
```diff
@@ -36,10 +38,12 @@ const Checkbox = ({
 	value,
 	onChange,
 	isDisabled,
+	alignSelf
 }) => {
 	/* TODO move sizes to theme */
 	const sizes = { small: "14px", medium: "16px", large: "18px" };
 	const theme = useTheme();
+	const override = alignSelf?  { alignSelf: "flex-start" } : {}
```
marcelluscaio, line 46: I am not sure what this is doing, can you provide a print screen @shemy-gan ?

Client/src/Components/TabPanels/Status/GeneralSettingsPanel.jsx
```diff
@@ -0,0 +1,66 @@
...
+						<Typography component="h2">Access</Typography>
+						<Typography component="p">
+							If your status page is ready, you can mark it as published.
+						</Typography>
+					</Stack>
+				</Box>
+				<Checkbox
+					id="published-to-public"
+					label={`Published and visible to the public`}
+					onChange={(e) => handleChange(e)}
```
marcelluscaio, line 57: We can pass handle change just like handleBlur for consistency

Client/src/Pages/Status/CreateStatus/index.jsx
```diff
@@ -0,0 +1,82 @@
...
+ * @returns {JSX.Element}
+ */
+
+const CreateStatus = ({ open = "general settings" }) => {
+	const theme = useTheme();
+	const navigate = useNavigate();
+	const tab = open;
+	const handleTabChange = (event, newTab) => {
+		navigate(`/status/${newTab}`);
+	};
```
marcelluscaio, line 21: we are calling this without passing the new tab. I suggest deleting the second parameter, adding a name to the input, and getting the name from the event target, then passing it to to navigate

marcelluscaio (commented):
Great work, I asked some questions and made some suggestions

Client/src/Components/Inputs/Checkbox/index.jsx
jennifer-gan: The use case is when the label is an element instead of string, it will align the checkbox on top of the container vertically; Alex suggested to expose this explicitly, so the fix was here :https://github.com/bluewave-labs/checkmate/pull/1306/commits/54c2826205c3aaa5413eb0394af044457c8b5fc4

marcelluscaio: Can you show the two cases in a print screen?

Client/src/Pages/Status/CreateStatus/index.jsx
jennifer-gan: yes, the onChange of TabList has the (event, value) as param/signature, so those were passed on to the handleTabChange

Client/src/Components/TabPanels/Status/GeneralSettingsPanel.jsx
jennifer-gan: Agreed, fixed in feat/fe/statuspage 2

Client/src/Components/Inputs/Checkbox/index.jsx
jennifer-gan: Please move on to feat/fe/statuspage 2, the layout is as below
Screenshot from 2024-12-11 19-28-38

marcelluscaio: "Please move on to feat/fe/statuspage 2" what do you mean?

Client/src/Pages/Status/CreateStatus/index.jsx
marcelluscaio: @shemy-gan Please avoid resolving conversations ahead of time. 
HandleTabChange is called on line 46 without the second parameter. I am suggesting removing the second parameter altogether, an deal with that data through the name attribute from the input.

jennifer-gan: The document/signature of onChange has event and value,

(property) onChange?: ((event: React.SyntheticEvent, value: any) => void) | undefined
Callback fired when the value changes.

so  here 	<TabList
						onChange={handleTabChange}
						aria-label="status tabs"
					>
the handleTabChange did have the second param

ajhollid: @jennifer-gan I believe the point here is that if we use the name attribute we don't have to use the second parameter.

Similar to the implementation in the refactor of the various create pages I pushed the other day.

Always nice to reduce the number of parameters needed if at all possible

Client/src/Components/Inputs/Checkbox/index.jsx
marcelluscaio: Oh, ok. So should we close this PR, since it is already contemplated in the other one?

marcelluscaio:
I think we can close this pr since we have https://github.com/bluewave-labs/checkmate/pull/1307 right?

ajhollid:
Closing this PR as there will be major revisions as discussed in the team meeting today

PR #1307, opened 2024-12-06
Feat/fe/statuspage 2

jennifer-gan:
- [x] Updated general settings tab to match BE fields, added validation to fields that are available in BE
- [x] Logo upload is working
- [x] Link general settings and contents tab to sidebar and navigation
- [x] Started working on content tab servers list section 
see below for reference
[Screencast from 2024-12-06 10-34-14.webm](https://github.com/user-attachments/assets/0774e3e4-aeae-4b31-99a3-20d23c9db521)


Client/src/Components/Sidebar/index.jsx
```diff
@@ -51,7 +52,14 @@ const menu = [
 	{ name: "Pagespeed", path: "pagespeed", icon: <PageSpeed /> },
 	{ name: "Infrastructure", path: "infrastructure", icon: <Integrations /> },
 	{ name: "Incidents", path: "incidents", icon: <Incidents /> },
-	// { name: "Status pages", path: "status", icon: <StatusPages /> },
+	{
+		name: "Status pages",
+		icon: <StatusPages />,
+		nested: [
+			{ name: "General Settings", path: "status/general-settings", icon: <Settings /> },
```
ajhollid, line 59: This will open the "settings" submenu as it matches the path. 

```
useEffect(() => {
	const matchedKey = Object.keys(PATH_MAP).find((key) =>
		location.pathname.includes(key)
	);

	if (matchedKey) {
		setOpen((prev) => ({ ...prev, [PATH_MAP[matchedKey]]: true }));
	}
}, [location]);
```

This will have to be revised to function properly.

Client/src/Components/TabPanels/Status/GeneralSettingsPanel.jsx
```diff
@@ -0,0 +1,281 @@
...
+							onBlur={handleBlur}
+							helperText={errors["companyName"]}
+							error={errors["companyName"] ? true : false}
+						/>
+						<TextInput
+							id="url"
+							type="url"
+							label="SubURL"
+							value={localData.url}
+							startAdornment={<HttpAdornment https={true} />}
```
ajhollid, line 197: This won't necessarily be HTTPS, we should provide an http/https selector as in the Create Monitors page

Client/src/Components/TabPanels/Status/GeneralSettingsPanel.jsx
```diff
@@ -0,0 +1,281 @@
...
+						<Select
+							id="theme"
+							label="Theme"
+							value={localData.theme}
+							onChange={handleChange}
+							onBlur={handleBlur}
+							items={[
+								{ _id: "light", name: "Light" },
+								{ _id: "dark", name: "Dark" },
+							]}
```
ajhollid, line 260: I believe we discussed removing this as dark/light theme is set elsewhere in the app

Client/src/Components/TabPanels/Status/GeneralSettingsPanel.jsx
```diff
@@ -0,0 +1,281 @@
...
+								If your status page is ready, you can mark it as published.
+							</Typography>
+						</Stack>
+					</Box>
+					<Stack gap={theme.spacing(6)}>
+						<Checkbox
+							id="published-to-public"
+							label={`Published and visible to the public`}
+							isChecked={localData.publish}
+							value={"localData.publish"}
```
ajhollid, line 165: Looks like accidental stringification here

Client/src/Components/TabPanels/Status/GeneralSettingsPanel.jsx
```diff
@@ -0,0 +1,281 @@
...
+							onBlur={handleBlur}
+							items={timezones}
+							error={errors["display-timezone"]}
+						/>
+						<ImageField
+							id="logo"
+							src={localData.logo?.src ?? logo?.src}
+							loading={progress.isLoading && progress.value !== 100}
+							onChange={handleLogo}
+						/>
```
ajhollid, line 230: This is the field used for the Avatar I believe, I'm not sure if we want this to be a circular image as with the avatar?  @gorkem-bwl can you confirm design spec here?

ajhollid (changes requested):
Looking pretty good to me!  There's a bug in the `Sidebar` due to path matching that needs to be fixed.

Some other minor changes requested throughout.  

Great progress here, nice work!

gorkem-bwl:
@jennifer-gan you can remove the "Theme" section - we won't need it for now.

Client/src/Components/TabPanels/Status/GeneralSettingsPanel.jsx
gorkem-bwl: Is this about the company logo? If so, it is a rectangular shape as can be seen in the Figma file, and not circular. It will be located on top left of the public page, FYI @jennifer-gan

jennifer-gan: Yes, I missed that, I will change that to rectangle

jennifer-gan:
> @jennifer-gan you can remove the "Theme" section - we won't need it for now. If you have worked on it and came so far that removing doesn't make sense, you can keep it.

Yes, I can remote that

Client/src/Components/TabPanels/Status/GeneralSettingsPanel.jsx
jennifer-gan: yes, fixed in next branch: Feat/fe/Status page 3

Client/src/Components/TabPanels/Status/GeneralSettingsPanel.jsx
jennifer-gan: Fixed in : Fixed here: https://github.com/bluewave-labs/checkmate/pull/1307/commits/c301656a7aeb61a6739d4510f1587932dd81bfd4

Client/src/Components/TabPanels/Status/GeneralSettingsPanel.jsx
jennifer-gan: Fixed in : Fixed here: https://github.com/bluewave-labs/checkmate/pull/1307/commits/c301656a7aeb61a6739d4510f1587932dd81bfd4

Client/src/Components/Sidebar/index.jsx
jennifer-gan: I do not see what this part is doing, currently the nest menu item is collapsed initially, and when click on the "Status" on sidebar, the nested is expanded; Let's discuss in the meeting, I thought what we are trying to fulfil here is : 
1. When there is no status page settings, show the static page containing a create status page button
2. When there is status settings, hide the above static page; instead showing the status page with 2 nested panels, each has its entry in the sidebar
Is this correct?

gorkem-bwl: Sure let's discuss. I am not sure I understood this (couldn't wrap my head lol) but we can surely have a look during the meeting :)

Client/src/Pages/Status/CreateStatus/index.jsx
```diff
@@ -0,0 +1,97 @@
...
+ */
+
+const CreateStatus = ({ open = "general-settings" }) => {
+	const theme = useTheme();
+	const navigate = useNavigate();
+	let tabList = ["General Settings", "Contents"];
+	const tab = open
+		.split("-")
+		.map((a) => capitalizeFirstLetter(a))
+		.join(" ");
```
marcelluscaio, line 27: We are using two different things to determine the active tab: navigation + the open prop. 
I suggest ditching the open prop, and use the useParams hook or similar to get the url params.
That way we can simply navigate to the route, without the need to parse that data.

Client/src/Pages/Status/CreateStatus/index.jsx
```diff
@@ -0,0 +1,97 @@
...
+
+	const [tabIdx, setTabIdx] = useState(tabList.indexOf(tab));
+
+	const handleTabChange = (event, newTab) => {
+		let toNavString = newTab.split(" ").map((a) => toLowerCaseFirstLetter(a));
+		setTabIdx(tabList.indexOf(newTab));
+		toNavString = toNavString.join("-");
+		navigate(`/status/${toNavString}`);
+	};
```
marcelluscaio, line 36: Here, instead of having event and newTab, we can just have event, and pass a name to the inputs. We then destructure the name from event.target.name.
Let's try to keep the name the same as the param, to avoid the text transformation back and forth

Client/src/Pages/Status/CreateStatus/index.jsx
```diff
@@ -0,0 +1,97 @@
...
+const CreateStatus = ({ open = "general-settings" }) => {
+	const theme = useTheme();
+	const navigate = useNavigate();
+	let tabList = ["General Settings", "Contents"];
+	const tab = open
+		.split("-")
+		.map((a) => capitalizeFirstLetter(a))
+		.join(" ");
+
+	const [tabIdx, setTabIdx] = useState(tabList.indexOf(tab));
```
marcelluscaio, line 29: If we are changing tab navigation through the url we probably won't need the state

Client/src/Pages/Status/CreateStatus/index.jsx
```diff
@@ -0,0 +1,97 @@
...
+	const handleTabChange = (event, newTab) => {
+		let toNavString = newTab.split(" ").map((a) => toLowerCaseFirstLetter(a));
+		setTabIdx(tabList.indexOf(newTab));
+		toNavString = toNavString.join("-");
+		navigate(`/status/${toNavString}`);
+	};
+
+	useEffect(() => {
+		setTabIdx(tabList.indexOf(tab));
+	}, [tab]);
```
marcelluscaio, line 40: if we define the tab state on the url we probably won't need the effect

Client/src/Pages/Status/CreateStatus/index.jsx
```diff
@@ -0,0 +1,97 @@
...
+							<Tab
+								label={label}
+								key={index}
+								value={label}
+								sx={{
+									fontSize: 13,
+									color: theme.palette.text.tertiary,
+									textTransform: "none",
+									minWidth: "fit-content",
+									minHeight: 0,
```
marcelluscaio, line 74: Why do we need the min height 0?

Client/src/Pages/Status/CreateStatus/index.jsx
```diff
@@ -0,0 +1,97 @@
...
+						onChange={handleTabChange}
+						aria-label="status tabs"
+					>
+						{tabList.map((label, index) => (
+							<Tab
+								label={label}
+								key={index}
+								value={label}
+								sx={{
+									fontSize: 13,
```
marcelluscaio, line 70: Refer to theme font size. Adjust size if needed

Client/src/Pages/Status/CreateStatus/index.jsx
```diff
@@ -0,0 +1,97 @@
...
+									textTransform: "none",
+									minWidth: "fit-content",
+									minHeight: 0,
+									paddingLeft: 0,
+									paddingY: theme.spacing(4),
+									fontWeight: 400,
+									marginRight: theme.spacing(8),
+									"&:focus": {
+										outline: "none",
+									},
```
marcelluscaio, line 81: Why are we removing outline on focus?

Client/src/Utils/stringUtils.js
```diff
@@ -15,3 +15,22 @@ export const capitalizeFirstLetter = (str) => {
...
+		  return "";
+	  }
+	  if (typeof str !== "string") {
+		  throw new TypeError("Input must be a string");
+	  }
+	  if (str.length === 0) {
+		  return "";
+	  }  
+	  return str.charAt(0).toLowerCase() + str.slice(1);
+  };
```
marcelluscaio, line 36: What is the use case for this? Wouldn't just toLowerCase() be enough?

Client/src/Components/TabPanels/Status/GeneralSettingsPanel.jsx
```diff
@@ -0,0 +1,268 @@
...
+	const [errors, setErrors] = useState({});
+	const [logo, setLogo] = useState();
+	const [progress, setProgress] = useState({ value: 0, isLoading: false });
+	const intervalRef = useRef(null);
+	// Local state for form data, errors, and file handling
+	const [localData, setLocalData] = useState({
+		companyName: "",
+		url: "",
+		timezone: "America/Toronto",
+		color: "#4169E1",
```
marcelluscaio, line 33: is this in our theme?

Client/src/Components/TabPanels/Status/GeneralSettingsPanel.jsx
```diff
@@ -0,0 +1,268 @@
...
+		clearInterval(intervalRef.current);
+		setProgress({ value: 0, isLoading: false });
+	};
+
+	const handleChange = (event) => {
+		event.preventDefault();
+		const { value, id } = event.target;
+		setLocalData((prev) => ({
+			...prev,
+			[id]: value,
```
marcelluscaio, line 63: Why are we passing the value of the id as a key for the actual value?

Client/src/Components/Inputs/TextInput/Adornments/index.jsx
```diff
@@ -63,3 +66,38 @@ PasswordEndAdornment.propTypes = {
...
+						pointerEvents: "none",
+						display: "none",
+					},
+				}}
+			>
+				<DeleteTwoToneIcon/>
+			</IconButton>
+		</InputAdornment>
+	);
+};
```
marcelluscaio, line 103: I suggest splitting these components in subfolders

marcelluscaio (commented):
Great work @shemy-gan . I asked some questions and made some suggestions

Client/src/Pages/Status/CreateStatus/index.jsx
jennifer-gan: Hello Caio, although what you suggest sounds like reasonable, I found all of them are needed, just for diffeent cases;  The open prop is for the initial landing, e.g, coming from nested sidebar menu ; the navigation is for when click on the tab to switch between the tabs; the tabIdx state is for showing the actual tab panel; these 3 things look all necessary at the moment as they all take care of different things

Client/src/Pages/Status/CreateStatus/index.jsx
jennifer-gan: event and newTab are all passed down from onChange of Tablist

Client/src/Pages/Status/CreateStatus/index.jsx
jennifer-gan: I found the tabIdx state is in fact needed for actually showing the respective tab panel

Client/src/Pages/Status/CreateStatus/index.jsx
jennifer-gan: this is to show the correspondent tabpanel when switched tab

Client/src/Pages/Status/CreateStatus/index.jsx
marcelluscaio: Yes, I understand. I am suggesting you bing the url parameter to the state. That way you won't need the prop, you will keep the state in the url. 

if the url is status/first-tab you will get the parameter, and you have the state 'first-tab'. That way you can ditch  all of those (props and the functions) and keep just the url and the navigate

Client/src/Pages/Status/CreateStatus/index.jsx
marcelluscaio: I see that, I am suggesting we do not have the new tab anymore, and we get that information from event.target.name

Client/src/Pages/Status/CreateStatus/index.jsx
marcelluscaio: Not if we get our state from the url. Call me if you want me to clarify that

Client/src/Pages/Status/CreateStatus/index.jsx
marcelluscaio: Not if we get our state from the url. Call me if you want me to clarify that

Client/src/Utils/stringUtils.js
jennifer-gan: toLowerCase() will change all to lowercases, this function is needed when only the first letter need to change to lower case, not the rest

marcelluscaio: Yeah, but do we have a use case for that?

Client/src/Pages/Status/CreateStatus/index.jsx
jennifer-gan: in theory that is correct, however event.target.name is empty in this case, I think we should be open to new way , in this case, tablist provided out of box the onChange that gave us the tab value already, why not just use that?

marcelluscaio: Here we are discussing something else, state on the url. The onChange x nae would be on the other thread

Client/src/Pages/Status/CreateStatus/index.jsx
marcelluscaio: Here we are discussing onChange x name or value. Do you see any problem with using the name for that? Otherwise it would be best to follow that for consistency

ajhollid: @jennifer-gan I think what @marcelluscaio suggests here makes sense, `newTab` is indeed passed to the handler, but in our case it makes more sense to use the event since we need a specific value that is not the same as `newTab`.

We gain two benefits

1.  Consistency with how changes are handled with other pages where we use `event.target.name`
2.  Avoiding text transformation - The reason for the transformation won't be clear to newcomers to the project and is an extra step we can avoid.

I think it's worth the change unless you have a strong objection to using the `name` property?

Client/src/Pages/Status/CreateStatus/index.jsx
ajhollid: @marcelluscaio would you mind showing a quick implementation of what you mean here?  I think I understand what you're suggesting but it would be good to see it in on paper so we're all on the same page

ajhollid:
Closing this PR as there will be major revisions as discussed in the team meeting today

PR #1308, opened 2024-12-06
feat: Edit infrastructure monitor page

### LINKED ISSUE #1277: Infrastructure monitors should be editable
Currently an infra monitor cannot be edited. We need to apply the same methodology we use in "Monitors" section. We can add a "Configure" button in details section of an infra monitor so the end user can edit and save again. 

Here is a sample screenshot showing where the Configure button is (this is for the monitors section and you can apply the same method to the Infrastructure section):

image

peterpardo:
### Summary
This PR adds the "Configure" button at the infrastructure details page to edit the selected infrastructure.

### Changes Made
- Added **Infrastructure configure page**
- Added **Configure Button** at the top right of the _Infrastructure details page_ which redirects to the _Infrastructure configure page_
- Auto populate fields when editing infrastructure
- Added **Pause/Resume Infrastructure monitor** functionality
- Added **Delete Infrastructure monitor** functionality

### Related Issue/s
Fixes #1277 

### Screenshots / Demo Video
https://github.com/user-attachments/assets/e58e7b9d-4aad-44b5-af5c-278882139f1e

### Additional Context/s
- **Edit** functionality not yet working. We can create another PR for it
- Noticed that the **pulse dot** which indicates if the infra monitor is active or not doesn't go back to **green** when resumed
- Auto populating the **Customized Alerts** not yet working. We can create another PR for it
- Copy pasted the code in the _Pagespeed configure page_ and made adjustment from there. We can create a PR to refactor redundant codes and do some cleanup

### DIFF
--- a/Client/src/App.jsx
+++ b/Client/src/App.jsx
@@ -41,6 +41,8 @@ import { logger } from "./Utils/Logger"; // Import the logger
 import { networkService } from "./main";
 import { Infrastructure } from "./Pages/Infrastructure";
 import InfrastructureDetails from "./Pages/Infrastructure/Details";
+import ConfigureInfrastructureMonitor from "./Pages/Infrastructure/Configure";
+
 function App() {
 	const AdminCheckedRegister = withAdminCheck(Register);
 	const MonitorsWithAdminProp = withAdminProp(Monitors);
@@ -137,6 +139,10 @@ function App() {
 					<Route
 						path="infrastructure/:monitorId"
 						element={<ProtectedRoute Component={InfrastructureDetailsWithAdminProp} />}
+					/>
+					<Route
+						path="infrastructure/configure/:monitorId"
+						element={<ProtectedRoute Component={ConfigureInfrastructureMonitor} />}
 					/>
 
 					<Route
--- a/Client/src/Features/InfrastructureMonitors/infrastructureMonitorsSlice.js
+++ b/Client/src/Features/InfrastructureMonitors/infrastructureMonitorsSlice.js
@@ -388,6 +388,7 @@ const infrastructureMonitorsSlice = createSlice({
 				state.isLoading = false;
 				state.success = action.payload.success;
 				state.msg = action.payload.msg;
+				state.selectedInfraMonitor = action.payload.data;
 			})
 			.addCase(pauseInfrastructureMonitor.rejected, (state, action) => {
 				state.isLoading = false;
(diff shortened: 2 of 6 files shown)


ajhollid:
Hi @peterpardo !

As always we greatly appreciate your contributions to the project!

We haven't decided on an approach for the "Configure" pages yet however, so we won't be able to address this PR at the current time.

The issue at hand is that the "Configure" pages are practically identical to the "Create" pages, so it probably makes sense just to refactor the "Create" pages for dual use.

Appreciate your hard work here, if we decide to go with the approach of having two separate pages we can revisit this PR!

peterpardo:
> Hi @peterpardo !
> 
> As always we greatly appreciate your contributions to the project!
> 
> We haven't decided on an approach for the "Configure" pages yet however, so we won't be able to address this PR at the current time.
> 
> The issue at hand is that the "Configure" pages are practically identical to the "Create" pages, so it probably makes sense just to refactor the "Create" pages for dual use.
> 
> Appreciate your hard work here, if we decide to go with the approach of having two separate pages we can revisit this PR!

Sure! No worries. Actually, I noticed that as well that there are code duplications in some parts of the pages.

I can try to make a PR to refactor the **Create** and **Configure** pages, but I'll implement it on the **Infrastructure Monitor** only first to avoid a large PR. Let me know if that sounds good.

ajhollid:
> > Hi @peterpardo !
> > As always we greatly appreciate your contributions to the project!
> > We haven't decided on an approach for the "Configure" pages yet however, so we won't be able to address this PR at the current time.
> > The issue at hand is that the "Configure" pages are practically identical to the "Create" pages, so it probably makes sense just to refactor the "Create" pages for dual use.
> > Appreciate your hard work here, if we decide to go with the approach of having two separate pages we can revisit this PR!
> 
> Sure! No worries. Actually, I noticed that as well that there are code duplications in some parts of the pages.
> 
> I can try to make a PR to refactor the **Create** and **Configure** pages, but I'll implement it on the **Infrastructure Monitor** only first to avoid a large PR. Let me know if that sounds good.

I think it would be best to hold off until we decide what direction we're going with these pages.  The maintainers are meeting tomorrow so we'll discuss this topic then, I'll update the issues page after that with the results of our discussion.  Thanks!

ilyakonovalenko:
Hello,

Is there any progress on this?

ajhollid:
Closing this one as the project has moved on considerably.

@peterpardo if you'd still like to implement this one still the approach we've decided on for config pages is to reuse the create pages as they are nearly identical.

PR #1310, opened 2024-12-09
fix: fe/create infra refactor

ajhollid:
This PR refactors the Create Infrastrucutre Monitor page to reduce complexity.

- [x] Refactor `handleChange` method to match implementations in Monitors and PageSpeed create pages
- [x] Simplify state mamangement for `CustomThreshold` 
- [x] Remove unnecessary mapping to reduce complexity
- [x]  Set name on all inputs, use `event.target.name` instead of passing an extra `formItemName` variable to `handleChange`
- [x] Separate handling of notifications
- [x] Remove `onBlur` for the moment, I don't think we need both `onBlur` and `onChange`?  If I recall the discussion correctly the idea behind using `onBlur` was to avoid using `onChange` and the associated performance hit.  If we want to use `onBlur` then I assume the idea is to not use `onChange`?

### DIFF
--- a/Client/src/Pages/Infrastructure/CreateMonitor/index.jsx
+++ b/Client/src/Pages/Infrastructure/CreateMonitor/index.jsx
@@ -1,16 +1,12 @@
 // React, Redux, Router
 import { useTheme } from "@emotion/react";
-import { useNavigate, useParams } from "react-router-dom";
-import { useEffect } from "react";
+import { useNavigate } from "react-router-dom";
 import { useState } from "react";
 import { useSelector, useDispatch } from "react-redux";
 
 // Utility and Network
 import { infrastructureMonitorValidation } from "../../../Validation/validation";
-import {
-	createInfrastructureMonitor,
-	checkInfrastructureEndpointResolution,
-} from "../../../Features/InfrastructureMonitors/infrastructureMonitorsSlice";
+import { createInfrastructureMonitor } from "../../../Features/InfrastructureMonitors/infrastructureMonitorsSlice";
 import { capitalizeFirstLetter } from "../../../Utils/stringUtils";
 
 // MUI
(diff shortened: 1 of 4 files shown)


Client/src/Pages/Infrastructure/CreateMonitor/index.jsx
```diff
@@ -239,56 +250,75 @@ const CreateInfrastructureMonitor = () => {
...
 							isOptional={true}
 							value={infrastructureMonitor.name}
-							onBlur={handleBlur}
 							onChange={handleChange}
 							error={errors["name"]}
 						/>
 						<TextInput
 							type="text"
 							id="secret"
+							name="secret"
```
marcelluscaio, line 319: Why was the name here secret and not email?

Client/src/Pages/Infrastructure/CreateMonitor/index.jsx
```diff
@@ -311,12 +341,11 @@ const CreateInfrastructureMonitor = () => {
 								(notification) => notification.type === "email"
 							)}
 							value={user?.email}
-							onChange={(e) => handleChange(e)}
+							onChange={(event) => handleNotifications(event, "email")}
```
marcelluscaio, line 344: We are calling on change with the previous params?

Client/src/Pages/Infrastructure/CreateMonitor/index.jsx
```diff
@@ -38,173 +73,149 @@ const CreateInfrastructureMonitor = () => {
...
+		const { checked } = event.target;
+		setInfrastructureMonitor({
+			...infrastructureMonitor,
+			[name]: checked,
+		});
+	};
+
+	const handleBlur = (event) => {
+		console.log("handleBlur", event);
+	};
```
marcelluscaio, line 192: I agree, if we are handling errors onChange, we can ditch onBlur

marcelluscaio (commented):
Great implementation. I added some remarks

Client/src/Pages/Infrastructure/CreateMonitor/index.jsx
```diff
@@ -38,173 +73,149 @@ const CreateInfrastructureMonitor = () => {
...
 			description: form.name,
 			teamId: user.teamId,
 			userId: user._id,
 			type: "hardware",
 			notifications: infrastructureMonitor.notifications,
 			thresholds,
 		};
-		return form;
+
+		console.log(form);
```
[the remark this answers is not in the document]
ajhollid: Woops! Thanks for the catch

Client/src/Pages/Infrastructure/CreateMonitor/index.jsx
ajhollid: This field is for the API secret for the Capture monitor I believe

Client/src/Pages/Infrastructure/CreateMonitor/index.jsx
ajhollid: Either that or use onBlur exclusively and ditch onChange?

I'm not sure what best practice is here 🤔

marcelluscaio: We can go with onChange, that's what we are using mostly

Client/src/Pages/Infrastructure/CreateMonitor/index.jsx
ajhollid: > We are calling on change with the previous params?

Eventually we will have more notification types to deal with than "Email" so I added a `type` param to the `handleNotification` method signature:

```
const handleNotifications = (event, type) => {...}
```

This seemed like the easiest way of dealing with potentially different types of notifications :thinking:   I'm open to alternatives as always though!

ajhollid:
> Great implementation. I added some remarks

Thanks for the review, changes made :+1:

Client/src/Pages/Infrastructure/CreateMonitor/index.jsx
marcelluscaio: My bad, I misread the function call!

marcelluscaio (approved):

Issue #1333, opened 2024-12-10
"Monitors" should be "Uptime"


gorkem-bwl:
Since "Monitors" is now a generic name, we should change the wording to "Uptime". 

<img width="903" alt="image" src="https://github.com/user-attachments/assets/31b852ef-2567-4737-9bfe-1eebb88194b6">

felixiho:
hey @ajhollid 
Could you assign this to me ?

ajhollid:
Hi @felixiho ,

Thanks for your interest in the project, this issue has already been resolved however!  Please check out other issues in the log if you'd like to contribute :+1:

PR #1338, opened 2024-12-10
fix:  Monitors -> Uptime resovles #1333

ajhollid:
This PR renames "Monitors" to Uptime and updates all references

- [x] Change "Monitors" to "Uptime"
- [x] Change all links and references "monitors" -> "uptime"


Client/src/Pages/Uptime/Details/index.jsx
```diff
@@ -100,7 +100,7 @@ const DetailsPage = ({ isAdmin }) => {
 			}
 		};
 		fetchCertificate();
-	}, [authToken, monitorId, monitor]);
+	}, [authToken, monitorId, monitor, dateRange, uiTimezone, dateFormat]);
```
marcelluscaio, line 103: Why did we add these 3 dependencies?

marcelluscaio (commented):
Great job. Just one question there

Client/src/Pages/Uptime/Details/index.jsx
ajhollid: I believe they are used in the hook, so we probably want the hook to run again if those values change?

marcelluscaio: Oh yeah, I found two of them but not dateRange

ajhollid: > Oh yeah, I found two of them but not dateRange

You're right, good catch!  I must have fat-fingered it in by accident.  Removed and commited, thank you!

marcelluscaio (approved):

PR #1340, opened 2024-12-10
Adding color for light theme

### LINKED ISSUE #1335: Red button mouse hover status has wrong color
Mouseover color must be #912018 for light theme. 

image

nakul-py:
Fixes #1335


ajhollid:
@nakul-py thanks for your contribution here, however this is a bit too literal of an implementation.

The color is hardcoded and only applies to one button.

If you'd like to work on this issue further please explore the theme and see how colors are set for components.

We would need the color to apply to all buttons.  I'll close this PR for now as it's not quite on target.

PR #1344, opened 2024-12-10
fix: update red palette colors

### LINKED ISSUE #1328: Monitors > Monitor details > Average response time chart color issue
The following chart looks incomplete (the bigger arc has #FFFFFE color which is almost invisible)

image

peterpardo:
### Summary
This PR fixes the issue for average response time chart color under `Monitor > Details` page.

### Changes Made
- Updated color palette for red color
- Updated the old `red300` to `red400`
- Updated `red300` with value of `#EBA8A9`

### Related Issue/s
Fixes #1328 

### Screenshots / Demo Video
**Before:**
issue1328-before

**After:**
issue1328-after

### Additional Context
N/A

### DIFF
--- a/Client/src/Utils/Theme/constants.js
+++ b/Client/src/Utils/Theme/constants.js
@@ -51,7 +51,8 @@ const paletteColors = {
 	red50: "#F9ECED",
 	red100: "#FBD1D1",
 	red200: "#F04438",
-	red300: "#D32F2F",
+	red300: "#EBA8A9",
+	red400: "#D32F2F",
 	red700: "#542426",
 	red800: "#301A1F",
 	orange50: "#FEF8EA",
@@ -109,12 +110,12 @@ const semanticColors = {
 	},
 	error: {
 		main: {
-			light: paletteColors.red300,
-			dark: paletteColors.red300,
+			light: paletteColors.red400,
+			dark: paletteColors.red400,
 		},
 		contrastText: {
-			light: paletteColors.gray50,
-			dark: paletteColors.gray50,
+			light: paletteColors.red300,
+			dark: paletteColors.red300,
 		},
 		light: {
 			light: paletteColors.red100,


Client/src/Utils/Theme/constants.js
```diff
@@ -109,12 +110,12 @@ const semanticColors = {
...
-			light: paletteColors.red300,
-			dark: paletteColors.red300,
+			light: paletteColors.red400,
+			dark: paletteColors.red400,
 		},
 		contrastText: {
-			light: paletteColors.gray50,
-			dark: paletteColors.gray50,
+			light: paletteColors.red300,
+			dark: paletteColors.red300,
```
ajhollid, line 118: I'm not sure about this change, gray to red is quite a change :laughing:   Please wait for @marcelluscaio 's review on this one as he is more familiar with the colors than I am.

ajhollid (commented):
The result looks good to me!  I'm not sure about the contrast text change however, let's wait for the FE team to weigh in on that one, they are more familiar with the color setup than I am.

Client/src/Utils/Theme/constants.js
peterpardo: Got it. Will wait for the feedback.

marcelluscaio: I don't think red 300 is a good color for the contrast text. I would keep colors as they are, and change the component that is calling the colors. I suggest you revert the changes made to the theme, and try to add another color to the component (that probably is calling error.contrastText). If you cannot work with the colors we currently have, the we can think of a way of adding the new shade of red. What do you think?

marcelluscaio (changes requested):

Client/src/Utils/Theme/constants.js
peterpardo: Sounds good! I think adding a new shade of red would be the way to go. We can revert back the `error.contrastText` to use `gray50`. 

For the red colors, I inserted the `#EBA8A9` to be new value of the `red300` and created `red400` to be `#D32F2F` because it's a darker shade of red than the one I inserted.

Then we can just directly use new value of the `red300` into the chart.

What do you think?

marcelluscaio: I would try to solve with the current colors before adding a new one. Let's add the #EBA8A9 only if we cannot solve with our current ones

peterpardo: Got it. I think @gorkem-bwl will be able to help us with this.

These are the shades of color that we have right now:
#F9ECED
#FBD1D1
#F04438
#D32F2F
#542426
#301A1F

Are there any of the colors that you think we can use for the chart? Or should we stick with the #EBA8A9 as the color of the chart?

marcelluscaio: I think you can go with
#FBD1D1
over
#F04438

We need to check light and dark modes. main prblem here was the use of contrastText, that is gray

peterpardo: Here is how the chart looks right now using `#FBD1D1`:

image

I used the `error.light` color palette for this.

What do you think?

marcelluscaio: I like how it looks, we would need to see dark mode as well. You haven't committed that yet right?

peterpardo: Yep. I haven't committed the changes yet.

Here is how it looks in dark mode:
image

peterpardo: > Yep. I haven't committed the changes yet.
> 
> Here is how it looks in dark mode: image

It's using the `red700`: `#542426` as the main color and `red800`: `#301A1F` as the bg color.

marcelluscaio: I think you can commit that. It solves the original issue, and we can open a new one if we need to adjust colors - adjusting colors impacts the whole project and goes beyond the scope of this issue

peterpardo: Sound good! Thank you for reviewing this PR @marcelluscaio. I really appreciate it.

marcelluscaio: I am the one to thank you for contributing to the project =)

ajhollid (approved):

ajhollid:
Nice work figuring out a simple solution to this one @peterpardo and @marcelluscaio , looks good!

PR #1345, opened 2024-12-10
- Remove the login landing page

jennifer-gan:
- [x] Remove the login landing page
- [x] Remove the back button as it is not needed anymore
- [x] Adjust layout due to the removal of back button and remove the propTypes 
requirement for onBack function

as seen below

[Screencast from 2024-12-10 11-19-04.webm](https://github.com/user-attachments/assets/6400ed74-f0c4-46ae-bb22-54424a5999e2)


Client/src/Pages/Auth/Login.jsx
```diff
@@ -19,99 +19,6 @@ import { logger } from "../../Utils/Logger";
 import "./index.css";
```
ajhollid, line 19: `import Mail from "../../assets/icons/mail.svg?react";`

This icon import can be removed, it is no longer used

Client/src/Pages/Auth/Login.jsx
```diff
@@ -168,26 +75,8 @@ const StepOne = ({ form, errors, onSubmit, onChange, onBack }) => {
 					/>
 					<Stack
 						direction="row"
-						justifyContent="space-between"
+						justifyContent="flex-end"
 					>
-						<Button
-							variant="outlined"
-							color="info"
-							onClick={onBack}
```
ajhollid, line 176: `onBack` is no longer used, it can be removed from `StepOne` props

ajhollid (changes requested):
Looks good to me!  Some unused imports and props dangling around that can be removed and we're good to go!

Client/src/Pages/Auth/Login.jsx
```diff
@@ -384,7 +272,7 @@ const Login = () => {
 		password: DEMO !== undefined ? "Demouser1!" : "",
 	});
 	const [errors, setErrors] = useState({});
-	const [step, setStep] = useState(0);
+	const [step, setStep] = useState(1);
```
ajhollid, line 275: Let's stick with 0 indexing as everything else is 0 indexed, we should be consistent.

Client/src/Pages/Auth/Login.jsx
```diff
@@ -533,15 +421,12 @@ const Login = () => {
 					},
 				}}
 			>
-				{step === 0 ? (
-					<LandingPage onContinue={() => setStep(1)} />
-				) : step === 1 ? (
+				{step === 1 ? (
```
ajhollid, line 424: This can be changed to 0 to maintain 0 indexing

Client/src/Pages/Auth/Login.jsx
```diff
@@ -533,15 +421,12 @@ const Login = () => {
...
 					<StepOne
 						form={form}
 						errors={errors}
 						onSubmit={handleSubmit}
 						onChange={handleChange}
-						onBack={() => setStep(0)}
 					/>
 				) : (
 					step === 2 && (
```
ajhollid, line 432: Similarly this can be changed to 1

ajhollid (changes requested):
Just noticed the indexing for steps, lets change it back to 0 indexing as everything else in the app is 0 indexed.

Client/src/Pages/Auth/Login.jsx
jennifer-gan: Fixed in: https://github.com/bluewave-labs/checkmate/commit/76bfb7717ac3eada7d94e11a4ce9291119b45081

Client/src/Pages/Auth/Login.jsx
jennifer-gan: This one was removed

Client/src/Pages/Auth/Login.jsx
jennifer-gan: Fixed in : https://github.com/bluewave-labs/checkmate/commit/76bfb7717ac3eada7d94e11a4ce9291119b45081

Client/src/Pages/Auth/Login.jsx
jennifer-gan: Fixed in: https://github.com/bluewave-labs/checkmate/commit/ba8cc04f2d58cfc0320be926d4af5776a70f7b80

Client/src/Pages/Auth/Login.jsx
```diff
@@ -12,106 +12,12 @@ import TextInput from "../../Components/Inputs/TextInput";
...
-						onClick={onContinue}
-						sx={{
-							width: "100%",
-							"& svg": {
-								mr: theme.spacing(4),
-								"& path": {
-									stroke: theme.palette.other.icon,
-								},
-							},
-							"&:focus-visible": {
```
marcelluscaio, line 56: Maybe we shouldn't ditch all of the initial greeting. We should check that with @gorkem-bwl . I'd keep
- We are pleased to see you again!
- By continuing, you agree to our Terms of Service and Privacy Policy.

Client/src/Pages/Auth/Login.jsx
```diff
@@ -433,9 +320,9 @@ const Login = () => {
 				setErrors((prev) => ({ ...prev, email: error.details[0].message }));
 				createToast({ body: error.details[0].message });
 			} else {
-				setStep(2);
```
marcelluscaio, line 436: @ajhollid I know you requested this change, but this gets a bit confusing to me: we are on the component StepOne, but it is  the step 0. If we are changing to step 0 I would suggest changing the name of the component as well

marcelluscaio (changes requested):
Great, @shemy-gan ! On this one it would be a good opportunity to refactor and apply our rule of not having more than one component per file. Step One and Two could be their own component in their own file. The folder would be Login/components/StepOne and .../StepTwo

Client/src/Pages/Auth/Login.jsx
ajhollid: > @ajhollid I know you requested this change, but this gets a bit confusing to me: we are on the component StepOne, but it is the step 0. If we are changing to step 0 I would suggest changing the name of the component as well

Agreed, it is confusing.  I agree changing the name of the component is a good idea anyways, `StepOne` is really not descriptive as to what Step One actually is.

@jennifer-gan can you please change the names when refactoring out the components?  Thank you!

jennifer-gan: sure, fixed here: https://github.com/bluewave-labs/checkmate/commit/c72334253d9bc4db2cf4daf900c975b884984e9e

ajhollid (approved):
Looks good to me, thanks for refactoring those components

PR #1346, opened 2024-12-10
bug: fix-the-hover-effect-in-button

### LINKED ISSUE #1335: Red button mouse hover status has wrong color

theanuragg:
### what does this PR do ?

- Resolves the hover effect issue for the button.
- Updated hover color for the light theme to **#912018.**

fix #1335

**Changes Made**

- Adjusted the hover styles in the button component.
- Added conditional hover behavior based on the light/dark theme mode.

### DIFF
--- a/Client/src/Pages/Settings/index.jsx
+++ b/Client/src/Pages/Settings/index.jsx
@@ -257,7 +257,14 @@ const Settings = ({ isAdmin }) => {
 									onClick={() =>
 										setIsOpen({ ...deleteStatsMonitorsInitState, deleteStats: true })
 									}
-									sx={{ mt: theme.spacing(4) }}
+									sx={{
+										mt: theme.spacing(4),
+										'&:hover': {
+										  backgroundColor: theme.palette.mode === 'light' 
+											? theme.palette.error.main 
+											: theme.palette.error.dark,
+										},
+									  }}									  
 								>
 									Clear all stats
 								</Button>


theanuragg:
@ajhollid hey

gorkem-bwl:
Please send a video for this. Thanks!

ajhollid (changes requested):
Hi @theanuragg ,

Thanks for taking the time to contribute to the project!

This solution is a bit to literal however, it is only applied to the one `Button`, moreover there is no difference between hover and regular state now, they are both `#D32F2F`

If you'd like to work on this issue further then the change must be applied to _all_ buttons, which means working with the theme file instead of individual buttons.

Client/src/Pages/Settings/index.jsx
```diff
@@ -257,7 +257,14 @@ const Settings = ({ isAdmin }) => {
 									onClick={() =>
 										setIsOpen({ ...deleteStatsMonitorsInitState, deleteStats: true })
 									}
-									sx={{ mt: theme.spacing(4) }}
+									sx={{
+										mt: theme.spacing(4),
+										'&:hover': {
+										  backgroundColor: theme.palette.mode === 'light' 
+											? theme.palette.error.main 
+											: theme.palette.error.dark,
```
ajhollid, line 265: These colors don't match the desing spec in #1335, they should aslo apply to all buttons, not just this specific one

ajhollid:
I'll close this PR for now as it doesn't address the root issue.

Thanks for taking the time to submit a fix, it's not quite on target though.

PR #1354, opened 2024-12-11
fix: FE monitor pause count resolves #1325

### LINKED ISSUE #1325: When a monitor is paused, the paused box doesn't show the correct number
When I pause a monitor, the box showing the number of paused monitors don't show the correct number (is not updated).

image

ajhollid:
This PR resovles the monitor pause count not being updated when a monitor is paused.

The issue is a result of how network calls are being made.  The network call for aggregate monitor stats including pause count happens in the `Monitors` page, whereas the data for individual montiors is fetched in the Monitors Table.

This PR adds a trigger state to the aggergate stat fetching `useEffect` dependency array, and the trigger state is set when the pause button is pressed.  This will cause the aggregate data to be refetched and the pause count to be correctly updated.


marcelluscaio (approved):
Great implementation

PR #1356, opened 2024-12-11
Fix/fe/remove console

ajhollid:
This PR removes a console.log statement from Step three of the registration page


ajhollid:
Wrong base

PR #1366, opened 2024-12-12
Refactor/rever infrastructure

marcelluscaio:
Making all buttons the same


ajhollid (approved):
Thanks for fixing that!

PR #1405, opened 2024-12-15
Issue#1402 - Default color is set as checkmate blue color if no avatar is available

SASHWAT12:
This PR addresses the Default background color when no Avatar is available.


ajhollid (changes requested):
No hardcoded values, please use the theme

PR #1419, opened 2024-12-17
fix: incorrect password on password change causes user to be logged out, resolves #1384

### LINKED ISSUE #1384: Changing a wrong password logs you out
Under the Profile Page and when updating password, if I give a wrong password, the system logs me out. 

It should just say "Wrong password entered" as a warning only and not log me out.

ajhollid:
This PR resolves the issue where if you enter an incorrect password when tryring to update your password you are logged out.

The reason for this is that we have an axios interceptor that boots the user on any 401 responses, ie trying to access resources that they don't have access to.

This is a special case where we don't want the user booted, so an incorrect password on reset now returns 403 instead of 401 which doesn't trigger the interceptor.

- [x] Incorrect password response code changed from 401 -> 403

### DIFF
--- a/Server/controllers/authController.js
+++ b/Server/controllers/authController.js
@@ -281,7 +281,8 @@ const editUser = async (req, res, next) => {
 			const user = await req.db.getUserByEmail(email);
 			// Compare passwords
 			const match = await user.comparePassword(req.body.password);
-			// If not a match, throw a 401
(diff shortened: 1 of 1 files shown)


jennifer-gan (commented):
Looks good and working as expected, just small change to update the comment to 403 instead of 401, then it is good to be merged in

PR #1424, opened 2024-12-17
feat: fe/team empty view, resolves #1400

### LINKED ISSUE #1400: If there are no team members, add a placeholder
If there are no team members in the server, write "There are no team members." instead of showing an empty table. 

Here is an example:

image

ajhollid:
This PR adds an empty veiw Prop to the `BasicTable` component and displays the supplied string as a table row.

- [x] Add an empty view to `BasicTable`
- [x] Supply team table with empty message

image


marcelluscaio (approved):
All right to me

PR #1425, opened 2024-12-17
fix: remove advanced status page, resolves #1379

### LINKED ISSUE #1379: Advanced settings area should be detached and moved to the settings page. Also 'About' section in Advanced Settings should be removed
We have an advanced settings area of its own, but when a user clicks on it, he navigates from the Settings area which is not a good UX practice. We should merge 'Advanced Settings' blocks (3 in total) with the Settings area.

ajhollid:
This PR removes the advanced settings page as discussed with @gorkem-bwl.  The Advanced Settings page exposes settings that can damage the application and are best left to be set by environmental vars in Docker Compose

- [x] Remove advanced settings page
- [x] Remove link to advanced settings page from Settings page
- [x] Remove advanced settigns page from router


marcelluscaio (approved):
All right!

PR #1433, opened 2024-12-18
fix: fe/refactor config box

ajhollid:
This PR creates a reusable `ConfigBox` component and applies it throughout the app.  It also refactors the CreateMaintenance page to use the `ConfigBox`

There were three instances of `ConfigBox` created and exported that were basically identical.  This PR removes those instances and replaces them with a reusable component.

- [x] Create a reusable `ConfigBox` component
- [x] Refactor CreateMaintenaceWindow page to use the component


marcelluscaio (approved):
Nice catch!

PR #1435, opened 2024-12-18
fix: fe/refactor icon box

ajhollid:
This PR creates a reusable IconBox component

- [x] Extract duplicate `IconBox` and refactor into reusable component
- [x] Use `IconBox` throughout applicatication


marcelluscaio (approved):
Good implementation!

PR #1436, opened 2024-12-18
fix: protected route proptype

ajhollid:
This PR fixes the prop type for the `ProtectedRoute` component

- [x] Fix prop type `elementType` -> `element`


marcelluscaio (approved):
Right, it should be element

PR #1444, opened 2024-12-19
fix: add additional not loading flag to fallback conditional rendering, resolves #1426

### LINKED ISSUE #1426: When there are no uptime monitors, do not show a skeletion
If there are no monitors, the skeleton image still shows up there for a couple of seconds, and then vanishes away. We must make sure to check number of monitors first and then disable showing the skeleton if none exists.

image

ajhollid:
This PR adds a `!loading` flag to the conditional rendering of the fallback component.

This component should not display while the page is loading, it should only display when it is certain that there are no montiors.

- [x] Add flag to conditioanl rendering

### DIFF
--- a/Client/src/Pages/Uptime/Home/index.jsx
+++ b/Client/src/Pages/Uptime/Home/index.jsx
@@ -72,11 +72,11 @@ const UptimeMonitors = () => {
 				</Stack>
 				<Greeting type="uptime" />
 			</Box>
-			{!loading && noMonitors && <Fallback isAdmin={isAdmin} />}
 			{loading ? (
 				<SkeletonLayout />
 			) : (
 				<>
+					{noMonitors && <Fallback isAdmin={isAdmin} />}
 					{hasMonitors && (
 						<>
 							<Stack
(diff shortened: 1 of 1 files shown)


Client/src/Pages/Uptime/Home/index.jsx
```diff
@@ -72,7 +72,7 @@ const UptimeMonitors = () => {
 				</Stack>
 				<Greeting type="uptime" />
 			</Box>
-			{noMonitors && <Fallback isAdmin={isAdmin} />}
+			{!loading && noMonitors && <Fallback isAdmin={isAdmin} />}
 			{loading ? (
 				<SkeletonLayout />
 			) : (
```
marcelluscaio, line 78: We could turn these into a ternary:
Loading => Skeleton layout
Else => no monitors => Fallback
Else => has monitors

marcelluscaio (commented):
Good catch, I suggested something that might make the code simpler

Client/src/Pages/Uptime/Home/index.jsx
ajhollid: I'm on board with that, do you want to make changes and push to my branch?

ajhollid: Simplified and pushed

PR #1450, opened 2024-12-20
Feat/be/handle distributed monitors

ajhollid:
This PR lays the groundwork for handling distributed uptime responses

- [x] Add an early return to `JobQueue` if `networkResponse` is `null`.  Network response from distributed uptime will purposefully return `null` as a signal to return early from the job as we don't want to follow the regular monitor flow for distributed uptime monitors

- [x] Add a placeholder request for distributed uptime.  Currently calls the callback, will eventually call third party API endpoint

- [x] Move up ngrok config in the server config so the URL is available before job queue is spun up

### DIFF
--- a/Server/service/networkService.js
+++ b/Server/service/networkService.js
@@ -264,15 +264,18 @@ class NetworkService {
 	}
 
 	async requestDistributed(job) {
+		let callbackUrl;
+		if (process.env.NODE_ENV === "development") {
+			callbackUrl = `${process.env.NGROK_URL}/api/v1/distributed-uptime/results`;
+		} else {
+			callbackUrl = `${process.env.NGROK_URL}/api/v1/distributed-uptime/results`;
+		}
 		try {
-			await this.axios.post(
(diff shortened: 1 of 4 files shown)


Server/service/networkService.js
```diff
@@ -262,6 +263,28 @@ class NetworkService {
...
+		if (process.env.NODE_ENV === "development") {
+			callbackUrl = `${process.env.NGROK_URL}/api/v1/distributed-uptime/results`;
+		} else {
+			callbackUrl = `${process.env.NGROK_URL}/api/v1/distributed-uptime/results`;
+		}
+		try {
+			await this.axios.post(callbackUrl, {
+				targetUrl: job.data.url,
+				id: job.data._id,
+				callback: `${process.env.NGROK_URL}/api/v1/distributed-uptime/results`,
```
jennifer-gan, line 277: line 277 could be replaced by callback: callbackUrl  ?

PR #1456, opened 2024-12-21
Fix: Old name reference on the Server Side

mertssmnoglu:
- [x] Update `openapi.json` to work with new Checkmate name instead of old BlueWave Uptime. Also update description below with the same text on the README.md
- [x] Replace all `BlueWave Uptime` with `Checkmate` under the `Server/` directory.

There are also old naming under the `Client/` directory. If you are thinking changing all `BlueWave Uptime` strings to `Checkmate` in this PR please let me know.

I'm not thinking about updating docs, guides for now.

### DIFF
--- a/Server/controllers/authController.js
+++ b/Server/controllers/authController.js
@@ -361,7 +361,7 @@ const requestRecovery = async (req, res, next) => {
 				url,
 			},
 			email,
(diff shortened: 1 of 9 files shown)


ajhollid (approved):
Thanks for taking care of this @mertssmnoglu, much appreciated

PR #1457, opened 2024-12-23
fix: pie chart colors and Metrics box that was overflowing

marcelluscaio:
- Fixed pie chart background color
- Added new color to theme 
- Changed success light to the new color (attention: this impacts current uses of success light in the project. In case of problem, change success light to success contrastText - which has the same value as success.light had before)
- Changed success contrast text for dark theme
- Fixed overflow on Metric box in pagespeed > details

Solves https://github.com/bluewave-labs/checkmate/issues/1331


ajhollid (approved):
Looking good!

PR #1458, opened 2024-12-23
fix: row background color on hover

marcelluscaio:
I decided to move away from changing background color and try a filter approach, that is applied to the whole row, which will help  keep consistency (everything changes at once)

Solves: https://github.com/bluewave-labs/checkmate/issues/1323


ajhollid (commented):
I don't see any technical issues, @gorkemcetin if you're OK with the styling I'm good on the technical side.

Here are screen captures, first row is hovered in both dark and light mode

image

image

gorkem-bwl:
> I don't see any technical issues, @gorkemcetin if you're OK with the styling I'm good on the technical side.
> 
> Here are screen captures, first row is hovered in both dark and light mode

I am fine with that! Thanks @marcelluscaio @ajhollid

PR #1460, opened 2024-12-23
Fix: group buttons contrast text

marcelluscaio:
Added new semantic color (neuter) and added to MUI settings
Fixed Buttons contrast

### DIFF
--- a/Client/src/Utils/Theme/constants.js
+++ b/Client/src/Utils/Theme/constants.js
@@ -108,6 +108,12 @@ const semanticColors = {
 			dark: paletteColors.green900,
 		},
 	},
+	neuter: {
+		contrastText: {
+			light: paletteColors.blueGray900,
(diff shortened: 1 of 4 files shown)


Client/src/Utils/Theme/constants.js
```diff
@@ -108,6 +108,12 @@ const semanticColors = {
 			dark: paletteColors.green900,
 		},
 	},
+	neuter: {
```
ajhollid, line 111: Is this meant to be "neutral"?

Don't often see neuter used in this context, more often in [this context](https://en.wikipedia.org/wiki/Neutering?wprov=sfla1) 😂

marcelluscaio: OMG YES xD

ajhollid (approved):
Looks good to me :+1:

PR #1479, opened 2024-12-27
feat: fe/tcp port monitoring, resolves #1476

### LINKED ISSUE #1476: Port monitoring
As per the discussions here (https://github.com/bluewave-labs/checkmate/discussions/1447) Checkmate needs a port monitoring functionality. 

Basically it should do the following: 

1. Go to the dashboard and click on "Create New"

2. User selects "TCP port" from the monitor type, as an option in this box: 

image

3. General settings are should change so that the user can add hostname/IP address and the port number.

4. User clicks on Save, and the port monitor is saved.

Check frequency and incident notifications are won't change.

ajhollid:
This PR is the FE half of port monitoring

- [x] Add a Port option to the create Uptime monitor page
- [x] Add validation for Port type monitors   

image

### DIFF
--- a/Server/validation/joi.js
+++ b/Server/validation/joi.js
@@ -144,10 +144,12 @@ const getMonitorsAndSummaryByTeamIdQueryValidation = joi.object({
 	type: joi
 		.alternatives()
 		.try(
-			joi.string().valid("http", "ping", "pagespeed", "docker", "hardware"),
+			joi.string().valid("http", "ping", "pagespeed", "docker", "hardware", "port"),
 			joi
 				.array()
-				.items(joi.string().valid("http", "ping", "pagespeed", "docker", "hardware"))
+				.items(
+					joi.string().valid("http", "ping", "pagespeed", "docker", "hardware", "port")
+				)
 		),
 });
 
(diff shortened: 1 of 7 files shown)


Server/validation/joi.js
```diff
@@ -144,10 +144,12 @@ const getMonitorsAndSummaryByTeamIdQueryValidation = joi.object({
...
 		.alternatives()
 		.try(
-			joi.string().valid("http", "ping", "pagespeed", "docker", "hardware"),
+			joi.string().valid("http", "ping", "pagespeed", "docker", "hardware", "port"),
 			joi
 				.array()
-				.items(joi.string().valid("http", "ping", "pagespeed", "docker", "hardware"))
+				.items(
+					joi.string().valid("http", "ping", "pagespeed", "docker", "hardware", "port")
+				)
```
marcelluscaio, line 152: We could extract these to a variable, maybe an array, and destructure them here. That way we can update in one place and use it in four, do you like the idea?

marcelluscaio (approved):
Just added a small suggestion

Server/validation/joi.js
ajhollid: Absolutely, it was a pain tracking these down all over the app 😂.

Good idea, how about I open a separate PR for that one?

marcelluscaio: Yeah, it's better =)

PR #1481, opened 2024-12-28
fix: be/db indices

ajhollid:
This PR adds indices to some models to improve query performance.  Some queries, especially check queries, have a lot of data and can benefit from DB indices.

- [x] Add indices to Check model
- [x]  Add indices to HardwareCheck model
- [x] Force index use in `getMonitorStatsById`

### DIFF
--- a/Server/db/models/HardwareCheck.js
+++ b/Server/db/models/HardwareCheck.js
@@ -64,5 +64,7 @@ const HardwareCheckSchema = mongoose.Schema(
 );
 
 HardwareCheckSchema.index({ createdAt: 1 });
+HardwareCheckSchema.index({ monitorId: 1, createdAt: 1 });
+HardwareCheckSchema.index({ monitorId: 1, createdAt: -1 });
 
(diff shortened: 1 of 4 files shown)


Server/db/models/HardwareCheck.js
```diff
@@ -64,5 +64,7 @@ const HardwareCheckSchema = mongoose.Schema(
 );
 
 HardwareCheckSchema.index({ createdAt: 1 });
+HardwareCheckSchema.index({ monitorId: 1, createdAt: 1 });
+HardwareCheckSchema.index({ monitorId: 1, createdAt: -1 });
```
jennifer-gan, line 68: Line 67 and 68 will be needed both from client (ascending and descending createdAt)?  so most recent and oldest will be queried in one client request in order to show the details, for past 24 hours and historical data , I see, 
this will improve performance when query for most recent and oldest (such as historical data)

jennifer-gan (approved):
Good work to dig out the mongoDB index part

PR #1486, opened 2024-12-29
fix(server-lint): Resolve lint errors in server-side test files

mertssmnoglu:
This PR mostly includes lint error fixes in the test files. Lint errors happen when **eslint** checks your code and finds problems. These problems can be things like bad formatting, unused variables, or small mistakes. Fixing them helps standardize the code base.

## Changes

- [x] Replace arrow functions with normal functions.
  - Ref: https://mochajs.org/#arrow-functions
  - Rule Name `mocha/no-mocha-arrows`
  - Rule Ref: https://github.com/lo1tuma/eslint-plugin-mocha/blob/main/docs/rules/no-mocha-arrows.md
- [x] Add chai globals to eslint config.
  It was throwing error 'expect is undefined', but we are globally registered the `expect`. Resolved with adding globals.

### DIFF
--- a/Server/tests/controllers/checkController.test.js
+++ b/Server/tests/controllers/checkController.test.js
@@ -9,9 +9,10 @@ import {
 import jwt from "jsonwebtoken";
 import { errorMessages, successMessages } from "../../utils/messages.js";
 import sinon from "sinon";
-describe("Check Controller - createCheck", () => {
+describe("Check Controller - createCheck", function() {
 	let req, res, next, handleError;
-	beforeEach(() => {
+
+	beforeEach(function() {
 		req = {
 			params: {},
 			body: {},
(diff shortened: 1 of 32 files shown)


Server/tests/controllers/checkController.test.js
```diff
@@ -9,9 +9,10 @@ import {
 import jwt from "jsonwebtoken";
 import { errorMessages, successMessages } from "../../utils/messages.js";
 import sinon from "sinon";
-describe("Check Controller - createCheck", () => {
+describe("Check Controller - createCheck", function() {
```
ajhollid, line 12: What's the reasoning for replacing the arrow functions with traditional funciton expression?

ajhollid (approved):
Thanks for the updates here :+1:

PR #1493, opened 2024-12-30
Create status page PR1

jennifer-gan:
- [x] Updated existing components for the status page usage
- [x] Added validation rules for status page
- [x] Added createStatusPage and getStatusPageByUrl
- [x] Add libraries for drag and drop
- [ ] Add the rest of status page ( will be in another PR )


Client/src/Components/Inputs/TextInput/Adornments/index.jsx
```diff
@@ -23,14 +25,15 @@ export const HttpAdornment = ({ https }) => {
 				color={theme.palette.text.secondary}
 				sx={{ lineHeight: 1, opacity: 0.8 }}
 			>
-				{https ? "https" : "http"}://
+				{prefix !== undefined ? prefix : https ? "https://" : "http://"}
```
ajhollid, line 28: What's the purpose of adding the `prefix` prop here?  I'm not sure I understand the use case, could you elaborate?

Client/src/Validation/error.js
```diff
@@ -48,6 +66,10 @@ const hasValidationErrors = (form, validation, setErrors) => {
 					"refreshTokenTTL",
 					"jwtTTL",
 					"notify-email-list",
+					"_id",
+					"__v",
+					"createdAt",
+					"updatedAt"
```
ajhollid, line 72: These look like MongoDB values, why are they needed in the validation?

ajhollid (commented):
I don't see any huge issues here, just a couple things I don't quite understand.  Please see comments and let me know if you can elaborate! :rocket:

PR #1524, opened 2025-01-07
Inform Admin incase of server issue.

jasneetsingh6114:
## Describe your changes

Created toast notifications that will alert the admin whenever there is a server issue and would inform admin too when there would be proper functioning of the server.

## Issue number
1482: https://github.com/bluewave-labs/checkmate/issues/assigned/jasneetsingh6114

- [ Done] I deployed the application locally.
- [ Done] I have performed a self-review and testing of my code.
- [ Done] I have included the issue # in the PR.
- [Done ] I have labelled the PR correctly.
- [ Done] The issue I am working on is assigned to me.
- [ I didn't hard-coded any values] I didn't use any hardcoded values (otherwise it will not scale, and will make it difficult to maintain consistency across the application).
- [Done ] I made sure font sizes, color choices etc are all referenced from the theme.
- [Done ] My PR is granular and targeted to one specific feature.
- [Done] I took a screenshot and attached to this PR if there is a UI change.
Screenshot 2025-01-06 203810
Screenshot 2025-01-06 205058

### DIFF
--- a/Client/src/Utils/NetworkService.js
+++ b/Client/src/Utils/NetworkService.js
@@ -1,14 +1,20 @@
+import { toast } from 'react-toastify';
 import axios from "axios";
 const BASE_URL = import.meta.env.VITE_APP_API_BASE_URL;
 const FALLBACK_BASE_URL = "http://localhost:5000/api/v1";
 import { clearAuthState } from "../Features/Auth/authSlice";
 import { clearUptimeMonitorState } from "../Features/UptimeMonitors/uptimeMonitorsSlice";
+ 
 class NetworkService {
 	constructor(store, dispatch, navigate) {
 		this.store = store;
 		this.dispatch = dispatch;
 		this.navigate = navigate;
 		let baseURL = BASE_URL;
+		this.lastToastTimestamp = 0;
+		this.errorOccurred = false; 
+		this.user = this.store.getState().auth?.user;
+		this.isAdmin = (this.user?.role?.includes("admin") ?? false) || (this.user?.role?.includes("superadmin") ?? false);
 		this.axiosInstance = axios.create();
 		this.setBaseUrl(baseURL);
 		this.unsubscribe = store.subscribe(() => {
@@ -21,15 +27,17 @@ class NetworkService {
 				baseURL = FALLBACK_BASE_URL;
 			}
 			this.setBaseUrl(baseURL);
-		});
+		});		
 		this.axiosInstance.interceptors.response.use(
-			(response) => response,
-			(error) => {
-				if (error.response && error.response.status === 401) {
-					dispatch(clearAuthState());
-					dispatch(clearUptimeMonitorState());
-					navigate("/login");
+			(response) => {	
+				if (this.isAdmin && this.errorOccurred && response.status >= 200 && response.status < 300) {
+					this.debounceToast("Checkmate server is running properly.", "success");
+					this.errorOccurred = false;// Reset error flag if a successful response is received after errors
 				}
+				return response;
+			},
+			(error) => {
+				this.handleError(error);
 				return Promise.reject(error);
 			}
 		);
@@ -39,6 +47,29 @@ class NetworkService {
 		this.axiosInstance.defaults.baseURL = url;
 	};
 
+	handleError(error) {
+		if (error.response) {
+			const status = error.response.status;
+			if (status === 401) {
+				this.dispatch(clearAuthState());
+				this.dispatch(clearUptimeMonitorState());
+				this.navigate("/login");
+			} else if (this.isAdmin && status >= 500) {
+				this.debounceToast("Checkmate server is not running or has issues. Please check.", "error");
+				this.errorOccurred = true;
+			}
+		} 
+	}
+
+	debounceToast(message, type, debounceDuration = 5000) {
+		const now = Date.now();
+		if (now - this.lastToastTimestamp > debounceDuration) {
+			this.lastToastTimestamp = now;
+			type === "success" ? toast.success(message) : toast.error(message);
+		}
+	}
+
+	
 	cleanup() {
 		if (this.unsubscribe) {
 			this.unsubscribe();
--- a/Client/src/Utils/NetworkService.js
+++ b/Client/src/Utils/NetworkService.js
@@ -1,20 +1,16 @@
-import { toast } from 'react-toastify';
 import axios from "axios";
 const BASE_URL = import.meta.env.VITE_APP_API_BASE_URL;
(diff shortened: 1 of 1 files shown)


gorkem-bwl:
Out of curiosity are those the toast messages same as the ones in components folder?

jasneetsingh6114:
I see the configuration of toast in file(https://github.com/bluewave-labs/checkmate/blob/develop/Client/src/Utils/toastUtils.jsx) which is commonly used under components. Presently, I haven't used exactly those same toast but if you want me I can update the code with using those particular toast.

gorkem-bwl:
> I see the configuration of toast in file(https://github.com/bluewave-labs/checkmate/blob/develop/Client/src/Utils/toastUtils.jsx) which is commonly used under components. Presently, I haven't used exactly those same toast but if you want me I can update the code with using those particular toast.

Yes. If there is a same functionality under components, there is no need to reinvent the wheel.

ajhollid:
Thanks for the submission @jasneetsingh6114, I'll have a look at it as soon as I can!

ajhollid:
@jasneetsingh6114 I had a brief look and I think you can safely remove all the admin checks and just display the toast for all users.

All users should be aware the server isn't responding

Client/src/Utils/NetworkService.js
```diff
@@ -1,14 +1,20 @@
...
 class NetworkService {
 	constructor(store, dispatch, navigate) {
 		this.store = store;
 		this.dispatch = dispatch;
 		this.navigate = navigate;
 		let baseURL = BASE_URL;
+		this.lastToastTimestamp = 0;
+		this.errorOccurred = false; 
+		this.user = this.store.getState().auth?.user;
+		this.isAdmin = (this.user?.role?.includes("admin") ?? false) || (this.user?.role?.includes("superadmin") ?? false);
```
ajhollid, line 17: No need to do admin checks, the network error toast shoudl be shown to all users

Client/src/Utils/NetworkService.js
```diff
@@ -1,14 +1,20 @@
...
 import { clearUptimeMonitorState } from "../Features/UptimeMonitors/uptimeMonitorsSlice";
+ 
 class NetworkService {
 	constructor(store, dispatch, navigate) {
 		this.store = store;
 		this.dispatch = dispatch;
 		this.navigate = navigate;
 		let baseURL = BASE_URL;
+		this.lastToastTimestamp = 0;
+		this.errorOccurred = false;
```
ajhollid, line 15: We don't need to worry about informing the user if the server goes back up, if it's up they will know

Client/src/Utils/NetworkService.js
```diff
@@ -39,6 +47,29 @@ class NetworkService {
...
+			}
+		} 
+	}
+
+	debounceToast(message, type, debounceDuration = 5000) {
+		const now = Date.now();
+		if (now - this.lastToastTimestamp > debounceDuration) {
+			this.lastToastTimestamp = now;
+			type === "success" ? toast.success(message) : toast.error(message);
+		}
```
ajhollid, line 69: Why do you need to debounce the toast?  This doesn't seem necessary?

Client/src/Utils/NetworkService.js
```diff
@@ -21,15 +27,17 @@ class NetworkService {
...
-			(error) => {
-				if (error.response && error.response.status === 401) {
-					dispatch(clearAuthState());
-					dispatch(clearUptimeMonitorState());
-					navigate("/login");
+			(response) => {	
+				if (this.isAdmin && this.errorOccurred && response.status >= 200 && response.status < 300) {
+					this.debounceToast("Checkmate server is running properly.", "success");
+					this.errorOccurred = false;// Reset error flag if a successful response is received after errors
 				}
```
ajhollid, line 36: I appreciate the idea here, but I believe it is unnecessary complexity.  If the server responds then the user is arleady aware it is functioning properly.  This can be removed

Client/src/Utils/NetworkService.js
```diff
@@ -39,6 +47,29 @@ class NetworkService {
...
+	handleError(error) {
+		if (error.response) {
+			const status = error.response.status;
+			if (status === 401) {
+				this.dispatch(clearAuthState());
+				this.dispatch(clearUptimeMonitorState());
+				this.navigate("/login");
+			} else if (this.isAdmin && status >= 500) {
+				this.debounceToast("Checkmate server is not running or has issues. Please check.", "error");
+				this.errorOccurred = true;
```
ajhollid, line 59: This won't work if the server doesn't respond at all, which is the issue at hand.

If there is no response, `erorr.response` will be undefined and the user will not be notfied.

We should be checking for `error.request`, which is the property that will be defined in the event of a request error.

ajhollid (changes requested):
@jasneetsingh6114 as mentioned earlier the admin check is not necessary for this functionality, it can be removed.

The logic for checking if the server is responding or not is not quite correct either.

All that needs to be done in this PR is the following:

```js
if (error.response) {
   // current response error implementation
} else if (error.request) {
    // Show a toast informing the user the server didn't respond
}

That's all that is needed here, this should be like a 5 line PR.

Simple is often best :smile:

Client/src/Utils/NetworkService.js
jasneetsingh6114: done

Client/src/Utils/NetworkService.js
jasneetsingh6114: done, updated code

Client/src/Utils/NetworkService.js
jasneetsingh6114: The purpose of debounceToast is to prevent multiple toast notifications from being displayed in quick succession, which could overwhelm or annoy the user. It ensures that toast messages are displayed only after a specified interval has passed since the last one. That was my thought when i decided to use it but i removed in the updated code but now it throws multiple errors continously.

Client/src/Utils/NetworkService.js
jasneetsingh6114: Sure did that

Client/src/Utils/NetworkService.js
ajhollid: Hmm yes I see your point, if there is more than one network request sent off per page then there will be a toast displayed for each one.

I think ultimately the solution is to reduce the number of network requests, we'll work on that in the future. 

Appreciate your consideration of the user experience!

ajhollid (approved):
Thanks for your patience and for making requested changes, this looks good :+1: 

I understand your reasoning for debouncing the toast, but I think that we should address that at a higher level and resolve the issue of multiple network requests entirely.

Thank you for your contribution to the project!

Issue #1112, opened 2024-11-04
Incidents page shows meaningless records for all servers, filtering not working


YunusStar:
**Describe the bug**
Incidents page shows meaningless records like Network Error, Service Unavailable for the same timestamps. When you select related monitoring, it does not show this records.

**To Reproduce**
1. Go to 'Incidents'
2. Click on 'All Servers'
3. See error

**Expected behavior**
Filtering should work as expected and we should see filtered records.

**Screenshots**
<img width="939" alt="Screenshot 2024-11-04 at 09 25 58" src="https://github.com/user-attachments/assets/047f1014-89bc-48bb-9364-4b1eac3b3a93">

**Desktop (please complete the following information):**
 - Browser [Chrome, Version 122.0.6261.129 (Official Build) (arm64)]
 - Version [BlueWave Uptime v1.1.0]

**Additional context**

ajhollid:
HI @YunusStar,

Are you talking about filtering with these buttons:

![image](https://github.com/user-attachments/assets/44a3bbce-7b1f-4f43-9ee1-0919194588ea)

If so yes this is an open issue, that feature is unimplemented at the moment.

gorkem-bwl:
@shyna92 assigned this to you :)

YunusStar:
Hi @ajhollid,

Yes, I mean that filtering buttons, but also there are records that does not match with real time scenario. The sites are not down but there are down records with the same timestamps. Also, when you try to filter from left site, changing from Incidents for _all servers_ to Incidents for _(MONITOR-NAME)_, you can not see these down records.

KalyanBhowmick:
Hi @gorkem-bwl is this issue still open?

gorkem-bwl:
Yes, please check the issue and try to find the root cause.

KalyanBhowmick:
Sure @gorkem-bwl , I will check and raise the PR

KalyanBhowmick:
Root cause analysis: As @ajhollid already said filter functionality is not yet implemented.
Solution: I will update the query for the applied filter so that we get filtered data on each button click

@gorkem-bwl

gorkem-bwl:
Go ahead @KalyanBhowmick ! :)

PR #1526, opened 2025-01-07
Query updated for Incidents page filter

KalyanBhowmick:
## Describe your changes

Query updated for Incidents page to make the filters working. As per the BaseCheckSchema status can have true and false values in it. We have 3 filters in this page All, Down and Cannot Resolve in UI , on schema side we can true for up, false for down. So now when we click Down we will get status which are set to false and clicking on Cannot resolve will bring records which have status true.

## Issue number

#1112 

### DIFF
--- a/Server/db/mongo/modules/checkModule.js
+++ b/Server/db/mongo/modules/checkModule.js
@@ -175,9 +175,10 @@ const getTeamChecks = async (req) => {
 				case "all":
 					break;
 				case "down":
+					checksQuery.status = false;
 					break;
 				case "resolve":
-					checksQuery.statusCode = 5000;
+					checksQuery.status = true;
 					break;
 				default:
 					logger.warn({
--- a/Server/db/mongo/modules/checkModule.js
+++ b/Server/db/mongo/modules/checkModule.js
@@ -178,7 +178,7 @@ const getTeamChecks = async (req) => {
 					checksQuery.status = false;
 					break;
 				case "resolve":
-					checksQuery.status = true;
+					checksQuery.statusCode = 5000;
 					break;
 				default:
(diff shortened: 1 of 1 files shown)


Server/db/mongo/modules/checkModule.js
```diff
@@ -175,9 +175,10 @@ const getTeamChecks = async (req) => {
 				case "all":
 					break;
 				case "down":
+					checksQuery.status = false;
```
ajhollid, line 178: This is good :+1:

Server/db/mongo/modules/checkModule.js
```diff
@@ -175,9 +175,10 @@ const getTeamChecks = async (req) => {
...
 					break;
 				case "resolve":
-					checksQuery.statusCode = 5000;
+					checksQuery.status = true;
 					break;
```
ajhollid, line 182: the "resolve" case is meant to return all checks that have an indeterminate status, ie there was a network error, the server didn't respond, etc.  These checks are given status 5000 when they fail to resolve.

This should not return checks with `status === true`, it should retrun checks where `statusCode === 5000`

ajhollid (changes requested):
Thanks for working on this!  The "down" case is correct, but the "resolve" case is meant to return checks with `statusCode === 5000` so you can revert that part.

Server/db/mongo/modules/checkModule.js
KalyanBhowmick: Changes are done and pushed . Check once when you are back @ajhollid

ajhollid (approved):
Looking good, thank you for your contribution!

PR #1556, opened 2025-01-12
Fix/colors general

marcelluscaio:
This is a draft, please pull and test


ajhollid (changes requested):
I'll post the issues here as I come across them

Button color is missing:

image

`theme.palette.percentage` is no longer defined:

image

Pending color seems incorrect:

image

Seems to be an issue with displaying pagespeed monitors:

image

There is an issue with icon for infrastrucutre monitors:

image

Issue with table header for darkmode, already noted but here for reference:

image

I can't review further into the details pages as they don't load up at this point, will get to those once these are resolved!

Issue #1545, opened 2025-01-10
Build a notification system that will cover Discord, Slack, Telegram, Apprise and webhooks


gorkem-bwl:
**This issue is a part of the [$5000 grant funding program](https://checkmate.so/grants).** 

More information and screenshots will follow.

Enissay:
I hope this includes support for [ntfy](https://ntfy.sh/)

gorkem-bwl:
> I hope this includes support for [ntfy](https://ntfy.sh/)

Currently not, but since the notification system is there for uptime monitoring, we can easily add more. :) Thanks for the heads up @Enissay

Br0wnHammer:
Closing as completed.

profucius:
I see that this was closed, but just want to comment in support of adding Ntfy to the notifications system. I use Ntfy for everything else, so it would be perfect for me to use it instead of discord or telegram, etc. Please (re)consider adding support for this in the future. Thanks

gorkem-bwl:
Yes, we can do, but I must admit it's way under the priority list for now. There are several features / functionalities to add really before ntfy.sh is in.

PR #1589, opened 2025-01-19
Integrate Apprise notification service.

Skorpios604:
## Describe your changes
Add Apprise notification service integration

## Issue number
#1545


Server/db/models/Notification.js
```diff
@@ -8,7 +8,18 @@ const NotificationSchema = mongoose.Schema(
...
+		appriseUrl: {
+			type: String,
+			validate: {
+				validator: function(v) {
+					return !this.type || 
+						   this.type !== 'apprise' || 
+						   /^[a-zA-Z]+:\/\//.test(v);
+				},
+				message: 'Invalid Apprise URL format'
+			}
```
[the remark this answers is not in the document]
Skorpios604: This suggestion might not be the best for Apprise URLs.

Apprise uses custom URL schemes for different services, for example:
discord://webhook_id/webhook_token
telegram://bottoken/ChatID
slack://tokenA/tokenB/tokenC
matrix://hostname?t=token

Skorpios604 (commented):
1. Only runs when type is 'apprise'
2. Checks length to prevent buffer overflow
3. Validates against Apprise's URL format
4. Allows for authentication in the URL
5. Supports all Apprise notification services

Skorpios604 (commented):
- Pin Apprise API to stable version 1.2.0
- Add healthcheck monitoring
- Set resource limits and reservations

.env
```diff
@@ -0,0 +1 @@
+APPRISE_API_URL=http://localhost:8000
```
ajhollid, line 1: .env files shouldn't be committed, best to remove this commit entirely

Server/db/models/Notification.js
```diff
@@ -8,7 +8,21 @@ const NotificationSchema = mongoose.Schema(
...
+		},
+		appriseUrl: {
+			type: String,
+			validate: {
+				validator: function(v) {
+					if (this.type !== 'apprise') return true;
+					if (!v || v.length > 2048) return false;
+					
+					// Apprise URL format: scheme://[user:pass@]host[:port]/path
+					const appriseUrlRegex = /^[a-zA-Z]+:\/\/([^:\s]+:[^@\s]+@)?[^\s\/:]+(:\d+)?(\/[^\s]*)?$/;
```
ajhollid, line 21: This regex will also allow for invalid Apprise URLs, this will match for any alphabetical string for schema, like `http://` or `abcd://`

If we want to strictly verify Apprise URLs only, we'll have to match for all the specific protocols.  

Otherwise we may as well just verify that this is a string and leave the responsibility for a valid URL with the user.

docker-compose.yml
```diff
@@ -0,0 +1,21 @@
...
+    timeout: 10s
+    retries: 3
+  deploy:
+    resources:
+      limits:
+        cpus: '0.50'
+        memory: 256M
+      reservations:
+        cpus: '0.25'
+        memory: 128M
```
ajhollid, line 21: I can see this docker compose file is meant to spin up an instance of Apprise, but how does this fit into the application structure?

I don't believe we should have the responsibility of maintaining a docker compose file that runs an instance of Apprise for end users, I think that's their responsibility.  I believe it is sufficient that we provide the integration :thinking: 

I will clarify the requirements, but if it turns out we do need to provide an instance of Apprise then this should be baked into our existing docker compose files.

Server/db/models/Notification.js
```diff
@@ -8,7 +8,21 @@ const NotificationSchema = mongoose.Schema(
 		},
 		type: {
 			type: String,
-			enum: ["email", "sms"],
+			enum: ["email", "sms", "apprise"],
```
ajhollid, line 11: :+1:

ajhollid (changes requested):
Looks like a good start :+1: 

There's a `.env` that has been comitted, we should pop off that commit so it is not included in our repo history.

I'm not 100% clear on the requirements for this PR, so I'm not sure how the docker compose file fits into things just yet.  Will clarify requirements and revisit this!

Skorpios604:
Going with a different approach

Issue #1573, opened 2025-01-16
Custom JavaScript validation for HTTP(s) JSON responses


gorkem-bwl:
Some services require more complex checks of the response than a simple KEYWORD or REGEX to perform the validation "Status UP" or "Status DOWN". To accommodate this, it would be highly beneficial to allow custom JavaScript code to be executed against a parsed JSON object from an HTTP(s) response.

The parsed JSON object could be accessible from the custom JavaScript code as a variable named `msg`. 

The custom JavaScript code would be required to return:

-  Either`true` or `false` for binary validation results.
- or an array for more precise or detailed validation results.

### Solution
Create a new monitor type: **"HTTP(s) JSON Custom Function"**, derived from the existing **"HTTP(s) - Keyword"** monitor type.

In the UI:
- Add a new text zone where users can write custom JavaScript code to validate the response.
- Provide clear instructions and examples to guide users on writing valid code.

### Example usage

1. The HTTP(s) response body is parsed into a JSON object.
2. The JSON object is passed to the custom JavaScript function, accessible as `msg`.
3. Example:

    ```javascript
    // Check if all users in the response are active
    return msg.users.every(user => user.isActive);
    ```

4. The result (`true`, `false`, or an array) determines the status of the monitor.

This allows users to implement advanced validation logic that cannot be expressed using simple keywords or regular expressions, and increases flexibility and use cases for the monitoring tool.

### Potential Challenges

- Ensuring the security of executing user-provided JavaScript code.
- Implementing proper error handling and logging for invalid or malicious scripts.

### Additional Notes
- Consider running the JavaScript code in a secure sandboxed environment to mitigate potential security risks.
- Provide example templates for common use cases in the UI to simplify adoption.

jasneetsingh6114:
Hey @gorkem-bwl  I would love to take up this issue

gorkem-bwl:
Sure. Could you please arrange a time with me here to discuss on the details of this issue? It's a big one so I wanted to make sure we're on the same page. 

https://tidycal.com/bluewavelabs/general

You can have some sort of research (eg what to code/how to design the UI), so we can deep dive about talking about the implementation. Feel free to check other similar products.

gorkem-bwl:
Wanted to check back here @jasneetsingh6114

YinDongFang:
May I take up this if it's still open?

[...]

PR #1634, opened 2025-01-26
Support custom scripts to determine whether it is up/down based on the response JSON

YinDongFang:
## Test Cases （UI change is only for test in local, not included in PR）

- HTTP(S) without user scripts
image

image

- HTTP(S) with error testScripts 
image
image

(with error logs)
image

- HTTP(S) with testScripts (return true)
image

image

- HTTP(S) with testScripts (return false)
image

image

- HTTP(S) with testScripts (infinite loop)
image

image

(with error logs)
image

## Describe your changes

- add `sandbox.js` for securely executing user scripts.
- add `testScripts` poperty to Monitor to store custom scripts.
- modify `requestHttp` function to execute the script and determine the result status if `testScripts` is not empty.

## Issue number

#1573

### DIFF
--- a/Server/db/models/Monitor.js
+++ b/Server/db/models/Monitor.js
@@ -38,6 +38,9 @@ const MonitorSchema = mongoose.Schema(
 				"distributed_http",
 			],
 		},
+		testScripts: {
+			type: String,
+		},
 		url: {
 			type: String,
 			required: true,
--- a/Server/package-lock.json
+++ b/Server/package-lock.json
@@ -19,6 +19,7 @@
 				"handlebars": "^4.7.8",
 				"helmet": "^8.0.0",
 				"ioredis": "^5.4.2",
+				"isolated-vm": "^5.0.3",
 				"joi": "^17.13.1",
 				"jsonwebtoken": "9.0.2",
 				"mailersend": "^2.2.0",
@@ -2642,7 +2643,6 @@
 			"version": "6.0.0",
 			"resolved": "https://registry.npmjs.org/decompress-response/-/decompress-response-6.0.0.tgz",
 			"integrity": "sha512-aW35yZM6Bb/4oJlZncMH2LCoZtJXTRxES17vE3hoRiowU2kWHaJKFkSBDnDR+cm9J+9QhXmREyIfv0pji9ejCQ==",
-			"dev": true,
 			"license": "MIT",
 			"dependencies": {
 				"mimic-response": "^3.1.0"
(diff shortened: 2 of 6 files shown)


ajhollid (commented):
This is a potentially _very_ dangerous pull request.  Executing user provided is not something to be taken lightly.

I am umfailiar with the sandbox library used here, and thus I am not prepared to approve this PR.

If you have some resources that can show me that this library is absolutely bulletproof and that you've thoroughly tested this implementation we can look at it.

I appreciate the contribution very much, however this is probably _the_ most dangerous thing you can do in a web based application so we must be thorough here :joy:

YinDongFang:
Thank you for your feedback.

From what I initially learned, the `vm` can be used to execute code in Node. But recording to [What is the difference between Node's vm and vm2](https://www.npmjs.com/package/vm2#what-is-the-difference-between-nodes-vm-and-vm2) and [vm2 discontinued](https://github.com/patriksimek/vm2/issues/533), that is not a reliable solution.

`isolated-vm` is recommended by the maintainer of `vm2`, and there is [case of migration](https://github.com/patriksimek/vm2/issues/533#issuecomment-1630056522). According to the [README](https://github.com/laverdet/isolated-vm?tab=readme-ov-file#who-is-using-isolated-vm), some projects are using this library.

Executing untrusted code is dangerous, and I cannot show you absolute safety (just like the vulnerability that was discovered in `vm2`, despite its high download count). I think the security of the `isolated-vm` library is sufficient to meet this requirement. No Node API provided in `isolated-vm` and also don't need it in user scripts.

For more security, perhaps could use `child_process` to run in a separate process?

Alternatively, if there are any other suggestions/possible solutions, please feel free to share them, and I would be happy to do more research. If there are other implementation approaches for this issue (such as avoiding user code execution), Maybe we can discuss further in the issue.

ajhollid:
> Thank you for your feedback.
> 
> From what I initially learned, the `vm` can be used to execute code in Node. But recording to [What is the difference between Node's vm and vm2](https://www.npmjs.com/package/vm2#what-is-the-difference-between-nodes-vm-and-vm2) and [vm2 discontinued](https://github.com/patriksimek/vm2/issues/533), that is not a reliable solution.
> 
> `isolated-vm` is recommended by the maintainer of `vm2`, and there is [case of migration](https://github.com/patriksimek/vm2/issues/533#issuecomment-1630056522). According to the [README](https://github.com/laverdet/isolated-vm?tab=readme-ov-file#who-is-using-isolated-vm), some projects are using this library.
> 
> Executing untrusted code is dangerous, and I cannot show you absolute safety (just like the vulnerability that was discovered in `vm2`, despite its high download count). I think the security of the `isolated-vm` library is sufficient to meet this requirement. No Node API provided in `isolated-vm` and also don't need it in user scripts.
> 
> For more security, perhaps could use `child_process` to run in a separate process?
> 
> Alternatively, if there are any other suggestions/possible solutions, please feel free to share them, and I would be happy to do more research. If there are other implementation approaches for this issue (such as avoiding user code execution), Maybe we can discuss further in the issue.

Hi @YinDongFang,

Thanks for sharing some resources there, I will do some reading.

As this project is mainly meant to be a self hosted solution I am concerned about users running untrusted code on their own machines.

I'll dig into those resources as soon as I have a chance to do so and we can touch base again there and see if we agree that security is sufficient.

Thanks for your contribution and I appreaciate your patience, we have to be careful we don't make a mistake with this one :+1:

ajhollid:
Closing this one as we have an alternative approach now.  Thank you for your contribution here though, much appreciated!

Issue #1640, opened 2025-01-27
Redundant instantiation `NetworkService` in `NetworkServiceProvider`


YinDongFang:
**Describe the bug**
Direct create new instance of `NetworkService` in `NetworkServiceProvider` component will create redundant instance.

**To Reproduce**

Need run in local

1. add `console.log('new');` in class `NetworkService` constructor
2. add `console.log('store change');` in `store.subscribe`;
3. change page by click left menu (You will see 'new' in console every time)
4. change store state by click `Action/Configure` in uptime home page (You will see duplicate 'store change' in console)

**Expected behavior**
NetworkService is single instance, should be create only once.

PR #1641, opened 2025-01-27
Fix redundant instantiation of NetworkService

YinDongFang:
## Describe your changes

Fix redundant instantiation of NetworkService

## Issue number

#1640 

### DIFF
--- a/Client/src/Utils/NetworkServiceProvider.jsx
+++ b/Client/src/Utils/NetworkServiceProvider.jsx
@@ -1,14 +1,16 @@
 import { useDispatch } from "react-redux";
 import { useNavigate } from "react-router";
 import { setNetworkService } from "./NetworkService";
-import NetworkService from "./NetworkService";
+import NetworkService, { networkService } from "./NetworkService";
 import { store } from "../store";
 
 const NetworkServiceProvider = ({ children }) => {
 	const dispatch = useDispatch();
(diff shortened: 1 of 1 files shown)


Client/src/Utils/NetworkServiceProvider.jsx
```diff
@@ -1,14 +1,16 @@
...
+import NetworkService, { networkService } from "./NetworkService";
 import { store } from "../store";
 
 const NetworkServiceProvider = ({ children }) => {
 	const dispatch = useDispatch();
 	const navigate = useNavigate();
-	const networkService = new NetworkService(store, dispatch, navigate);
-	setNetworkService(networkService);
+	if (!networkService) {
+		const networkService = new NetworkService(store, dispatch, navigate);
```
ajhollid, line 11: Let's either rename this or assign directly without saving to a variable first.

We don't want to shadow the `networkService` exporeted from `NetworkService`

Good catch here!

ajhollid (changes requested):
Good catch on the redundant instantiaton :+1:   Can you either rename the local `networkService ` variable or use the newly created `NetworkService` directly with the `setNetworkService` method?  Best to avoid shadowing the imported `networkService` which is quite confusing.  Thank you!

ajhollid (approved):
Thanks for making the requested changes!

PR #123, opened 2024-06-11
Feat/forgot password, resolves #121

ajhollid:
Implemented backend routes for forgot password flow

1.  `POST /recovery/request` - Requests a recovery token from the backend, 10 minute expiration
2. `POST /recovery/validate` - Validates a recovery token issued from the backend
3. `POST /recovery/reset` - Sets `User` password to new password

PR #1647, opened 2025-01-27
feat: i18n support for backend

Cihatata:
## Describe your changes

i18n support has been added to the backend using POEditor APIs. During the build process, translations registered in POEditor are retrieved. Messages are then returned based on the user's language. At the middleware layer, the user's language is detected from the request headers.

Note
Before merging, POEditor keys must be added to the project.

## Issue number

[Mention the issue number(s) this PR addresses (e.g., #123).](https://github.com/bluewave-labs/Checkmate/issues/1396)

- [ ] I have labelled the PR correctly.
- [x ] The issue I am working on is assigned to me.
- [ ] My PR is granular and targeted to one specific feature.


Server/utils/messages.js
```diff
@@ -1,147 +1,156 @@
...
+	FRIENDLY_ERROR: (language) => getTranslatedMessage('FRIENDLY_ERROR', language),
+	UNKNOWN_ERROR: (language) => getTranslatedMessage('UNKNOWN_ERROR', language),
 
 	// Auth Controller
-	UNAUTHORIZED: "Unauthorized access",
-	AUTH_ADMIN_EXISTS: "Admin already exists",
-	AUTH_INVITE_NOT_FOUND: "Invite not found",
+	UNAUTHORIZED: (language) => getTranslatedMessage('UNAUTHORIZED', language),
+	AUTH_ADMIN_EXISTS: (language) => getTranslatedMessage('AUTH_ADMIN_EXISTS', language),
+	AUTH_INVITE_NOT_FOUND: (language) => getTranslatedMessage('AUTH_INVITE_NOT_FOUND', language),
```
ajhollid, line 19: Rather than have to pass `req.language` to every single piece of the backend code why don't we create a `StringService` and just instantiate it once with the users preferred language and handle getting the required strings there?

That would be a much cleaner and maintainable solution.  Adding a layer of abstraction will also allow us to swap to i18n tools if we ever need to.

Server/service/translationService.js
```diff
@@ -0,0 +1,187 @@
+import axios from 'axios';
```
ajhollid, line 1: Please use the NetworkService for all network requests

Server/service/translationService.js
```diff
@@ -0,0 +1,187 @@
...
+import fs from 'fs';
+import path from 'path';
+import { formattedKey } from '../utils/formattedKey.js';
```
ajhollid, line 4: Please inject all depdencies rather than importing to facilitate testing

Server/service/translationService.js
```diff
@@ -0,0 +1,187 @@
...
+      }
+    } catch (error) {
+      this.logger.error({
+        message: error.message,
+        service: 'TranslationService',
+        method: 'initialize',
+        stack: error.stack
+      });
+    }
+  }
```
ajhollid, line 33: What happens to the application if initialization here fails entirely?  Does it still run?

ajhollid (changes requested):
This is a great start!

I believe it would make more sense to create a `StringService` or something along those lines to handle providing strings rather than pass `req.language` to every single part of the application that requires a `string`.

How is error handling and fallback handled?  If the i18n service fails entirely will the app still run as expected?  It looks like it defaults to `en`, that should be the case in event of any error.  Please elaborate on this :+1:

Server/controllers/authController.js
```diff
@@ -76,7 +77,7 @@ class AuthController {
 			// If superAdmin exists, a token should be attached to all further register requests
 			const superAdminExists = await this.db.checkSuperadmin(req, res);
 			if (superAdminExists) {
-				await this.db.getInviteTokenAndDelete(inviteToken);
+				await this.db.getInviteTokenAndDelete(inviteToken, req.language);
```
ajhollid, line 80: We shouldn't need req.language anymore, you can pass a reference to `StringService` to `MongoDB` when it is instantiated

Server/controllers/authController.js
```diff
@@ -176,7 +177,7 @@ class AuthController {
 			userWithoutPassword.avatarImage = user.avatarImage;
 
 			return res.success({
-				msg: successMessages.AUTH_LOGIN_USER,
+				msg: successMessages.AUTH_LOGIN_USER(req.language),
```
ajhollid, line 180: Same for these, these strings should now all be provided by the `StringService`

ajhollid (changes requested):
Continuing good work!

I haven't had a chance to do a thorough review yet, but I still see `req.language` being used throughout the controllers.

I believe the idea is to move all of the strings to `StringService`, that way we should never need to use `req.language` and all the strings will be in one location.

I'll have a more thorough look as soon as possible!

Cihatata:
> Continuing good work!
> 
> I haven't had a chance to do a thorough review yet, but I still see `req.language` being used throughout the controllers.
> 
> I believe the idea is to move all of the strings to `StringService`, that way we should never need to use `req.language` and all the strings will be in one location.
> 
> I'll have a more thorough look as soon as possible!

I have prepared the stringService architecture. In places where we will use translations, we can retrieve texts using:

`stringService.authIncorrectPassword`

If this implementation is okay for you, I will replace the text occurrences accordingly.
You can check`Server/controllers/authController.js` for an example usage.

ajhollid:
> > Continuing good work!
> > I haven't had a chance to do a thorough review yet, but I still see `req.language` being used throughout the controllers.
> > I believe the idea is to move all of the strings to `StringService`, that way we should never need to use `req.language` and all the strings will be in one location.
> > I'll have a more thorough look as soon as possible!
> 
> I have prepared the stringService architecture. In places where we will use translations, we can retrieve texts using:
> 
> `stringService.authIncorrectPassword`
> 
> If this implementation is okay for you, I will replace the text occurrences accordingly. You can check`Server/controllers/authController.js` for an example usage.

I see many references to `req.language` being passed around still, I don't want to couple the controllers or DB tightly to this i8n implementation.

Here's what I'd like to see in the service:

1.  Set up once on initializaiton with the user's language preference
2.  Service provides _all_ strings that will be used on the backend
3.  All strings from the `/utils/messages` file should be moved to the service
4.  Replace all references to `messages.whateverMessage` should be replaced with `stringService.getWhateverMessage` or something along those lines.

eg

```
	if (match !== true) {
		const error = new Error(errorMessages.AUTH_INCORRECT_PASSWORD);
		error.status = 401;
		next(error);
		return;
	}
```

should now be

```
	if (match !== true) {
		const error = new Error(stringSerice.get(AUTH_INCORRECT_PASSWORD));
		error.status = 401;
		next(error);
		return;
	}
```

or something along those lines.

The exact implementation here is up to you, but it must be decoupled and have minimal impact on the rest of the application.

Server/service/translationService.js
Cihatata: If we cannot fetch the translations from POEditor API, it will fetch the translations from `local_en.json` in local as defaults

Server/controllers/authController.js
```diff
@@ -8,19 +8,21 @@ import {
...
 } from "../validation/joi.js";
 import logger from "../utils/logger.js";
-import { errorMessages, successMessages } from "../utils/messages.js";
 import jwt from "jsonwebtoken";
 import { getTokenFromHeaders, tokenType } from "../utils/utils.js";
 import crypto from "crypto";
 import { handleValidationError, handleError } from "./controllerUtils.js";
 const SERVICE_NAME = "authController";
+import ServiceRegistry from "../service/serviceRegistry.js";
+import StringService from "../service/stringService.js";
```
ajhollid, line 17: These should be initialized in the root server file and passed to the controllers as a dependency

Server/controllers/authController.js
```diff
@@ -8,19 +8,21 @@ import {
 	newPasswordValidation,
 } from "../validation/joi.js";
 import logger from "../utils/logger.js";
-import { errorMessages, successMessages } from "../utils/messages.js";
 import jwt from "jsonwebtoken";
 import { getTokenFromHeaders, tokenType } from "../utils/utils.js";
 import crypto from "crypto";
 import { handleValidationError, handleError } from "./controllerUtils.js";
 const SERVICE_NAME = "authController";
+import ServiceRegistry from "../service/serviceRegistry.js";
```
ajhollid, line 16: No imports in any classes except the root server directory please.

All dependencies should be injected through the constructor

Server/controllers/authController.js
```diff
@@ -176,7 +178,7 @@ class AuthController {
 			userWithoutPassword.avatarImage = user.avatarImage;
 
 			return res.success({
-				msg: successMessages.AUTH_LOGIN_USER,
+				msg: this.stringService.authLoginUser,
```
ajhollid, line 181: 👍 This is what I something like what I was thinking, looks good

Server/db/mongo/modules/inviteModule.js
```diff
@@ -1,6 +1,7 @@
 import InviteToken from "../../models/InviteToken.js";
 import crypto from "crypto";
-import { errorMessages } from "../../../utils/messages.js";
+import ServiceRegistry from "../../../service/serviceRegistry.js";
+import StringService from "../../../service/stringService.js";
```
ajhollid, line 4: Again no imports in db models please, everything should be injected to facilitate testing

ajhollid (changes requested):
This is looking better, nice work!

All dependencies should be injected via constructor and not imported, each component should be totally decoupled from all other components.

A controller should not depend on a service that is imported. It should interact with the service via a reference it gets from the constructor.

This greatly facilitates testing and modularity of the application.

For example if this translation library breaks all I have to do is replace string service with a totally different implementation. 

So long as a service fulfils the same API no other class should care how the service is implemented.

Nice work, thanks for making changes!

Server/db/mongo/modules/inviteModule.js
Cihatata: I don't understand how to do this for inviteModule. InviteModule is not a class. Can you give an example?

It has been done this way for `monitorModule` before.

Server/utils/messages.js
```diff
@@ -1,147 +1,156 @@
...
-	EXPIRED_REFRESH_TOKEN: "Refresh token expired",
-	REQUEST_NEW_ACCESS_TOKEN: "Request new access token",
+	UNKNOWN_SERVICE: (language) => getTranslatedMessage('UNKNOWN_SERVICE', language),
+	NO_AUTH_TOKEN: (language) => getTranslatedMessage('NO_AUTH_TOKEN', language),
+	INVALID_AUTH_TOKEN: (language) => getTranslatedMessage('INVALID_AUTH_TOKEN', language),
+	EXPIRED_AUTH_TOKEN: (language) => getTranslatedMessage('EXPIRED_AUTH_TOKEN', language),
+	NO_REFRESH_TOKEN: (language) => getTranslatedMessage('NO_REFRESH_TOKEN', language),
+	INVALID_REFRESH_TOKEN: (language) => getTranslatedMessage('INVALID_REFRESH_TOKEN', language),
+	EXPIRED_REFRESH_TOKEN: (language) => getTranslatedMessage('EXPIRED_REFRESH_TOKEN', language),
+	REQUEST_NEW_ACCESS_TOKEN: (language) => getTranslatedMessage('REQUEST_NEW_ACCESS_TOKEN', language),
```
ajhollid, line 28: This file is not needed at all anymore, correct?  It can be deleted

ajhollid (changes requested):
Looks great!  Thanks for incoporating those changes.

A few things to clean up:

1.  Can you move the service init and register to the respecitve sections?  They are noted by comments

`// Create Services`

`// Register Services`

just to keep things nice and tidy :+1:   

2.  I believe we can delete `messages.js` now as it is no longer needed

3.  Can the fallback `locales_en.json` file be moved to the `locales` directory where the other translations will be?

4.  A nodemon config is needed to run the server in dev mode, currently it is stuck in a bootloop

Nice work here!

Server/index.js
```diff
@@ -156,6 +160,11 @@ const startApp = async () => {
 		}
 	}
 
+	const networkService = new NetworkService(axios, ping, logger, http, Docker, net);
+	const translationService = new TranslationService(logger, networkService);
+	const stringService = new StringService(translationService);
+	ServiceRegistry.register(StringService.SERVICE_NAME, stringService);
```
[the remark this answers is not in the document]
Cihatata: > Please move this registration to the service registration section

I have to do this here because I need to register the String Service before the MongoDB is created.

Cihatata:
> Looks great! Thanks for incoporating those changes.
> 
> A few things to clean up:
> 
> 1. Can you move the service init and register to the respecitve sections?  They are noted by comments
> 
> `// Create Services`
> 
> `// Register Services`
> 
> just to keep things nice and tidy 👍
> 
> 2. I believe we can delete `messages.js` now as it is no longer needed
> 3. Can the fallback `locales_en.json` file be moved to the `locales` directory where the other translations will be?
> 4. A nodemon config is needed to run the server in dev mode, currently it is stuck in a bootloop
> 
> Nice work here!

I have completed everything except the first item. You can review the changes.

1 - I have to do this here because I need to register the String Service before the MongoDB is created.

ajhollid:
Hey @Cihatata,

I think @gorkem-bwl has discussed already with you about some changes that will be required for this implementation as we will have to drop API support.

Do you want to close this PR and open a new one as the changes will be significant?

gorkem-bwl:
> Hey @Cihatata,
> 
> I think @gorkem-bwl has discussed already with you about some changes that will be required for this implementation as we will have to drop API support.
> 
> Do you want to close this PR and open a new one as the changes will be significant? Or if you wan to keep working on this PR that's fine too.

Just fyi and to be on the same page with you guys - I'll request a Github integration for our account soon so we'll be using that one moving forward. Let's also wait to hear from me/them as well.

Cihatata:
i will open new pull request

Issue #1396, opened 2024-12-14
Support for translation (and other localization)


ceskyDJ:
**Is your feature request related to a problem? Please describe.**
I've just found your interesting project and it looks just… great! I quickly went through the project and didn't find any localization stuff. I want every good software to be fully localized to as much as possible languages, so it could be used by even non-english speaking people. So, I'm kindly asking you about adding localization or at least simple translation possibility to your product.

**Describe the solution you'd like**
I'm a Czech man, so please consider that my opinions could be biased. I'd recommend to use some translation management system as [Weblate](https://weblate.org/en/) or [Tolgee](https://tolgee.io/). When you choose a good one, integration of translations (and a few other localization stuff) would be pretty easy and efficient. These two have their advantages and you definitely should take a look at them. Weblate is very popular solution used by many big projects. It has very good integration with Git, so it's like another developer in your team. The interface for translators is very clean and easy to use, do you can get so many translations in a short piece of time. I have great experience with it as an active translator of some open-source projects. Tolgee is relatively new player. It brings in-context automatic translation in the scene and in-app translation possibilities for translators. It's very good approach with great results in point of translation quality. As a translator I know the context is just very important thing to care about.

**Describe alternatives you've considered**
Of course, there are other systems you can use. Or, as many projects, you don't need any of them. But, as I've seen so far, these projects doesn't grow so quickly as these with translation management systems. They are just too far for non-technical oriented translators.

**Additional context**
If you decide to support localization, you can count on me with Czech translation and helping with some other localization for Czech people (our ancestors had just great ideas, how the best language and it's typography look like…). I offer maintained translation, as I started to do some time ago. Do, I'll constantly update the translation, as new strings will arrive with the growth of the product. I check older translations and localization stuff, too, so I try to achive very quality experience for our people.

gorkem-bwl:
Thank you Michal. We'll definitely have a look at Weblate and Tolgee. I tried to sign up for Weblate but it seems like their signup page is broken. Should be a temporary thing tho. We'll definitely adding this feature to the roadmap soon as more than 50% of users are outside of US & Canada, so that makes a real sense to me.

ceskyDJ:
> I tried to sign up for Weblate but it seems like their signup page is broken. Should be a temporary thing tho.

Oh, interesting. I hope they'll fix it if the problem persists. But, as I tried it a few months ago, it seemed to work. I don't know if they have Discord, Slack, etc., but I'm sure they can help you via email.

[...]

PR #1709, opened 2025-02-06
I18n support for server

Cihatata:
## Describe your changes
This is a continuation of the closed Pull Request. =>  https://github.com/bluewave-labs/Checkmate/pull/1647

Add i18n Support & Fetch Translations in CI/CD
We decided not to store the POEditor API key in the repository. Instead, translations will now be fetched during the CI/CD pipeline and committed to the repository. On the server side, translations will be served directly from the locales directory.

To enable fetching translations from POEditor in the pipeline, we need a new GitHub Action. However, I couldn't add it due to insufficient permissions.

This PR introduces i18n support for Server while ensuring a secure and automated way to manage translations.

## Issue number

#1396 

- [ ] I have labelled the PR correctly.
- [ ] The issue I am working on is assigned to me.
- [ ] My PR is granular and targeted to one specific feature.


Server/middleware/languageMiddleware.js
```diff
@@ -0,0 +1,11 @@
+const languageMiddleware = (stringService, translationService) => (req, res, next) => {
+  const acceptLanguage = req.headers['accept-language'] || 'en';
+  const language = acceptLanguage.split(',')[0].slice(0, 2).toLowerCase();
+
+  translationService.setLanguage(language);
+  stringService.setLanguage(language);
```
ajhollid, line 6: Is middleware necessary for this?  It seems excessive to set the language on every request :thinking: 

I imagine we can just set this once on intialization, and then we can update it when the user changes their settings?

ajhollid (commented):
This all looks pretty good to me!

The middleware setting the language on every request seems wasteful to me however, it seems to me it mihgt be a better approach to set the language once on init and then set language again if the user changes their preference?

Server/middleware/languageMiddleware.js
Cihatata: If you mean storing the language preference in the database within the user data and retrieving it from there, we can't do that for authentication pages.

If you’re suggesting setting it once and reusing it without rechecking, I’m not sure how to implement that. How would we detect when the language changes?

Cihatata:
> This all looks pretty good to me!
> 
> The middleware setting the language on every request seems wasteful to me however, it seems to me it mihgt be a better approach to set the language once on init and then set language again if the user changes their preference?

If you mean storing the language preference in the database within the user data and retrieving it from there, we can't do that for authentication pages.

If you’re suggesting setting it once and reusing it without rechecking, I’m not sure how to implement that. How would we detect when the language changes?

ajhollid:
> > This all looks pretty good to me!
> > The middleware setting the language on every request seems wasteful to me however, it seems to me it mihgt be a better approach to set the language once on init and then set language again if the user changes their preference?
> 
> If you mean storing the language preference in the database within the user data and retrieving it from there, we can't do that for authentication pages.
> 
> If you’re suggesting setting it once and reusing it without rechecking, I’m not sure how to implement that. How would we detect when the language changes?

Good point about the login pages, I didn't consider those.

As for how to detect changes, the language setting should be stored in the db, so when that setting is updated we can also update the translation service.

Let's leave it for now, we can revisit this again later if needed.

ajhollid (approved):

ajhollid:
@Cihatata can you resolve conflicts here please?  Thank you!

ajhollid (approved):

YinDongFang:
Hi @Cihatata! This is a great PR. I used used `stringService` to replace hardcoded strings in `requestHttp`. I have a question about the implementation:

I noticed the code uses both `translationService` and `stringService`, but it seems that `stringService` is just a wrapper for getting all keys from the locales json. It needs to add a `get` method to `stringService` for every locale key. I'm wondering why not use `translationService.getTranslation` directly?

Issue #1705, opened 2025-02-06
Modals display centered relative to the viewport, not their container


ajhollid:
Expected behaviour:  Modal displays in the center of its parent component

Actual behaviour: Modal displays in the center of the viewport

![Image](https://github.com/user-attachments/assets/a7e8022e-50b1-48e4-8f10-39e8f58dd85f)

bhavesh26:
@ajhollid can I work on this bug?

Br0wnHammer:
@ajhollid can I take this on?

SoloDevAbu:
### When ever i am trying to run this project then this error is showing 
Can anyone tell me why
@ajhollid @Br0wnHammer 
- store.js:36 Uncaught ReferenceError: Cannot access 'settingsReducer' before initialization at store.js:36:12

gorkem-bwl:
@bhavesh26 assigned to you. Thank you!

@SoloDevAbu if you have an issue with the application, do not comment on an arbitrary issue - please ask your question in the Github Discussions area.

bhavesh26:
@ajhollid  , @gorkem-bwl  
https://github.com/bluewave-labs/Checkmate/pull/1728 Please review and merge

Br0wnHammer:
Hey @gorkem-bwl, I think we can complete this too before the new version launch. What say?

gorkem-bwl:
Sure, go ahead and create a PR for this please.

Aditya-singh-divineboy:
i want to fix it.

Br0wnHammer:
> i want to fix it.

Hello @Aditya-singh-divineboy, a PR is already open addressing this issue. Feel free to contribute there if you have any specific idea.

gorkem-bwl:
@Aditya-singh-divineboy you can go ahead with this issue.

gorkem-bwl:
Yes, please. Thank you.

sumaiyahibrahim:
Hi! I'm new to open source and I’d love to work on this issue.  
Could you please assign it to me?

gorkem-bwl:
> Hi! I'm new to open source and I’d love to work on this issue. Could you please assign it to me?

Hello. This issue is already assigned and there is an open PR for this. Suggest checking other issues. Thank you.

gorkem-bwl:
> Hi [@gorkem-bwl](https://github.com/gorkem-bwl), Can I work on this issue?

[...]

PR #1728, opened 2025-02-09
bug(modal-at-center-div)-issue-1705

bhavesh26:
## Describe your changes

Removed translate transform so it centers to parent div instead of view port

## Issue number

Mention the issue number(s) this PR addresses  #1705 .

### DIFF
--- a/Client/src/Components/Dialog/genericDialog.jsx
+++ b/Client/src/Components/Dialog/genericDialog.jsx
@@ -20,7 +20,6 @@ const GenericDialog = ({ title, description, open, onClose, theme, children }) =
 					position: "absolute",
 					top: "50%",
 					left: "50%",
-					transform: "translate(-50%, -50%)",
 					width: 400,
 					bgcolor: theme.palette.primary.main,
 					border: 1,


Client/src/Components/Dialog/genericDialog.jsx
```diff
@@ -20,7 +20,6 @@ const GenericDialog = ({ title, description, open, onClose, theme, children }) =
 					position: "absolute",
 					top: "50%",
 					left: "50%",
-					transform: "translate(-50%, -50%)",
```
ajhollid, line 23: This does not change the reference position of the element, removing this just moves the element along it's x and y axis by 50%.  It is still positioned relative to the viewport.

The element should be positioned relative to its parent component as opposed to the veiwport.

ajhollid (changes requested):
This PR does not address the underlying issue, the dialog should be positioned relative to the parent component.

Client/src/Components/Dialog/genericDialog.jsx
bhavesh26: @ajhollid  I have passed the parent reference using useRef is that method fine? or do i need to do something materil ui specific?

ajhollid: > @ajhollid I have passed the parent reference using useRef is that method fine? or do i need to do something materil ui specific?

I'm not sure a ref is necessary here? This seems like a problem that should be able to be resolved purely with css.

marcelluscaio: Agree with Alex, this would not solve the issue. The problem is not in the dialog component but where it is being called, it probably lives inside of a div that is mispositioned?

marcelluscaio (changes requested):

ajhollid:
Hey @bhavesh26 ,

I'm closing this PR as the project has undergone a signicicant structure reorganization.  If you'd like to work on this please open a new PR once you've update your project.

Thank you!

Issue #1588, opened 2025-01-19
Suggestion/Feature Request: Export reports


louanfontenele:
- Why?
- In this case, I think it’s interesting to have this feature because we were previously pursuing legal action against a hosting provider. Before discovering Checkmate, we were using Uptime Kuma, which doesn’t have this functionality. I had to resort to a workaround to generate reports and present them in court. This option is excellent because it allows you to have saved data that can serve as evidence, not just for legal matters like the one I mentioned, but for other purposes as well.

Something in HTML would be more interesting, styled to match the site's own design, with options for dark mode or light mode, and including some key data such as:

- The date the site was added to the system  
- How many times the site went down  
- How many times the site came back online  
- How many times the site was paused  
- Uptime percentage  
- Minimum ping recorded  
- Maximum ping recorded  
- Average ping recorded (Average Response Time)  
- Total number of incidents  
- Total number of checks  
- Check interval (e.g., 1 minute, 3 minutes, etc.)  
- Active for X amount of time  
- Maximum active time  
- Minimum active time  
- Certificate expiration date  

With these data points, averages could also be generated, as already available for daily, weekly, and monthly periods. A fourth and final option could calculate averages from the date the site was added up to the date the report is generated.

You could use the style of this site but add the features mentioned earlier. I believe this would make it much more complete. You could even use this idea and set it as the default option. However, an HTML report with charts and numerical data would be perfect.

![Image](https://github.com/user-attachments/assets/a2451b76-5ba4-4ed8-b4df-764899f52c42)

DMJain:
I would like help developing this issue. will submit a first draft of approach and downloaded report (in HTML). From there we can confirm what details needs to be modify.

gorkem-bwl:
Yes, let's do a very minimal (3-4 point) featureset and then we go from there once it's reviewed. Basic idea is that, there will be a button (location TBD) and when the admin clicks on it, he'll see an HTML or PDF version of the report.

DMJain:
First Draft:

![Image](https://github.com/user-attachments/assets/5514649f-f762-47a0-8a7c-b05d51808a61)

We can download report in pdf or html. User can choose report file type with dropdown. Currently, the position of download button is beside configure button. Report is generated based on selected date range. 

Report contains: Active time, Last Checked, Last Response Time, Avg Response Time, Uptime, DownTime and History.

There is issue with theme for report, trying to fix it that's why it took time for First Draft. still working on it.

[...]

PR #1743, opened 2025-02-11
#1588 - Download Report Functionality using existing components

DMJain:
## Describe your changes

Added new functionality Download report in PDF/HTML format using existing components

This functionality will allow user to download the report in pdf or html format with necessary stats for it's monitor.

## Issue number
#1588 

Mention the issue number(s) this PR addresses (e.g., #123).

- [ ✅] (Do not skip this or your PR will be closed) I deployed the application locally.
- [✅ ] (Do not skip this or your PR will be closed) I have performed a self-review and testing of my code.
- [ ✅] I have included the issue # in the PR.
- [✅ ] I have labelled the PR correctly.
- [ ✅] The issue I am working on is assigned to me.
- [ ✅] I didn't use any hardcoded values (otherwise it will not scale, and will make it difficult to maintain consistency across the application).
- [ ✅] I made sure font sizes, color choices etc are all referenced from the theme.
- [ ✅] My PR is granular and targeted to one specific feature.
- [ ✅] I took a screenshot or a video and attached to this PR if there is a UI change.

PDF, Images and Video of the Feature.

[healthCheck Personal Portfolio-monitor-report-2025-02-11 (5).pdf](https://github.com/user-attachments/files/18757489/healthCheck.Personal.Portfolio-monitor-report-2025-02-11.5.pdf)
Screenshot 2025-02-12 003830

https://github.com/user-attachments/assets/a0d6f1ec-8b8b-49a0-9aa4-d2ba1ab5de14


Client/src/Components/MonitorStatusHeader/index.jsx
```diff
@@ -5,10 +5,11 @@ import { useTheme } from "@emotion/react";
 import useUtils from "../../Pages/Uptime/Monitors/Hooks/useUtils";
 import { formatDurationRounded } from "../../Utils/timeUtils";
 import ConfigButton from "./ConfigButton";
+import ReportDownloadButton from "./ReportDownloadButton";
 import SkeletonLayout from "./skeleton";
 import PropTypes from "prop-types";
 
-const MonitorStatusHeader = ({ path, shouldRender = true, isAdmin, monitor }) => {
+const MonitorStatusHeader = ({ path, shouldRender = true, isAdmin, monitor, certificateExpiry, reportData }) => {
```
ajhollid, line 12: This component is getting too bloated, lets split it up in to variants.

Let's have one Header that has a config button only and one that has a ReportDownloadButton as well.

A variant might look like this:

```
export const HeaderWithReport = ({ reportData }) => {
	return (
		<MonitorStatusHeade>
			<ReportDownloadButton
				shouldRender={isAdmin}
				monitor={reportData}
				certificateExpiry={certificateExpiry}
			/>
		</MonitorStatusHeade>
	);
};
```
That will keep things nice and clean

Client/src/Pages/Uptime/Details/index.jsx
```diff
@@ -69,6 +70,12 @@ const UptimeDetails = () => {
 		rowsPerPage,
 	});
 
+	const [reportData, reportIsLoading, reportNetworkError] = useReportFetch({
+		authToken,
+		monitorId,
+		dateRange,
+	});
```
ajhollid, line 77: This will be called even if the user doesn't want to download the reports.  We should fetch this data on demand only if the user wants a report

Client/src/Pages/Uptime/Details/Hooks/useReportFetch.jsx
```diff
@@ -0,0 +1,34 @@
...
+	useEffect(() => {
+		const fetchMonitors = async () => {
+			try {
+				const res = await networkService.getStatsByMonitorId({
+					authToken: authToken,
+					monitorId: monitorId,
+					dateRange: dateRange,
+					normalize: true,
+				});
+				setReportData(res?.data?.data ?? {});
```
ajhollid, line 21: Please use `undefined` as a fallback value, this is by convention our state for empty data.

Client/src/Utils/Report/downloadReport.jsx
```diff
@@ -0,0 +1,65 @@
+import { createRoot } from "react-dom/client";
+import { ProductReport } from "./report";
+import html2pdf from "html2pdf.js";
+
+export const downloadReport = ({ monitorData, downloadFormat, certificateExpiry }) => {
+	// Create a temporary div to render the report
+	const tempDiv = document.createElement("div");
+	tempDiv.style.position = "absolute";
+	tempDiv.style.left = "-9999px";
+	document.body.appendChild(tempDiv);
```
ajhollid, line 10: Why are we doing this?  Direct DOM manipulation is a nono in general with React.

Client/src/Utils/Report/downloadReport.jsx
```diff
@@ -0,0 +1,65 @@
...
+
+	// Create a promise to handle the rendering and download
+	return new Promise((resolve) => {
```
ajhollid, line 13: Any particular reason to not use `async/await` pattern?

Client/src/Utils/Report/downloadReport.jsx
```diff
@@ -0,0 +1,65 @@
...
+		const reportRoot = createRoot(tempDiv);
+		reportRoot.render(
+			<ProductReport
+				monitorData={monitorData}
+				certificateExpiry={certificateExpiry}
+			/>
+		);
+
+		// Wait for styles to be applied
+		setTimeout(() => {
```
ajhollid, line 24: Can you explain why a timeout is needed here?  I see in your comment you need to wait for styles to be applied, but why is that?

Client/src/Utils/Report/downloadReport.jsx
```diff
@@ -0,0 +1,65 @@
...
+						// Create and trigger PDF download
+						html2pdf().from(htmlContent).save(`${monitorData.name}-monitor-report-${new Date().toISOString().split("T")[0]}.pdf`);
+					}
+			} finally {
+				// Always cleanup
+				reportRoot.unmount();
+				document.body.removeChild(tempDiv);
+				resolve();
+			}
+		}, 100);
```
ajhollid, line 63: Why 100ms for timeout?

Client/src/Utils/Report/reportStyles.css
```diff
@@ -0,0 +1,20 @@
...
+
+.info-title {
+    font-weight:600;
+}
+
+hr{
+    border: 0; 
+    border-top: 1px solid #ddd;
+    width: 700px;
+}
```
ajhollid, line 20: No css please, we use inline styles if absolultely necessary and more importantly the MUI theme for styling.

Client/src/Utils/Report/reportUtils.js
```diff
@@ -0,0 +1,71 @@
...
+        durations.push(duration);
+    }
+
+    return durations.length
+        ? {
+                min: Math.min(...durations),
+                max: Math.max(...durations),
+            }
+        : { min: 0, max: 0 };
+};
```
ajhollid, line 71: Why are we doing these calculations?

If you need some specific metrics they should be calculated on the backend, almost certainly at the database level.

The frontend should not be doing number crunching.

ajhollid (changes requested):
Hi @DMJain , thanks for contributing to the project!

There are some minor changes that can be made to reduce some front end complexity, but my main concern with this PR is that the calculation of stats for the report are done on the frontend.

This is something that should be handled on the backend, preferrably at the database layer.  MongoDBs aggregation pipeline should be sufficient to build whatever data you need.

Client/src/Utils/Report/downloadReport.jsx
DMJain: I tried without setTimeout and frontend calculation, for some reason blank html/pdf report were getting download. not able to figure out why and waiting even 50ms was helping. so kept setTimeout

Client/src/Utils/Report/downloadReport.jsx
DMJain: 0ms also works, but for, but without setTimeout not able to download reports with detail. it is kept because of how react renders and browser engine manipulate DOM. these are asynchronous tasks, so puting the download function under setTimeout triggers it after render happens successfully by browser/react even if i create Element in detached mode and not add it to actual Document. So will be using 0ms and will put comment for others to understand.

Client/src/Utils/Report/downloadReport.jsx
DMJain: I found the reason. setTimeout is used because of how react renders document and how browser engine renders the DOM. it's not instantaneous and synchronous. using setTimeout with 0ms also works as it let react and browser engine to complete these asynchronous task before executing our download function.

'//Wait for Styles to be applied' is wrong comment and I will remove it. I was trying to use theme previously and it was not working without setTimeOut. so i thought react is taking time to apply the styles. it's mistake from my side. but we still need setTimeout.

Client/src/Utils/Report/reportUtils.js
DMJain: Currently, these values are not being provided by backend. 
There were 2 options:
1. to calculate the values from frontend with existing APIs (quick, not preferred by me)
2. create a new API endpoint for these data to be included or to modify API for getting monitor Stats controller-service to get all data.

i was not sure if second would be approved as it requires modifying returned data from existing API or to create new API. so i went for frontend calculation. for the first version of the Feature. for future version we can include this.

Client/src/Utils/Report/downloadReport.jsx
DMJain: removed it. now not appending tempDiv to document. but still creating a element for HTML report. now it will be detached from actual document.

DMJain:
@ajhollid can you verify all the changes made to the PR.

ajhollid:
Hi @DMJain ,

After further reviewing this PR I believe that report generation is best done entirely on the backend, I can't think of any good reason to build reports on the front end of the application.  This is a data intense application and we do not want to bog down the front end doing work that should be done on the backend.

Aside from that, there are many hardcoded absolute dimension values used for styling.  Using hardcoded values, especially absolute dimenisons, is not compatible with our project policy.

I appreciate your efforts here, if you'd like to continute working on this one we can open a new issue/PR and plan out how this should be implemented.  I'm happy to go over that with you if you like!

DMJain:
I would like to continue working on this feature effectively.  Let's do it correctly.

Moving forward, what approach do you suggest we take: creating a new API endpoint for report data or modifying the current one?

For front end, another approach I considered was to include the report component within the Detail page but not display it directly. {display: none} This would allow the report component to render within the ThemeProvider and inherit the app-wide theme.

There are lots of ways to achieve this functionality. can we connect and discuss the approach so that we can have a clear direction moving forward?

DMJain:
@ajhollid @gorkem-bwl I would like to continue working on this feature effectively. Let's do it correctly.

Moving forward, what approach do you suggest we take: creating a new API endpoint for report data or modifying the current one?

For front end, another approach I considered was to include the report component within the Detail page but not display it directly. {display: none} This would allow the report component to render within the ThemeProvider and inherit the app-wide theme.

There are lots of ways to achieve this functionality. can we connect and discuss the approach so that we can have a clear direction moving forward?

ajhollid:
> @ajhollid @gorkem-bwl I would like to continue working on this feature effectively. Let's do it correctly.
> 
> Moving forward, what approach do you suggest we take: creating a new API endpoint for report data or modifying the current one?
> 
> For front end, another approach I considered was to include the report component within the Detail page but not display it directly. {display: none} This would allow the report component to render within the ThemeProvider and inherit the app-wide theme.
> 
> There are lots of ways to achieve this functionality. can we connect and discuss the approach so that we can have a clear direction moving forward?

Hey @DMJain ,glad you're on board for this!

Let's create a new endpoint, endpoints are free 😂

Let's set it up at /api/v1/reports/<whatever_you_need>

You can create a router, controller, and db module for this purpose, please follow the general patten of the other controllers / modules / routers.

You will probably need a service to generate the reports, so you can have a look at how other services are set up and registered on the backend.

If you can open up a draft PR and push to it incrementally we can follow along with your progress.

Please try to keep library use to the minimum necessary and check dependencies to see if we have what you need there already. 

Thanks again for your contributions!  Feel free to post any time you'd like to consult.

PR #1754, opened 2025-02-13
Implement json validation UI

YinDongFang:
## Describe your changes

- (Client side) Implement json validation ui in create/configure Uptime
- (Server side) Support empty string value for `expectedValue` and `jsonPath`, so that user can remove json validation by editing them to an empty value.

### Screenshot

image

## Issue number

#1573 

### DIFF
--- a/Client/src/Pages/Uptime/Configure/index.jsx
+++ b/Client/src/Pages/Uptime/Configure/index.jsx
@@ -436,6 +436,69 @@ const Configure = () => {
 									onChange={(event) => handleChange(event, "interval")}
 									items={frequencies}
 								/>
+								{
+									monitor.type === "http" && <>
+										<Select
+											id="match-method"
+											label="Match Method"
+											value={monitor.matchMethod || "equal"}
+											onChange={(event) => handleChange(event, "matchMethod")}
+											items={[
+												{ _id: "equal", name: "Equal" },
+												{ _id: "include", name: "Include" },
+												{ _id: "regex", name: "Regex" },
+											]}
+										/>
+										<Stack>
+											<TextInput
+												type="text"
+												id="expected-value"
+												label="Expected value"
+												isOptional={true}
+												placeholder={{
+													regex: "^[\w.-]+@gmail.com$",
+													equal: "janet@gmail.com",
+													include: "@gmail.com"
+												}[monitor.matchMethod || "equal"]}
+												value={monitor.expectedValue}
+												onChange={(event) => handleChange(event, "expectedValue")}
+												error={errors["expectedValue"] ? true : false}
+												helperText={errors["expectedValue"]}
+											/>
+											<Typography
+												component="span"
+												color={theme.palette.primary.contrastTextTertiary}
+												opacity={0.8}
+											>
+												The expected value is used to match the response result, and the result determines the status.
+											</Typography>
+										</Stack>
+										<Stack>
+											<TextInput
+												type="text"
+												id="json-path"
+												label="JSON Path"
+												isOptional={true}
+												placeholder="data.email"
+												value={monitor.jsonPath}
+												onChange={(event) => handleChange(event, "jsonPath")}
+												error={errors["jsonPath"] ? true : false}
+												helperText={errors["jsonPath"]}
+											/>
+											<Typography
+												component="span"
+												color={theme.palette.primary.contrastTextTertiary}
+												opacity={0.8}
+											>
+												The expression is evaluated against the reponse JSON data and the result will get used to match with expected value. Check out&nbsp;
+												<Typography component="a" href="https://jmespath.org/" target="_blank" color="info">
+													jmespath.org
+												</Typography>
+												&nbsp;for the documentation about the query language.
+											</Typography>
+										</Stack>
+									</>
+								}
 							</Stack>
 						</ConfigBox>
 						<Stack
--- a/Client/src/Pages/Uptime/Configure/index.jsx
+++ b/Client/src/Pages/Uptime/Configure/index.jsx
@@ -62,6 +62,18 @@ const Configure = () => {
 		"notify-email-default": "notification-email",
 	};
 
+	const matchMethodOptions = [
+		{ _id: "equal", name: "Equal" },
+		{ _id: "include", name: "Include" },
+		{ _id: "regex", name: "Regex" },
+	];
+
+	const expectedValuePlaceholders = {
+		regex: "^[\w.-]+@gmail.com$",
+		equal: "janet@gmail.com",
+		include: "@gmail.com"
+	};
+
 	useEffect(() => {
 		const fetchMonitor = async () => {
 			try {
@@ -443,11 +455,7 @@ const Configure = () => {
 											label="Match Method"
 											value={monitor.matchMethod || "equal"}
 											onChange={(event) => handleChange(event, "matchMethod")}
-											items={[
-												{ _id: "equal", name: "Equal" },
-												{ _id: "include", name: "Include" },
-												{ _id: "regex", name: "Regex" },
-											]}
+											items={matchMethodOptions}
 										/>
 										<Stack>
 											<TextInput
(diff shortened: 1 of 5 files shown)


gorkem-bwl:
Looks rad to me. Two questions:

- Does this feature need to be enabled to function? What if I do not fill in any of the fields? 
- Would it be better to make matching a separate settings block?

YinDongFang:
> * Does this feature need to be enabled to function? What if I do not fill in any of the fields?

If don't fill expected value, it will handled as normal http monitor.
If fill expected value but no json path, it will match response data as plain string.

> * Would it be better to make matching a separate settings block?

I think it will be. A new block after `Advanced Setting`? A descriptive text can be added to provide a detailed introduction of the features (I think the logic might not be very clear for user). But my english is not very good, and it's somewhat difficult for me to describe this logic in appropriate brief terms. It would be greatly appreciated if you could provide the description text.

gorkem-bwl:
> > * Does this feature need to be enabled to function? What if I do not fill in any of the fields?
> 
> If don't fill expected value, it will handled as normal http monitor. If fill expected value but no json path, it will match response data as plain string.

That is fine with me.
 
> > * Would it be better to make matching a separate settings block?
> 
> I think it will be. A new block after `Advanced Setting`? A descriptive text can be added to provide a detailed introduction of the features (I think the logic might not be very clear for user). But my english is not very good, and it's somewhat difficult for me to describe this logic in appropriate brief terms. It would be greatly appreciated if you could provide the description text.

Rethinking again, you are right - currently there is no need for a change, and it can be under advanced settings - I'm fine with that.

I don't see a problem with your English - everything is pretty clear to me. For best results, you may want to use AI to translate, but I don't see a need for it. 

We can always refine sentences in the UI when necessary.

Client/src/Pages/Uptime/Configure/index.jsx
```diff
@@ -436,6 +436,69 @@ const Configure = () => {
...
+									monitor.type === "http" && <>
+										<Select
+											id="match-method"
+											label="Match Method"
+											value={monitor.matchMethod || "equal"}
+											onChange={(event) => handleChange(event, "matchMethod")}
+											items={[
+												{ _id: "equal", name: "Equal" },
+												{ _id: "include", name: "Include" },
+												{ _id: "regex", name: "Regex" },
```
ajhollid, line 449: Pleaes refactor these to a constant, no reason for this to be inline

Client/src/Pages/Uptime/Configure/index.jsx
```diff
@@ -436,6 +436,69 @@ const Configure = () => {
...
+											<TextInput
+												type="text"
+												id="expected-value"
+												label="Expected value"
+												isOptional={true}
+												placeholder={{
+													regex: "^[\w.-]+@gmail.com$",
+													equal: "janet@gmail.com",
+													include: "@gmail.com"
+												}[monitor.matchMethod || "equal"]}
```
ajhollid, line 462: Let's refactor out this placeholder map as well

Client/src/Pages/Uptime/Configure/index.jsx
```diff
@@ -436,6 +436,69 @@ const Configure = () => {
...
+												onChange={(event) => handleChange(event, "jsonPath")}
+												error={errors["jsonPath"] ? true : false}
+												helperText={errors["jsonPath"]}
+											/>
+											<Typography
+												component="span"
+												color={theme.palette.primary.contrastTextTertiary}
+												opacity={0.8}
+											>
+												The expression is evaluated against the reponse JSON data and the result will get used to match with expected value. Check out&nbsp;
```
ajhollid, line 493: Please change to:

>This expression will be evaluated against the reponse JSON data and the result will be used to match against the expected value. See&nbsp;

Client/src/Pages/Uptime/Configure/index.jsx
```diff
@@ -436,6 +436,69 @@ const Configure = () => {
...
+												<Typography component="a" href="https://jmespath.org/" target="_blank" color="info">
+													jmespath.org
+												</Typography>
+												&nbsp;for the documentation about the query language.
```
ajhollid, line 497: And this to:

> for query language documentation.

Client/src/Pages/Uptime/Configure/index.jsx
```diff
@@ -436,6 +436,69 @@ const Configure = () => {
...
+												onChange={(event) => handleChange(event, "expectedValue")}
+												error={errors["expectedValue"] ? true : false}
+												helperText={errors["expectedValue"]}
+											/>
+											<Typography
+												component="span"
+												color={theme.palette.primary.contrastTextTertiary}
+												opacity={0.8}
+											>
+												The expected value is used to match the response result, and the result determines the status.
```
ajhollid, line 473: Please change to:

> The expected value is used to match against response result, and the match determines the status.

Client/src/Pages/Uptime/Create/index.jsx
```diff
@@ -399,6 +405,69 @@ const CreateMonitor = () => {
...
+							monitor.type === "http" && <>
+								<Select
+									id="match-method"
+									label="Match Method"
+									value={monitor.matchMethod || "equal"}
+									onChange={(event) => handleChange(event, "matchMethod")}
+									items={[
+										{ _id: "equal", name: "Equal" },
+										{ _id: "include", name: "Include" },
+										{ _id: "regex", name: "Regex" },
```
ajhollid, line 418: This can be extracted as well

Client/src/Pages/Uptime/Create/index.jsx
```diff
@@ -399,6 +405,69 @@ const CreateMonitor = () => {
...
+									<TextInput
+										type="text"
+										id="expected-value"
+										label="Expected value"
+										isOptional={true}
+										placeholder={{
+											regex: "^[\w.-]+@gmail.com$",
+											equal: "janet@gmail.com",
+											include: "@gmail.com"
+										}[monitor.matchMethod || "equal"]}
```
ajhollid, line 431: Same with this

Client/src/Pages/Uptime/Create/index.jsx
```diff
@@ -399,6 +405,69 @@ const CreateMonitor = () => {
...
+									<Typography
+										component="span"
+										color={theme.palette.primary.contrastTextTertiary}
+										opacity={0.8}
+									>
+										The expression is evaluated against the reponse JSON data and the result will get used to match with expected value. Check out&nbsp;
+										<Typography component="a" href="https://jmespath.org/" target="_blank" color="info">
+											jmespath.org
+										</Typography>
+										&nbsp;for the documentation about the query language.
```
ajhollid, line 466: Please edit to match strings described above

ajhollid (changes requested):
Looking good!  Some minor edits to strings and we can extract some constant values.

Thank you for your contribution!

Client/src/Pages/Uptime/Configure/index.jsx
```diff
@@ -436,6 +436,69 @@ const Configure = () => {
...
+												error={errors["jsonPath"] ? true : false}
+												helperText={errors["jsonPath"]}
+											/>
+											<Typography
+												component="span"
+												color={theme.palette.primary.contrastTextTertiary}
+												opacity={0.8}
+											>
+												The expression is evaluated against the reponse JSON data and the result will get used to match with expected value. Check out&nbsp;
+												<Typography component="a" href="https://jmespath.org/" target="_blank" color="info">
```
marcelluscaio, line 494: What is this color info? Doesn't seem to be part of the theme

YinDongFang: Default mui theme color. I see it is used in `Image` component. 

image

marcelluscaio: Oh I see, thanks for the clarification. 

I am not sure how that contrast works on both light and dark theme. We have some colors in our palette, and all written text must be some inside of the contrastText category - which have been tested for contrast.

@ajhollid I say this should generate another issue, and I can tackle that, since it's out of scope for the issue this PR is solving, do you agree?

ajhollid: @marcelluscaio yep, that sounds like a good plan to me

ajhollid (approved):
Looks good to me, thank you for your contribution!

Issue #1772, opened 2025-02-14
UI Bug: Maintenance Tab Background Appears Clipped


vishnusn77:
When navigating to the Maintenance tab in the dashboard, the background does not fully extend to cover the entire screen, making it appear cut short. This issue is inconsistent with the design seen in other tabs.

Expected Behavior:
The background should cover the full height of the page, maintaining design consistency across all tabs.

![Image](https://github.com/user-attachments/assets/c08a933f-c006-4432-beb3-79bbd25ed023)

PR #1783, opened 2025-02-17
Fix Maintenance Page Layout issue

vishnusn77:
## Describe your changes

Moved Fallback outside the <Box> container
Before, Fallback was inside .maintenance.table-container, inheriting unwanted styles. Now, Fallback is returned directly 

## Issue number

#1772 

Before:
Screenshot 2025-02-14 155252

After:
Screenshot 2025-02-17 130110

### DIFF
--- a/Client/src/Pages/Maintenance/index.jsx
+++ b/Client/src/Pages/Maintenance/index.jsx
@@ -11,100 +11,100 @@ import { useNavigate } from "react-router-dom";
 import { useIsAdmin } from "../../Hooks/useIsAdmin";
 
 const Maintenance = () => {
-	const theme = useTheme();
-	const navigate = useNavigate();
-	const { authToken } = useSelector((state) => state.auth);
-	const { rowsPerPage } = useSelector((state) => state.ui.maintenance);
-	const isAdmin = useIsAdmin();
-	const [maintenanceWindows, setMaintenanceWindows] = useState([]);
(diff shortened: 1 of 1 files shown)


gorkem-bwl:
UI wise looks good to me!

Client/src/Pages/Maintenance/index.jsx
```diff
@@ -44,65 +44,67 @@ const Maintenance = () => {
...
+					"Stop sending alerts in maintenance windows",
+				]}
+				link="/maintenance/create"
+				isAdmin={isAdmin}
+			/>
+		);
+	}
+
 	return (
 		<Box
```
ajhollid, line 64: Is this box even needed anymore?  I don't believe it is as these styles are now defined in the Fallback component:

```
			position="relative"
			border={1}
			borderColor={theme.palette.primary.lowContrast}
			borderRadius={theme.shape.borderRadius}
			backgroundColor={theme.palette.primary.main}
			overflow="hidden"
			sx={{
				borderStyle: "dashed",
			}}
```

ajhollid (changes requested):
I think we can remove that Box container, it was used to create the dashed border around the fallback, but that's no longer used as far as I can tell.  Please confirm and if not needed we can remove it.

Client/src/Pages/Maintenance/index.jsx
vishnusn77: you're right! Got rid of the box.

PR #1791, opened 2025-02-19
proposal: image-upload component

ajhollid:
This is a propsoal for an image-upload component.

I've tried to keep it as simple as possible while fulfiling the design requirements and following MUI's suggestions for uploading files.

Please leave your thoughts and we can see if this is a good approach.

### DIFF
--- a/src/Components/Inputs/ImageUpload/index.jsx
+++ b/src/Components/Inputs/ImageUpload/index.jsx
@@ -1,0 +1,176 @@
+// Components
+import { Button, Box, Stack, Typography } from "@mui/material";
+import CloudUploadIcon from "@mui/icons-material/CloudUpload";
+import Image from "../../../Components/Image";
+
+// Utils
+import PropTypes from "prop-types";
+import { createToast } from "../../../Utils/toastUtils";
+import { formatBytes } from "../../../Utils/fileUtils";
+import { useCallback } from "react";
+import { useTheme } from "@emotion/react";
+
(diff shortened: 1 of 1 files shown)


vishnusn77 (commented):
Looks well structured and easy to follow. 

I noticed that the existing Image component is used here, which handles base64 validation, placeholder logic. 
So are we integrating that into this new component as well or leaving it as it is?

ajhollid:
> Looks well structured and easy to follow.
> 
> I noticed that the existing Image component is used here, which handles base64 validation, placeholder logic. So are we integrating that into this new component as well or leaving it as it is?

I think it makes sense to just use the Image component, no need to build a new one I don't think?

The original component doesn't work reliably, sometimes the preview image is loaded, sometimes it isn't.  I don't know why and I don't think it's worth the effort trying to find out :joy:

vishnusn77:
> > Looks well structured and easy to follow.
> > I noticed that the existing Image component is used here, which handles base64 validation, placeholder logic. So are we integrating that into this new component as well or leaving it as it is?
> 
> I think it makes sense to just use the Image component, no need to build a new one I don't think?
> 
> The original ImageUpload component doesn't work reliably, sometimes the preview image is loaded, sometimes it isn't. I don't know why and I don't think it's worth the effort trying to find out 😂

That makes sense. Think we can proceed with this. I can help with refining it/refactor the current code. Let me know how to proceed.

PR #1838, opened 2025-02-27
feat: status table onclick

ajhollid:
This PR adds a configurable `onClick` listener and styling for specifc cells in the `DataTable` component.

It adds an `onClick` listener to the status page table that opens the public status page if it is published.

image


ajhollid:
> Working as expected 👍 suggestion : public URL tablecell values could be centered to maintain consistency with other table cells. right now, it's left aligned. Screenshot 2025-02-27 161145

Perhaps this should be made configurable on a per table basis?

I believe it is not centered now as another table required this specific alignment.

vishnusn77 (approved):
Looks good 👍

PR #1839, opened 2025-02-27
feat: set monitor name instead of placeholder data

ajhollid:
This PR replaces placeholder data with correct data


vishnusn77 (approved):
Looks good!

Issue #1729, opened 2025-02-09
FE - Status Page graph not displaying well on high resolution screens


ajhollid:
On the status page graph bars do not have consistent gap spacing, which leads to issues on high resolution screens:

Expected:
![Image](https://github.com/user-attachments/assets/5113f12a-0ee6-4990-9fd9-71226e190433)

Actual: 
![Image](https://github.com/user-attachments/assets/5f1e1960-5b9a-4fa0-be41-9cafa3d93673)

gorkem-bwl:
If anyone would like to work on this, I'd be happy to provide some additional (minor) fix requests for the status page.

Br0wnHammer:
@gorkem-bwl can I take this on?

FlowerC9:
i can fix this could you please assign this to me?

gorkem-bwl:
@FlowerC9 currently I don't have a design for this as we're quite focused on the release issues. Do you have something in mind?

FlowerC9:
can you @ajhollid  please tell the steps to recreate this i have tried but it seems bar graphs is fine?

ajhollid:
@ajhollid Hi @FlowerC9 , I noticed the issue while viewing at 4k resolution.

Are you able to check at that resolution?

FlowerC9:
i not able to check in 4k @ajhollid .can you @gorkem-bwl please unassign me from this issue since i don't have any 4k medium so i'm not able to check is it fixed or not.

gorkem-bwl:
Sure, thanks for noting down.

Owaiseimdad:
@gorkem-bwl , I think I have resolution settings on my mac.

<img width="517" alt="Image" src="https://github.com/user-attachments/assets/6c206ed8-f873-4a41-ab69-3dd84ce6f3fc" />

Assign me this if I am not wrong.
@ajhollid .

gorkem-bwl:
You should be good to go @Owaiseimdad !

Owaiseimdad:
@ajhollid possible to help me in setup for checkmate?
I am following, https://docs.checkmate.so/users-guide/quickstart but to be honest very confusing.

-> When you tell Change directory to the Docker/dev directory which directory are you trying to change?
-> After I cloned the repo, I did not find any build_images.sh file?

I some how understood something so started with https://docs.checkmate.so/users-guide/quickstart#manual-install 

Also, I don’t see any server-client repo at all. What am I missing here?

[...]

PR #1845, opened 2025-03-01
Added a new HOC in status bar in order to make sure it works on both …

Owaiseimdad:
…the resolution, high and low.
## Describe your changes
Changes done in status bar in order to accommodate the resolution issue with styles in width of single bars.

## Issue number
FE - Status Page graph not displaying well on high resolution screens #1729

- [x ] (Do not skip this or your PR will be closed) I deployed the application locally.
- [x ] (Do not skip this or your PR will be closed) I have performed a self-review and testing of my code.
- [x ] I have included the issue # in the PR.
- [x ] I have labelled the PR correctly.
- [x ] The issue I am working on is assigned to me.
- [x ] I didn't use any hardcoded values (otherwise it will not scale, and will make it difficult to maintain consistency across the application).
- [x ] I made sure font sizes, color choices etc are all referenced from the theme.
- [x ] My PR is granular and targeted to one specific feature.
- [x ] I took a screenshot or a video and attached to this PR if there is a UI change.
- Before
Screenshot 2025-02-28 at 9 22 42 AM

- After
Screenshot 2025-02-28 at 9 22 54 AM
Screenshot 2025-02-28 at 9 33 06 AM
Screenshot 2025-02-28 at 9 33 19 AM

### DIFF
--- a/src/Components/Charts/StatusPageBarChart/index.jsx
+++ b/src/Components/Charts/StatusPageBarChart/index.jsx
@@ -24,6 +24,21 @@ const StatusPageBarChart = ({ checks = [] }) => {
 		checks = [...checks, ...placeholders];
 	}
 
+	const BarBox = ({ width, height, backgroundColor, borderRadius, children }) => (
+		<Box
+			marginRight="5px"
+			position="relative"
+			width={width}
+			height={height}
+			backgroundColor={backgroundColor}
+			sx={{
+				borderRadius: borderRadius || theme.spacing(1.5),
+			}}
+		>
+			{children}
+		</Box>
+	);
+
 	return (
 		<Stack
 			direction="row"
@@ -41,15 +56,11 @@ const StatusPageBarChart = ({ checks = [] }) => {
 					/* TODO what is the purpose of this box? 	*/
 					// CAIO_REVIEW the purpose of this box is to make sure there are always at least 25 bars
 					// even if there are less than 25 checks
-					<Box
+					<BarBox
 						key={`${check}-${index}`}
-						position="relative"
-						width="calc(30vw / 25)"
+						width="calc(30rem / 25)"
 						height="100%"
 						backgroundColor={theme.palette.primary.lowContrast}
-						sx={{
-							borderRadius: theme.spacing(1.5),
-						}}
 					/>
 				) : (
 					<Tooltip
@@ -140,14 +151,10 @@ const StatusPageBarChart = ({ checks = [] }) => {
 							},
 						}}
 					>
-						<Box
-							position="relative"
-							width="calc(30vw / 25)"
+						<BarBox
+							width="calc(30rem / 25)"
 							height="100%"
-							backgroundColor={theme.palette.primary.lowContrast} // CAIO_REVIEW
-							sx={{
-								borderRadius: theme.spacing(1.5),
-							}}
+							backgroundColor={theme.palette.primary.lowContrast}
 						>
 							<Box
 								position="absolute"
@@ -164,7 +171,7 @@ const StatusPageBarChart = ({ checks = [] }) => {
 									transition: "height 600ms cubic-bezier(0.4, 0, 0.2, 1)",
 								}}
 							/>
-						</Box>
+						</BarBox>
 					</Tooltip>
 				)
 			)}


Owaiseimdad:
cc: @gorkemcetin @ajhollid The PR for status bars

gorkem-bwl:
UI wise looks good to me. We'll review the code and get back to you. Thanks for this.

src/Components/Charts/StatusPageBarChart/index.jsx
```diff
@@ -41,15 +56,11 @@ const StatusPageBarChart = ({ checks = [] }) => {
 					/* TODO what is the purpose of this box? 	*/
 					// CAIO_REVIEW the purpose of this box is to make sure there are always at least 25 bars
 					// even if there are less than 25 checks
-					<Box
+					<BarBox
```
ajhollid, line 59: Let's just call this `Bar` as that's what it is

src/Components/Charts/StatusPageBarChart/index.jsx
```diff
@@ -41,15 +56,11 @@ const StatusPageBarChart = ({ checks = [] }) => {
...
 						key={`${check}-${index}`}
-						position="relative"
-						width="calc(30vw / 25)"
+						width="calc(30rem / 25)"
 						height="100%"
 						backgroundColor={theme.palette.primary.lowContrast}
-						sx={{
-							borderRadius: theme.spacing(1.5),
-						}}
 					/>
```
ajhollid, line 64: We need prop types and JSdocs for this component

src/Components/Charts/StatusPageBarChart/index.jsx
ajhollid, line 59: This component should not be defiend within the parent component, please move it ouside of the `StatusPageBarChart`

src/Components/Charts/StatusPageBarChart/index.jsx
```diff
@@ -24,6 +24,21 @@ const StatusPageBarChart = ({ checks = [] }) => {
 		checks = [...checks, ...placeholders];
 	}
 
+	const BarBox = ({ width, height, backgroundColor, borderRadius, children }) => (
+		<Box
+			marginRight="5px"
```
ajhollid, line 29: Why margin right?  Why 5px?  We don't accept hardcoded values unless there is a very specific reason

ajhollid (changes requested):
There are some issues to be looked at, and the public status page doesn't really seem improved in my opinion:

This PR:

image

vs original:

image

The bars look better, but there's far too much spacing between them.

vishnusn77 (commented):
UI looks better.  As Alex suggested, BarBox Should Be a Separate Component, you can move it outside the main function. Also, we need to add PropTypes & JSDoc for the component.

src/Components/Charts/StatusPageBarChart/index.jsx
Owaiseimdad: Implementing this.

Owaiseimdad: Implementing this.

src/Components/Charts/StatusPageBarChart/index.jsx
Owaiseimdad: makes sense @ajhollid , i should be avoiding it. If I cannot use margin then I have to change the UI such in such a way that we should not even depend on 30rem or anything.
I made a slight change, let me know if you are happy with UI.

https://github.com/user-attachments/assets/20a2d3b6-8eba-488f-927d-feae2fea7283
The bars will larger width but will never collide.

Owaiseimdad:
Have updated the PR @ajhollid and @vishnusn77 . Please let me know if there are anymore changes.

src/Components/Charts/Bar/index.jsx
```diff
@@ -0,0 +1,45 @@
...
+ * @param {string} width The width of the bar (e.g., "100px").
+ * @param {string} height The height of the bar (e.g., "50px").
+ * @param {string} backgroundColor The background color of the bar (e.g., "#FF5733").
+ * @param {string} [borderRadius] Optional border radius for the bar (e.g., "8px").
+ * @param {node} children The content to be rendered inside the bar.
+ *
+ * @returns {JSX.Element} The Bar component.
+ */
+
+const Bar = ({ width, height, backgroundColor, borderRadius, children }) => {
```
ajhollid, line 19: This doens't need to be in a separate file, you can leave it in the `StautsPageBarChart` component as it is only used there, it just should not be declared in the other componen'ts function.

src/Components/Charts/Bar/index.jsx
```diff
@@ -0,0 +1,45 @@
...
+		</Box>
+	);
+};
+
+Bar.propTypes = {
+	width: PropTypes.string.isRequired,
+	height: PropTypes.string.isRequired,
+	backgroundColor: PropTypes.string.isRequired,
+	borderRadius: PropTypes.string,
+	children: PropTypes.node,
```
ajhollid, line 42: :+1:

ajhollid (changes requested):
The bars are too wide now, we would like the original size to be maintianed for the non-public status page:

Original: 

image

Current:

image

image

Also the `Bar` component can stay in hte `StatusPageBarChart` component, it should be declared outside of the chart component though, not within it's funciton.

src/Components/Charts/Bar/index.jsx
Owaiseimdad: Okay, got it. So, basically it should be like this, 

const Bar = ({ children }) => {
	return (
		<Box>
			{children}
		</Box>
	);
};

const StatusPageBarChart = ({ checks = [] }) => {
	return (
		<Stack
		>
			{checks.map((check, index) =>
				check === "placeholder" ? (
					/* TODO what is the purpose of this box? 	*/
					// CAIO_REVIEW the purpose of this box is to make sure there are always at least 25 bars
					// even if there are less than 25 checks
					<Bar
						backgroundColor={theme.palette.primary.lowContrast}
					/>
				) .... all the other code.
)

export default StatusPageBarChart;

Let me know if I am wrong again here @ajhollid .

Owaiseimdad:
Yes, @ajhollid.

I have made to appear like the original size now.

Screenshots for reference.
Screenshot 2025-03-04 at 5 50 00 AM
Screenshot 2025-03-04 at 5 50 18 AM
Screenshot 2025-03-04 at 5 50 26 AM

src/Components/Charts/StatusPageBarChart/index.jsx
```diff
@@ -138,7 +176,7 @@ const StatusPageBarChart = ({ checks = [] }) => {
 						}}
 					>
 						<Bar
-							width="calc(85% / 25)"
+							width="calc(60% / 25)"
```
ajhollid, line 179: @Owaiseimdad this is looking better for the status page, but for public status page it's a bit too wide on large screens.

I think you're going to need a solution for different sized screens here.  There's also an issue where this value is duplicated requiring changes in two places in order to adjust this.

I propose we pass a config object to both components that lets us deal with different screen sizes:

```
	const barWidth = {
		xs: "calc(60% / 25)",
		xl: "calc(40% / 25)",
	};
```
You can pass that to the `Bar` component like so:

```
	<Bar
		key={`${check}-${index}`}
		width={barWidth}
		height="100%"
		backgroundColor={theme.palette.primary.lowContrast}
	/>
```
					
That kills two birds with one stone.  You can read more about this [in the docs](https://mui.com/system/getting-started/usage/#responsive-values)

ajhollid (changes requested):
I think we need a more responsive value here for the width of the `Bar` component, please see my comments and consider.  Thanks!

src/Components/Charts/StatusPageBarChart/index.jsx
Owaiseimdad: Thanks @ajhollid , looks like we can hit more birds with one stone if we introduce this styling method. Updating the PR.

Owaiseimdad:
@ajhollid public page, 

Before:
Screenshot 2025-03-04 at 3 52 47 PM

After:
Screenshot 2025-03-04 at 3 52 54 PM

ajhollid (approved):
Thanks for your contribution, looks much better now!

Issue #1847, opened 2025-03-01
The styles and title get distorted if the length of the Incident name becomes big


Owaiseimdad:
**Describe the bug**
In "Incident" page, if there is a incident with large string the other texts looks like they have come outside of the box.

**To Reproduce**
1. Go to "Incident" page
2. Click on "Incident for" and choose something which is big.
4. See the distorted in the image

**Expected behavior**
The text or the box should enlarge ideally and the incident should not be shown entirely if the name is big

**Screenshots**

<img width="1114" alt="Image" src="https://github.com/user-attachments/assets/e9f94bac-4eb0-4f88-98c4-c4bd767ca9fb" />
<img width="1140" alt="Image" src="https://github.com/user-attachments/assets/7ee49b40-e8a5-479b-b838-e656b3800a81" />

Owaiseimdad:
@ajhollid and @gorkemcetin looks like genius bug, please assign this to me if you this is not expected.

gorkem-bwl:
You are right. If it is big enough, and there is no room to extend it, we can just crop the name there and add "..." at the end of the name. Similar to this:

<img width="788" alt="Image" src="https://github.com/user-attachments/assets/9fcf8628-1830-440d-ab6a-321b4cc75b88" />

Owaiseimdad:
Assign this to me @gorkem-bwl

Owaiseimdad:
Working on this @gorkem-bwl , 
Updates:
1. Have added a logic in order to truncate the div if the length is more than 15 (why 15? Because I see if the length goes more than 15 then the UI looks a bit mess)
2. Also, made changes in the Styles of the Box components in order to make sure that they don’t overflow the text If zoom beyond lets say 150%.

https://github.com/user-attachments/assets/c8a8c7e4-ecbc-41da-a5ad-7fd43d40e812

gorkem-bwl:
Looks good to me.

Owaiseimdad:
@ajhollid as discussed will raise a new PR for truncate behaviour and we can validate the approach there.

Owaiseimdad:
@gorkem-bwl , the second issue is fixed which is distored frontend. Working on first issue now.

Owaiseimdad:
@gorkem-bwl and @ajhollid , pr raised for truncate of string separately.

Owaiseimdad:
@gorkem-bwl I think this issue has been addressed, we should be good to close this.

[...]

PR #1872, opened 2025-03-05
fix-1847 truncate fix for selection component in incident page (or in general)

Owaiseimdad:
## Describe your changes
Shrinking of the text if there are more than certain limit, and if there are more than a length add ... at end.

## Issue number
#1847 

https://github.com/user-attachments/assets/73a125ea-f8d6-4006-8f71-18f4e9563489


gorkem-bwl:
UI wise looks good to me!

Owaiseimdad:
@gorkem-bwl once this PR is reviewed lets implement the same in the other usecase we have for another repo.

src/Components/Inputs/Select/index.jsx
```diff
@@ -59,6 +66,29 @@ const Select = ({
...
 	};
 
+	// Dynamically adjust maxLength based on window width
+	useEffect(() => {
+		if (!truncate) return; // If truncate is false, do not run this effect
+
+		const calculateMaxLength = () => {
+			if (selectRef.current) {
+				const width = selectRef.current.offsetWidth;
+				const calculatedMaxLength = Math.floor(width / maxStringLength);
```
ajhollid, line 76: This calculation puzzles me, what does "maxStringLength" mean?  It's value is the integer 10, which has nothing to do with the length of a string.

What this really means is that the max width of a string is 1/10th of the screen width.  

This is misleading at best and a logical mistake at worst :thinking: 

I think even if the calculation made logical sense this is probaby overkill, I don't think we really need to dynamically adjust the truncation of the string.

I understand what you are doing here, but it seems like an excessively complex solution to a minor problem.

ajhollid (changes requested):
Hey @Owaiseimdad ,

I understand what you are trying to do with this PR, but I think it is an overly complex solution to a relatively simple problem.

My proposal is we follow the [KISS](https://en.wikipedia.org/wiki/KISS_principle) principle  and simply allow the passing of a `maxWidth` prop to the `Select` component and combine that with something really simple like this:

```
renderValue={(selected) => {
	return (
		<Typography
			noWrap
			sx={{
				overflow: "hidden",
				textOverflow: "ellipsis",
				whiteSpace: "nowrap",
			}}
		>
			{items.find((item) => item._id === selected)?.name}
		</Typography>
	);
}}
```

What do you think?

src/Components/Inputs/Select/index.jsx
Owaiseimdad: @ajhollid thanks for the feedback! The reason I added the responsiveness here was to address the potential zooming issue, where users might be viewing the app at different screen resolutions. By default, I set it to 1/10th of the element's width instead of the screen width, so it adapts to various zoom levels.

In my previous PR, there was no responsiveness at all, so if you think it's simpler to just have a default value and let developers adjust the length if needed, I'm open to that approach.

Let me know your thoughts—should we revert to the old implementation or keep the responsiveness? I agree that adding this responsiveness might be overcomplicating things.

Owaiseimdad:
Yes, @ajhollid you exactly right. I should thought of using a mui element instead of a span and used style in order to make is work.

<MuiSelect
				className="select-component"
				value={value}
				onChange={onChange}
				onBlur={onBlur}
				displayEmpty
				name={name}
				inputProps={{ id: id }}
				IconComponent={KeyboardArrowDownIcon}
				MenuProps={{ disableScrollLock: true }}
				sx={{
					fontSize: 13,
					minWidth: "125px",
					maxWidth: "250px",
					"& fieldset": {
						borderColor: theme.palette.primary.lowContrast,
					},
					"&:not(.Mui-focused):hover fieldset": {
						borderColor: theme.palette.primary.lowContrast,
					},
					"& svg path": {
						fill: theme.palette.primary.contrastTextTertiary,
					},
					...sx,
				}}
				renderValue={(selected) => {
					const selectedItem = items.find((item) => item._id === selected);
					const displayName = selectedItem ? selectedItem.name : placeholder;
					return (
						<Typography
							sx={{
								overflow: "hidden",
								textOverflow: "ellipsis",
								whiteSpace: "nowrap",
							}}
							title={displayName}
						>
							{displayName}
						</Typography>
					);
				}}
			>
			
			this works. But I will be using a constant value called maxWidth double of min which was defined earlier. 
			
			let me know what you think

ajhollid:
@Owaiseimdad why not have Select take a maxWidth prop?

Then you can adjust the width on a per component basis and also provide different widths for different screen sizes.

Owaiseimdad:
Great @ajhollid, that will be good. 

and with help of the value given we can just calculate 

maxWidth: {
                        xs: `${maxWidthValue * 0.5}px`,   
                        sm: `${maxWidthValue * 0.75}px`, 
                        md: `${maxWidthValue * 0.9}px`, 
                        lg: `${maxWidthValue}`,    // For large screens (default or passed as props)
                    }

This should be good. and then we can consume the maxWidthValue variable passed.

src/Components/Inputs/Select/index.jsx
```diff
@@ -52,12 +50,8 @@ const Select = ({
 	sx,
 	name = "",
 	labelControlSpacing = 2,
-	truncate = false,
-	maxStringLength = 10,
+	maxWidthValue = 250,
```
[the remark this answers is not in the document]
Owaiseimdad: This is updated now @ajhollid

Owaiseimdad:
Hi @ajhollid , we should good to give a final review for this pr.

ajhollid (approved):

ajhollid (approved):

PR #1895, opened 2025-03-09
fix: stringify token upon registration

renatoka:
## Describe your changes

Stringified "token" object that is saved in local storage, as it's currently stored as string.

## Issue number

https://github.com/bluewave-labs/Checkmate/issues/1894

- [ ] The issue I am working on is assigned to me.

### DIFF
--- a/src/Pages/Auth/Register/Register.jsx
+++ b/src/Pages/Auth/Register/Register.jsx
@@ -239,7 +239,6 @@ const Register = ({ isSuperAdmin }) => {
 		const action = await dispatch(register(registerForm));
 		if (action.payload.success) {
 			const authToken = action.payload.data;
-			localStorage.setItem("token", JSON.stringify(authToken));
 			navigate("/uptime");
 			createToast({
 				body: "Welcome! Your account was created successfully.",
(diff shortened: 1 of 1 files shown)


ajhollid:
This shouldn't even be in local storage, the token is stored in the redux store :thinking:   Not sure where this came from or what it's being used for.

renatoka:
In that case I would simply remove it as it's not being read anywhere else in the project, only being set upon register. I can see that we are also saving user in `authSlice.js` so it's accessible everywhere in the project. Good to remove it? @ajhollid

ajhollid:
> In that case I would simply remove it as it's not being read anywhere else in the project, only being set upon register. I can see that we are also saving user in `authSlice.js` so it's accessible everywhere in the project. Good to remove it? @ajhollid

I think so, have you found any usage of local storage in the application?  If not, then good to remove.

renatoka:
> > In that case I would simply remove it as it's not being read anywhere else in the project, only being set upon register. I can see that we are also saving user in `authSlice.js` so it's accessible everywhere in the project. Good to remove it? @ajhollid
> 
> I think so, have you found any usage of local storage in the application? If not, then good to remove.

Nothing like that, take a look. Pushing changes.

image

renatoka:
> Thanks for catching this 👍

No problem. Hoping to get more into contributing as much as possible besides main job.

Issue #1913, opened 2025-03-11
"Undefined" in Maintenance page drop down


gorkem-bwl:
When I choose a server inside the maintenance page, it prints out "Undefined" next to the server name.

<img width="728" alt="Image" src="https://github.com/user-attachments/assets/397df99e-5b8a-4c92-9b4f-3e4331f75ef7" />

Owaiseimdad:
@gorkem-bwl you can assign this to me as I am already working on it.

Owaiseimdad:
@gorkem-bwl looks like this requires changes in backend will raise the pr there.

Owaiseimdad:
@gorkem-bwl the issue is fixed in backend, ideally the check has to be there in frontend as well. Will make small tweaks in order to avoid showing undefined even if something wrong is sent in backend.

PR #1934, opened 2025-03-15
Fixed the check for the label in options object in search component

Owaiseimdad:
## Describe your changes

we had faced a issue where the backend api was not returning an object which consists of a key called "type" but instead the UI was defining it as undefined for end user (confusing for them)

## Issue number
#1913 

```Javascript
<div>{t('add')}</div>
```
- [ x I made sure font sizes, color choices etc are all referenced from the theme. I have no hardcoded dimensions.
- [] I took a screenshot or a video and attached to this PR if there is a UI change.

### DIFF
--- a/src/Components/Inputs/Search/index.jsx
+++ b/src/Components/Inputs/Search/index.jsx
@@ -154,6 +154,7 @@ const Search = ({
 			}}
 			renderOption={(props, option) => {
 				const { key, ...optionProps } = props;
+				const hasSecondaryLabel = secondaryLabel && option[secondaryLabel] !== undefined;
 				return (
 					<ListItem
 						key={key}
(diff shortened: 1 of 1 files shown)


Owaiseimdad:
@ajhollid PR to check whether label passed from parent exists in the options object.

ajhollid (changes requested):
It's good to do a check here, but lets adjust things a bit for clarity.

1.  Let's move the logic for checking if the `secondarLabel` is present or not up so that it is clear.

2.  Lets use strict equality `===` instead of truthy/falsey evaluation so we know exactly what values we are looking to filter out. 

```
	renderOption={(props, option) => {
          // Do checks here
          const { key, ...optionProps } = props;
		return (
			...
                 )
	}}
```

Owaiseimdad:
> It's good to do a check here, but lets adjust things a bit for clarity.
> 
> 1. Let's move the logic for checking if the `secondarLabel` is present or not up so that it is clear.
> 2. Lets use strict equality `===` instead of truthy/falsey evaluation so we know exactly what values we are looking to filter out.
> 
> ```
> 	renderOption={(props, option) => {
>           // Do checks here
>           const { key, ...optionProps } = props;
> 		return (
> 			...
>                  )
> 	}}
> ```

Got it @ajhollid , 

`const hasSecondaryLabel = secondaryLabel && option[secondaryLabel] !== undefined;`

and then use it below.

ajhollid (approved):
Thanks for making the requested changes, looks good to me.

Issue #1936, opened 2025-03-15
Cancel button on multiple popups have a dark BG.


Owaiseimdad:
**Describe the bug**
The cancel option on multiple popups are like link instead of buttons, need to be a like button for better UI.

**To Reproduce**
1. Go to settings page
2. Click on "clear all stats"
3. The popup appears

**Expected behaviour**
It should ideally be a button with gray or offset of gray colour.

**Screenshots**

<img width="464" alt="Image" src="https://github.com/user-attachments/assets/e4fd9039-5dfa-45d9-9154-6376d75a8e87" />

Owaiseimdad:
@gorkemcetin you can assign this to me

Owaiseimdad:
@gorkem-bwl , these are the changes I'm considering making. Let me know what you think.

https://github.com/user-attachments/assets/14de544b-0c48-4d9c-b2bd-9056bc3ff50f

gorkem-bwl:
Thanks for this Owaise - the new update looks prettier. Assigned.

PR #1938, opened 2025-03-15
Fix 1936 dark bg

Owaiseimdad:
## Describe your changes
The cancel option on multiple popups are like link instead of buttons, need to be a like button for better UI. I have made is like a button with gray bg.

## Feature number
#1936 

```Javascript
<div>{t('add')}</div>
```

Before:
Screenshot 2025-03-15 at 12 45 11 AM

After:
https://github.com/user-attachments/assets/0959bb86-6a9a-4404-afda-e926a904d08a

### DIFF
--- a/src/Components/Dialog/index.jsx
+++ b/src/Components/Dialog/index.jsx
@@ -30,12 +30,8 @@ const Dialog = ({
 			>
 				<Button
 					variant="contained"
+					color="secondary"
 					onClick={onCancel}
-					sx={{
-						"&:hover": {
-							backgroundColor: theme.palette.primary.lowContrast,
-						},
-					}}
 				>
 					Cancel
 				</Button>
--- a/src/Components/Dialog/index.jsx
+++ b/src/Components/Dialog/index.jsx
@@ -29,9 +29,13 @@ const Dialog = ({
 				justifyContent="flex-end"
 			>
 				<Button
-					variant="text"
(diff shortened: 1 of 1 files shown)


src/Components/Dialog/index.jsx
```diff
@@ -29,9 +29,13 @@ const Dialog = ({
...
 				<Button
-					variant="text"
-					color="info"
+					variant="contained"
 					onClick={onCancel}
+					sx={{
+						"&:hover": {
+							backgroundColor: theme.palette.primary.lowContrast,
+						},
+					}}
```
ajhollid, line 38: I don't think we need a special background for buttons in dialogs, just `variant="contained"` is sufficient here :thinking:

ajhollid (changes requested):

src/Components/Dialog/index.jsx
Owaiseimdad: Without &hover style,

https://github.com/user-attachments/assets/87cd3037-e844-4645-960f-07fbb20bb220

With hover style,

https://github.com/user-attachments/assets/5135197e-767d-4ac5-8620-a03b74bfa822

Let me know if the hover still not important, will remove it. Added clarity as it is already present in light mode.

gorkem-bwl:
Yes, hover style is a better approach here.

PR #1984, opened 2025-03-22
Added background for charts for clarity

Owaiseimdad:
## Describe your changes
Charts had a background of primary which is why most the bars figure seems missing but they are there just camoflaged

## Issue number
No issue was raised

- [] I have included the issue # in the PR.
```Javascript
<div>{t('add')}</div>
```

Before:
Screenshot 2025-03-22 at 9 50 00 AM

After:
Screenshot 2025-03-22 at 9 50 11 AM


src/Components/Charts/ChartBox/index.jsx
```diff
@@ -25,7 +25,7 @@ const ChartBox = ({
 			flex={1}
 			direction="row"
 			sx={{
-				backgroundColor: theme.palette.primary.main,
+				backgroundColor: theme.palette.tertiary.main,
```
Br0wnHammer, line 28: Hey @Owaiseimdad, I'm not sure the current color choice is appropriate. It makes the UI of the charts look out of place compared to the other designs. I suggest using a slightly lighter dark color to make it less noticeable and more cohesive with the overall design.

Br0wnHammer: To address the issue without significantly altering the UI, you could modify the background color of the gauge chart dial. This change will resolve the camouflaging problem with the background.

ajhollid: @Br0wnHammer I think it makes sense to apply the BG color to the chartbox as @Owaiseimdad has done here so that all charts will have the same background.

As for the specific color though, that's a question for @gorkem-bwl, I'm inclined to agree that a more subtle color might be more appropriate?

Br0wnHammer: @ajhollid, I believe that the issue is not with the chart background. The only UI element that blends into the background is the Radial Chart dial's background color, as shown in the figure. We need better color options for this element because the camouflaging problem persists in both light and dark modes. Additionally, we need different dial background colors for different response times, as shown for "Acceptable" in the screenshot.

Screenshot 2025-03-24 at 8 45 30 AM
Screenshot 2025-03-24 at 8 45 47 AM
Screenshot 2025-03-24 at 8 46 01 AM

Br0wnHammer: This is something that can be done.

Screenshot 2025-03-24 at 8 51 07 AM
Screenshot 2025-03-24 at 8 50 30 AM

Owaiseimdad: @Br0wnHammer, I understand your point. The problem is that changing the background of the charts would resolve the issue for all types of responses. However, if we need to update the UI of the radial chart, we should consider that similar issues might arise if we implement a chart where the bars are also divided by colors. We could approach this as a feature improvement and document it in a way that sets standard colors for most scenarios. Let me know your thoughts on this.

Br0wnHammer: @Owaiseimdad 
I understand your perspective now. We can approach this as a double feature improvement:

- Chart Background Update: We'll change the background of all charts to a slightly light darker color. This will address the current UI issue and provide a more consistent and visually appealing look across all chart types.
[Already being addressed in this PR]

- Radial Chart Dial Background: Along with the overall background change, we'll update the radial chart dial background color to align with the new coloring scheme. This will ensure that the radial chart maintains its unique appearance while fitting cohesively with the other charts.
[I can create a different PR to address this]

@gorkem-bwl, what say?

gorkem-bwl: That's a perfect discussion point in our call (TBH I had some difficulty following up with the responses so that's why I would like to understand it during the meeting) :-)

ajhollid:
@Owaiseimdad @Br0wnHammer what's the status on this one?  Is it ready for review again?

Br0wnHammer:
> @Owaiseimdad @Br0wnHammer what's the status on this one? Is it ready for review again?

I am not aware about what was decided in your meeting with @Owaiseimdad. If possible can you let me know what was decided?

Owaiseimdad:
@ajhollid and @Br0wnHammer will update the ticket soon, little occupied with something else right now. You will have a updated PR by end of week.

src/Components/Charts/ChartBox/index.jsx
Owaiseimdad: @Br0wnHammer @ajhollid now everything is as expected.

https://cdn.discordapp.com/attachments/1344728999502286940/1355658118662459685/Screenshot_2025-03-30_at_3.10.45_AM.png?ex=67e9ba82&is=67e86902&hm=2b1be910f02ee3301733cfc5d343ec3c26556f62019cb6eb38d7ee71f944d9e1&

https://cdn.discordapp.com/attachments/1344728999502286940/1355658322900029450/Screenshot_2025-03-30_at_3.12.03_AM.png?ex=67e9bab3&is=67e86933&hm=e778f2ac81654e993806d754a86ecc804504345c6b52df75a4fa6111c58da569&

https://cdn.discordapp.com/attachments/1344728999502286940/1355658467947315372/Screenshot_2025-03-30_at_3.12.38_AM.png?ex=67e9bad5&is=67e86955&hm=1d03b05e990a1a44c02451f1703832703b7a5cbf6ebdbc4fd1b76f2fd2a9c78a&

Br0wnHammer (approved):
LGTM!

ajhollid (approved):

PR #1985, opened 2025-03-22
fixed the status boxes on screen to be more responsive and better align with UI

Owaiseimdad:
## Describe your changes
The design of status boxes where pretty out of order and due to this the UI looked a bit off, added necessary css in order better align with needs

- [] I have included the issue # in the PR.
```Javascript
<div>{t('add')}</div>
```
- [] I didn't use any hardcoded values (otherwise it will not scale, and will make it difficult to maintain consistency across the application).
- [x ] I made sure font sizes, color choices etc are all referenced from the theme. I have no hardcoded dimensions.

Before:
https://github.com/user-attachments/assets/e09fd92e-3939-4c74-863b-b90bd098fd36

After:
https://github.com/user-attachments/assets/073378ba-576b-4e7c-8c62-ba09312e5676


ajhollid:
@gorkem-bwl can you weigh in on which design you prefer here?  I'm not qualified to make a judemgent :joy:

src/Components/StatBox/index.jsx
```diff
@@ -79,14 +79,27 @@ const StatBox = ({
 				},
 			};
 
+	const responsiveWidths = {
+		default: "calc(95% / 4)",
+		xl: "calc(95% / 3)",
+		md: "calc(95% / 2)",
```
ajhollid, line 85: Why is 95% the value to use here?

ajhollid (commented):
Seems fine to me, just wondering why 95% is the base value used for responsive width calculations?

src/Components/StatBox/index.jsx
ajhollid: @Owaiseimdad your thoughts on the 95%?

Owaiseimdad: Thats the only reason
or else we have to minus it with some px then

ajhollid: @Owaiseimdad did you manage to come up with anything for this per our discussion?

ajhollid: That's a bit too defeatist for me!  This can definitely be solved.

Why don't we use `flexBasis` to resolve this?

Gap in the parent is `theme.spacing(8)` right?  And we want 4 items per row.  Therefore we need to figure out their widths.  Seems to me all we have to do is subtract the width of the gaps from the flex basis and we should be good to go.

flexBasis of 100% / 4 items -  (3 * theme.spacing(8))  / 4 items 

should give us what we want.  Thiss reduces to 

`flexBasis: `calc(25% - (3 * ${theme.spacing(8)}/4))`

Seems to work to me, perhaps you can try it with the other sizes and verify

Owaiseimdad: > Gap in the parent is `theme.spacing(8)` right? And we want 4 items per row. Therefore we need to figure out their widths. Seems to me all we have to do is subtract the width of the gaps from the flex basis and we should be good to go.
> 
> flexBasis of 100% / 4 items - (3 * theme.spacing(8)) / 4 items
> 
> should give us what we want. Thiss reduces to
> 
> `flexBasis: `calc(25% - (3 * ${theme.spacing(8)}/4))`

YES @ajhollid , works pretty good.

For context:
The width for each item is the total available width (100%) divided by the number of items per row (4), minus the total gap between the items. There are 3 gaps between 4 items, so you subtract 3 * theme.spacing(8).
Same has been done for two status bars in UI as well. Thanks!! I think I gave up fast 🙈.

ajhollid: @Owaiseimdad no problem, in future when we have time we can refactor this into a grid which is more appropriate anyway

ajhollid (approved):

Issue #1990, opened 2025-03-24
Implement Filters on Uptime Monitors Page


Br0wnHammer:
**Is your feature request related to a problem? Please describe.**
We need to implement filtering on the uptime monitors page. This is the first stage of the filter implementation.

**Describe the solution you'd like**
Filters to Implement (Stage 1)

Type: `http`, `port`, `docker`, `ping`
Status: `Resumed`, `Paused`

Have a reset button present when a filter is selected.

**Backend Status**
✅ Mostly, the required DB queries are completed.

**Frontend Requirements**
Use the `Chip` component to handle filters effectively.
Ensure that filtering is smooth and intuitive.
`Debounce search queries` when filtering via string input, so we are not making a query for every keystroke.

**Tasks**

- [ ]  Implement Type and Status filters using the Chip component.
- [ ]  Integrate filtering logic with the backend.
- [ ]  Add debounce for string-based filtering.
- [ ]  Test and ensure proper functionality.

Br0wnHammer:
CC @gorkem-bwl, @ajhollid

gorkem-bwl:
Looks like a good start to me. You can play with the colors + remove the line widget explanation area (where it says "Type") later on. Also you can check whether the drop down menu line heights are the same with the ones in the components.

Br0wnHammer:
https://github.com/user-attachments/assets/d60e0188-2af8-4f66-8194-f7d405812984

@ajhollid @gorkem-bwl, need to clarify one thing.
How about we change the status filter box to `isActive (true, false)`

since down monitors also have the field isActive as true only their `status `is down.

gorkem-bwl:
> Filter.mp4 
> [@ajhollid](https://github.com/ajhollid) [@gorkem-bwl](https://github.com/gorkem-bwl), need to clarify one thing. How about we change the status filter box to `isActive (true, false)`
> 
> since down monitors also have the field isActive as true only their `status `is down. This is just a functionality testing draft. Final UI will be sent soon for approvals.

Yep, I am good with that. You might also think about right-aligning all the UI related elements added here.

Br0wnHammer:
> > Filter.mp4
> > [@ajhollid](https://github.com/ajhollid) [@gorkem-bwl](https://github.com/gorkem-bwl), need to clarify one thing. How about we change the status filter box to `isActive (true, false)`
[...]

PR #2010, opened 2025-03-27
Feat: Uptime Monitor Filters

Br0wnHammer:
## Describe your changes

This PR introduces an improved filtering system for monitors, allowing users to refine their search based on `Status`, `Type` and `State` using a multi-select dropdown UI.

## Issue number

#1990 

```Javascript
<div>{t('add')}</div>
```
- [] I didn't use any hardcoded values (otherwise it will not scale, and will make it difficult to maintain consistency across the application).


src/Pages/Uptime/Monitors/Components/SearchComponent/index.jsx
```diff
@@ -28,6 +28,7 @@ const SearchComponent = ({ monitors = [], onSearchChange, setIsSearching }) => {
 			width="25%"
 			minWidth={150}
 			ml="auto"
+			mt={2}
```
Br0wnHammer, line 31: For aligning with the remaining components.

src/Components/FilterHeader/index.jsx
```diff
@@ -0,0 +1,109 @@
...
+		"& .MuiSelect-icon": {
+			color: theme.palette.primary.contrastText,
+		},
+		"&:hover": {
+			backgroundColor: theme.palette.primary.main,
+		},
+		"&:hover .MuiOutlinedInput-notchedOutline": {
+			borderColor: theme.palette.primary.lowContrast,
+		},
+	};
```
ajhollid, line 33: Styles should not be applied selectively to components.

If this is a style for the Select component then it should be applied in the theme to all Select components.

There should be no styles/behaviours that are page specific.  Styles should be applie to components for consistency.

src/Components/FilterHeader/index.jsx
```diff
@@ -0,0 +1,109 @@
...
+	const menuItemStyles = {
+		"&:hover": {
+			backgroundColor: theme.palette.secondary.main,
+		},
+	};
+
+	return (
+		<div>
+			<FormControl
+				sx={{ m: theme.spacing(2), minWidth: 120 }}
```
ajhollid, line 44: Widths should not be fixed absolute values.  They shoud be relative, and there should be consideration given to responsiveness for small screens.

src/Components/FilterHeader/index.jsx
```diff
@@ -0,0 +1,109 @@
...
+								primary={
+									option === "http"
+										? "HTTP(S)"
+										: option === "ping"
+											? "Ping"
+											: option === "docker"
+												? "Docker"
+												: option === "port"
+													? "Port"
+													: option
```
ajhollid, line 90: This is over 10 levels of indentation deep, that should set alarm bells off for you :joy:  Nested ternaries are hard to read and prone to error, let's replace this with something flatter and easier to read.

src/Pages/Uptime/Monitors/Components/Filter/index.jsx
```diff
@@ -0,0 +1,109 @@
...
+	const statusOptions = ["Up", "Down"];
+	const stateOptions = ["Active", "Paused"];
+
+	return (
+		<Box
+			sx={{
+				display: "flex",
+				flexDirection: "row",
+				alignItems: "center",
+				ml: theme.spacing(80),
```
ajhollid, line 62: This is suspect, I suggest using flex values to align these items.  This will almost certainly cause us headaches when it comes to resonsive design

src/Pages/Uptime/Monitors/Components/Filter/index.jsx
```diff
@@ -0,0 +1,109 @@
...
+		>
+			<Button
+				color={theme.palette.primary.contrastText}
+				onClick={handleReset}
+				variant="contained"
+				endIcon={<ClearIcon />}
+				sx={{
+					"&:hover": {
+						backgroundColor: theme.palette.primary.lowContrast,
+					},
```
ajhollid, line 74: Please don't override default styles for components, behaviour and appearance of base components like Buttons should be consistent across all pages.

We do not want hover state of button to be one color on one page and a different color on another.

src/Pages/Uptime/Monitors/index.jsx
```diff
@@ -69,6 +70,9 @@ const UptimeMonitors = () => {
 	const [sort, setSort] = useState(undefined);
 	const [isSearching, setIsSearching] = useState(false);
 	const [monitorUpdateTrigger, setMonitorUpdateTrigger] = useState(false);
+	const [selectedTypes, setSelectedTypes] = useState([]);
+	const [toFilterStatus, setToFilterStatus] = useState(null);
+	const [toFilterActive, setToFilterActive] = useState(null);
```
ajhollid, line 75: Please use `undefined` as default state for states that should be empty.  This is the project convention.

src/Pages/Uptime/Monitors/Components/Filter/index.jsx
```diff
@@ -0,0 +1,109 @@
...
+import FilterHeader from "../../../../../Components/FilterHeader";
+import { useMemo, useState } from "react";
+import { Box, Button } from "@mui/material";
+import ClearIcon from "@mui/icons-material/Clear";
+
+const Filter = ({ selectedTypes, setSelectedTypes, setToFilterStatus, setToFilterActive }) => {
+	const theme = useTheme();
+
+	const [selectedState, setSelectedState] = useState([]);
+	const [selectedStatus, setSelectedStatus] = useState([]);
```
ajhollid, line 12: This component appears to control its own internal state as well as the state of a parent component.

This means that the component is tightly linked to the parent.  

If that is intentional, then this component's state should probably reside in the parent component as well.

ajhollid (changes requested):
A good start and funcitonal :rocket:   Lots of liltte things to clean up to get his up to par, please see comments in code review.

From a UX pov if a user selects any combination of `up/down` and `active/paused` we should display no monitors, as the sets of up/active up/puased down/active down/paused are all empty.

There's also an issue where when you deselect options and click elsewhere the select retains focus and we see an empty select until clicking a second time to move focus.  This should resolve on the first click.

image

src/Components/FilterHeader/index.jsx
```diff
@@ -0,0 +1,109 @@
...
+	InputLabel,
+	ListItemText,
+	MenuItem,
+	Select,
+} from "@mui/material";
+import { useTheme } from "@emotion/react";
+import PropTypes from "prop-types";
+import AddCircleOutlineIcon from "@mui/icons-material/AddCircleOutline";
+
+const FilterHeader = ({ header, options, value, onChange, multiple = true }) => {
```
Owaiseimdad, line 13: @Br0wnHammer possible to add a JSDOC, want to know what this files does in first place.

src/Components/FilterHeader/index.jsx
```diff
@@ -0,0 +1,109 @@
...
+							/>
+						</MenuItem>
+					))}
+				</Select>
+			</FormControl>
+		</div>
+	);
+};
+
+FilterHeader.propTypes = {
```
Owaiseimdad, line 101: So all the props are optional I guess that why the .isRequired is missing, let me know if I am correct here.

src/Components/FilterHeader/index.jsx
```diff
@@ -0,0 +1,109 @@
...
+						<MenuItem
+							key={option}
+							value={option}
+							sx={menuItemStyles}
+						>
+							<Checkbox
+								checked={value.includes(option)}
+								size="small"
+							/>
+							<ListItemText
```
Owaiseimdad, line 80: @Br0wnHammer , if the list of option increases in future then we need to start the editing here and again, is there a way to make it simple, like put all the options in a object and then use it here. 

Ex:
`
const optionLabels = {
  http: "HTTP(S)",
  ping: "Ping",
  docker: "Docker",
  port: "Port",
};

{
  options.map((option) => {
    const label = optionLabels[option] || option; // Fallback if no label found

    return (
      <MenuItem key={option} value={option} sx={menuItemStyles}>
        <Checkbox checked={value.includes(option)} size="small" />
        <ListItemText primary={label} />
      </MenuItem>
    );
  });
}

`

src/Pages/Uptime/Monitors/Components/Filter/index.jsx
```diff
@@ -0,0 +1,109 @@
...
+				header="State"
+				options={stateOptions}
+				value={selectedState}
+				onChange={handleStateChange}
+			/>
+		</Box>
+	);
+};
+
+Filter.propTypes = {
```
Owaiseimdad, line 102: Same, are these values optional? if not possible to add the .isRequired.

src/Pages/Uptime/Monitors/Components/Filter/index.jsx
```diff
@@ -0,0 +1,109 @@
+import { useTheme } from "@emotion/react";
+import PropTypes from "prop-types";
+import FilterHeader from "../../../../../Components/FilterHeader";
+import { useMemo, useState } from "react";
+import { Box, Button } from "@mui/material";
+import ClearIcon from "@mui/icons-material/Clear";
+
+const Filter = ({ selectedTypes, setSelectedTypes, setToFilterStatus, setToFilterActive }) => {
```
Owaiseimdad, line 8: @Br0wnHammer add the JSDOC if required, good for readability for future debugging or changes.

src/Pages/Uptime/Monitors/index.jsx
```diff
@@ -103,6 +107,19 @@ const UptimeMonitors = () => {
 			types: TYPES,
 			monitorUpdateTrigger,
 		});
+	
+	let field = sort?.field;
```
Owaiseimdad, line 111: possible to optimise this, 

ex:
`// Determine the field and filter values based on conditions
const field = toFilterStatus !== null ? "status" :
              toFilterActive !== null ? "isActive" :
              sort?.field;

const filter = toFilterStatus !== null ? toFilterStatus :
               toFilterActive !== null ? toFilterActive :
               search;
`

src/Components/FilterHeader/index.jsx
vishnusn77, line 1: please move the styles for the components to globalTheme. I would also suggest using `contrastTextSecondary` or `contrastTextTertiary` for the inputLabel and the icon, to blend it with rest of the UI.

Br0wnHammer:
> A good start and funcitonal 🚀 Lots of liltte things to clean up to get his up to par, please see comments in code review.
> 
> From a UX pov if a user selects any combination of `up/down` and `active/paused` we should display no monitors, as the sets of `up/active` `up/paused` `down/active` `down/paused` are all empty.
> 
> There's also an issue where when you deselect options and click elsewhere the select retains focus and we see an empty select until clicking a second time to move focus. This should resolve on the first click.
> 
> image

@ajhollid, should we handle this from BE? Since we need to have a query that returns no monitor found in such case.

ajhollid:
> > A good start and funcitonal 🚀 Lots of liltte things to clean up to get his up to par, please see comments in code review.
> > From a UX pov if a user selects any combination of `up/down` and `active/paused` we should display no monitors, as the sets of `up/active` `up/paused` `down/active` `down/paused` are all empty.
> > There's also an issue where when you deselect options and click elsewhere the select retains focus and we see an empty select until clicking a second time to move focus. This should resolve on the first click.
> > image
> 
> @ajhollid, should we handle this from BE? Since we need to have a query that returns no monitor found in such case.

I don't see any need to move this to the backend a tthis point, seems excessive to make a round trip http request for data that is guaranteed to be empty?

src/Pages/Uptime/Monitors/index.jsx
ajhollid: Same with `toFilterStatus` and `toFilterActive`, is there a reason these are `null` instead of `undefined`?

src/Pages/Uptime/Monitors/index.jsx
```diff
@@ -104,19 +110,41 @@ const UptimeMonitors = () => {
...
+		setToFilterActive(null);
+	};
+
+	const field =
+		toFilterStatus !== null
+			? "status"
+			: toFilterActive !== null
+				? "isActive"
+				: sort?.field;
+
```
ajhollid, line 127: Can we pleaese use a lookup here instead of nested ternaries?  If more fields are added in the future this will become even more unwieldy than it already is.

src/Pages/Uptime/Monitors/index.jsx
```diff
@@ -104,19 +110,41 @@ const UptimeMonitors = () => {
...
 	const [
 		monitorsWithChecks,
 		monitorsWithChecksCount,
 		monitorsWithChecksIsLoading,
 		monitorsWithChecksNetworkError,
 	] = useFetchMonitorsWithChecks({
 		teamId,
-		types: TYPES,
+		types: selectedTypes && selectedTypes.length > 0 ? selectedTypes : TYPES,
```
ajhollid, line 142: Let's move this out of the method parameters, logic should be handled before passing the value to the method.

ajhollid (changes requested):
Looking much better!

I'd like to eliminate those nested ternaries all together if possible, other than that mostly looking good to me :+1:   Nice work here

ajhollid (approved):
LGTM, thanks for making changes :+1:

PR #2016, opened 2025-03-29
Feature/email-uppercase-support

karins28:
FE side validation
added the validation to the interceptor itself

here is the BE ticket
https://github.com/bluewave-labs/checkmate-backend/pull/74

### DIFF
--- a/src/Components/TabPanels/Account/TeamPanel.jsx
+++ b/src/Components/TabPanels/Account/TeamPanel.jsx
@@ -107,12 +107,13 @@ const TeamPanel = () => {
 
 	const handleChange = (event) => {
 		const { value } = event.target;
+		const newEmail = value?.toLowerCase() || value
 		setToInvite((prev) => ({
 			...prev,
-			email: value,
+			email: newEmail,
 		}));
 
-		const validation = credentials.validate({ email: value }, { abortEarly: false });
+		const validation = credentials.validate({ email: newEmail }, { abortEarly: false });
 
 		setErrors((prev) => {
 			const updatedErrors = { ...prev };
@@ -153,7 +154,7 @@ const TeamPanel = () => {
 
 		try {
 			await networkService.sendInvitationToken({
-				email: toInvite.email.toLowerCase(),
+				email: toInvite.email,
 				role: toInvite.role,
 			});
 			closeInviteModal();
--- a/src/Components/TabPanels/Account/TeamPanel.jsx
+++ b/src/Components/TabPanels/Account/TeamPanel.jsx
@@ -153,7 +153,7 @@ const TeamPanel = () => {
 
 		try {
 			await networkService.sendInvitationToken({
-				email: toInvite.email,
+				email: toInvite.email.toLowerCase(),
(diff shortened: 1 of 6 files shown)


src/Utils/NetworkService.js
```diff
@@ -27,6 +27,9 @@ class NetworkService {
 		this.axiosInstance.interceptors.request.use(
 			(config) => {
 				const currentLanguage = i18next.language || "en";
+				if (config.data?.email) {
```
Br0wnHammer, line 30: Implementing the email lowercase conversion in the Axios interceptor is not the ideal place for this logic. The primary purpose of an Axios interceptor is to modify requests or responses globally. This should be handled in the form submission logic along with backend.

Br0wnHammer (commented):
You are headed in the right direction some changes are required for a proper implementation.

src/Utils/NetworkService.js
karins28: Thanks for the feedback!

I am debating between two approaches which do you think is more correct:
1.Do you think a solution on form submission similar to this
https://github.com/bluewave-labs/Checkmate/pull/2016/commits/fa8698843645e0ae47d290ebc3f6ffbb47bd2559
(but in all places) would be better

2.changing it on joi using alter/custom method similar to 
 .alter({
            lowercase: (schema) => schema.custom((value) => value.toLowerCase()),
        })

karins28: or 3. fix it in middleware for the related api requests

src/Validation/validation.js
```diff
@@ -65,13 +65,7 @@ const credentials = joi.object({
...
 		.trim()
 		.email({ tlds: { allow: false } })
-		.custom((value, helpers) => {
-			const lowercasedValue = value.toLowerCase();
-			if (value !== lowercasedValue) {
-				return helpers.message("Email must be in lowercase");
-			}
-			return lowercasedValue;
-		})
+		.lowercase()
```
ajhollid, line 68: This  doesn't appear to work, I can submit an email with uppercase letters

ajhollid (changes requested):
.lowercase() validation doesn't appear to be functional.

src/Validation/validation.js
karins28: @ajhollid My bad, 
I didn't realize the data transformation only works for the returned value 
https://www.haroldadmin.com/posts/joi-data-transformation

src/Components/TabPanels/Account/TeamPanel.jsx
```diff
@@ -153,7 +153,7 @@ const TeamPanel = () => {
 
 		try {
 			await networkService.sendInvitationToken({
-				email: toInvite.email,
+				email: toInvite.email.toLowerCase(),
```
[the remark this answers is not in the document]
ajhollid: @karins28 good point, I think it's OK to cause the value of the textfield to change though as the input should be lowercase anyways :thinking:

ajhollid (approved):
Looks good, thanks for making revisons :+1:

PR #2029, opened 2025-04-03
2002 fix: use available language

### LINKED ISSUE #2002: Warnings in LanguageSelector component
The console is being flooded with warnings from `LanguageSelector` component

Image

Is this because of system default language?

Cihatata:
## Describe your changes

compare appsettings language with available languages 

Fixes #2002 

### DIFF
--- a/src/App.jsx
+++ b/src/App.jsx
@@ -25,10 +25,14 @@ function App() {
 		if (authToken) {
 			dispatch(getAppSettings({ authToken })).then((action) => {
 				if (action.payload && action.payload.success) {
-					const { language } = action.payload.data;
-					if (language) {
+						const { language } = action.payload.data;
+				const availableLanguages = Object.keys(i18n.options.resources || {});
+					if (language && availableLanguages.includes(language)) {
 						dispatch(setLanguage(language));
 						i18n.changeLanguage(language);
+					} else {
+						dispatch(setLanguage(availableLanguages[0]));
+						i18n.changeLanguage(availableLanguages[0]);
 					}
 				}
 			});


src/App.jsx
```diff
@@ -25,10 +25,14 @@ function App() {
 		if (authToken) {
 			dispatch(getAppSettings({ authToken })).then((action) => {
 				if (action.payload && action.payload.success) {
-					const { language } = action.payload.data;
-					if (language) {
+						const { language } = action.payload.data;
+				const availableLanguages = Object.keys(i18n.options.resources || {});
+					if (language && availableLanguages.includes(language)) {
 						dispatch(setLanguage(language));
 						i18n.changeLanguage(language);
```
ajhollid, line 32: This works if you're logged in, but we also need a fix for the auth pages when you are not logged in

ajhollid (changes requested):
Thanks for getting this in, we need a fix for auth pages as well :+1:

src/App.jsx
Cihatata: I don't think this is necessary. The default value is set in `/src/utils/i18n.js` for not logged in case.

ajhollid: I still get the errors thrown and no appropriate value set to the selector on the auth pages, it's still trying to load my node env var

Cihatata: Are you sure about this? Can you clear the Redux store and browser storage and try again? Because I don't encounter it on the auth pages.

ajhollid: Yes, I cleared my redux state and local storage 🤔

I will check again in a couple hours when I get set up.

Cihatata: Hi @ajhollid, Did you test again? Is the problem still there?

ajhollid: @Cihatata yep appears to be OK, thanks for checking in to it!

ajhollid (approved):

PR #2049, opened 2025-04-10
Feat: Filters UI

Br0wnHammer:
## Describe your changes

This PR implements a new UI for Filters.

```Javascript
<div>{t('add')}</div>
```

Screenshot 2025-04-11 at 12 25 13 AM

### DIFF
--- a/src/Components/FilterHeader/index.jsx
+++ b/src/Components/FilterHeader/index.jsx
@@ -57,14 +57,6 @@ const FilterHeader = ({ header, options, value, onChange, multiple = true }) => 
 						horizontal: "left",
 					},
 				}}
-				sx={{
-					"& .MuiOutlinedInput-input": {
-						paddingLeft: theme.spacing(14),
-					},
-					"& .MuiSelect-icon": {
-						left: theme.spacing(4),
-					},
-				}}
 			>
 				{options.map((option) => (
 					<MenuItem
--- a/src/Components/FilterHeader/index.jsx
+++ b/src/Components/FilterHeader/index.jsx
@@ -57,6 +57,14 @@ const FilterHeader = ({ header, options, value, onChange, multiple = true }) => 
 						horizontal: "left",
 					},
 				}}
+				sx={{
+					"& .MuiOutlinedInput-input": {
+						paddingLeft: theme.spacing(14),
+					},
+					"& .MuiSelect-icon": {
+						left: theme.spacing(4),
+					},
+				}}
 			>
 				{options.map((option) => (
 					<MenuItem
--- a/src/Components/FilterHeader/index.jsx
+++ b/src/Components/FilterHeader/index.jsx
@@ -42,7 +42,7 @@ const FilterHeader = ({ header, options, value, onChange, multiple = true }) => 
 						return header;
 					}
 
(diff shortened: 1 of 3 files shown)


gorkem-bwl:
LGTM!

src/Utils/Theme/globalTheme.js
```diff
@@ -477,13 +477,15 @@ const baseTheme = (palette) => ({
...
+						paddingLeft: theme.spacing(14),
 					},
 					"& .MuiOutlinedInput-notchedOutline": {
 						borderColor: theme.palette.primary.lowContrast,
 						borderRadius: theme.shape.borderRadius,
 					},
 					"& .MuiSelect-icon": {
 						color: theme.palette.primary.contrastTextSecondary, // Dropdown + color
+						left: theme.spacing(4),
 					},
```
vishnusn77, line 489: I noticed that adding the `paddingLeft` and `left` in globalTheme is breaking the styles for other Select components.

image

It would be better to add the padding at component level.

vishnusn77 (changes requested):

src/Utils/Theme/globalTheme.js
Br0wnHammer: Oh shoot!
Let me fix it right away.
Thanks for the catch.

vishnusn77 (approved):
Looks good to me!

src/Components/FilterHeader/index.jsx
```diff
@@ -57,6 +57,14 @@ const FilterHeader = ({ header, options, value, onChange, multiple = true }) =>
...
 					},
 				}}
+				sx={{
+					"& .MuiOutlinedInput-input": {
+						paddingLeft: theme.spacing(14),
+					},
+					"& .MuiSelect-icon": {
+						left: theme.spacing(4),
+					},
+				}}
```
ajhollid, line 67: We should avoid component specific theme overrides.

If this is a fix for select components then it should apply to all select components, otherwise we will end up with selects with different styles on different pages which defeats the purpose of having a theme

ajhollid (changes requested):
Overall looks fine to me, but I really don't like to see the component specific theme overrides, components should look the same on all pages.

src/Utils/Theme/globalTheme.js
vishnusn77: @ajhollid he added the padding in the sx props because it was messing with other select components

src/Components/FilterHeader/index.jsx
Br0wnHammer: @gorkem-bwl, what say? Should we not have the + icon on the left and instead let it be on the default position?

gorkem-bwl: I don't have a big preference having the + icon on either side. Hence, whichever method is easier and doesn't mess up with other components, we can go with that.

Br0wnHammer:
@ajhollid, have made the changes and should be fine for merging now.


Screenshot 2025-04-13 at 7 25 25 PM

ajhollid (approved):
LGTM, thankls for the contribution and revisions!

PR #2128, opened 2025-04-25
fix: save settings

ajhollid:
This PR udpates the async thunks for saving settings and implements better error handling while saving

## Summary by CodeRabbit

- **Refactor**
  - Improved performance of settings save operation by running user and app settings updates concurrently.
  - Simplified app settings update to only include the language property.
  - Enhanced translation support with a new message for insufficient permissions.
  - Updated error response for insufficient permissions to use the correct HTTP status code.
  - Broadened permissions for updating app settings to include both admin and superadmin roles.

- **Style**
  - Minor formatting improvements in alert message definitions.

### DIFF
--- a/client/src/Features/Settings/settingsSlice.js
+++ b/client/src/Features/Settings/settingsSlice.js
@@ -30,26 +30,9 @@ export const getAppSettings = createAsyncThunk(
 export const updateAppSettings = createAsyncThunk(
 	"settings/updateSettings",
 	async ({ settings }, thunkApi) => {
- 		// The reason for commenting is that, previously, we had the flexibility to set the API base. 
-  		// However, now this could lead to an issue where it gets set to undefined.
-		// networkService.setBaseUrl(settings.apiBaseUrl);
 		try {
 			const parsedSettings = {
-				apiBaseUrl: settings.apiBaseUrl,
-				logLevel: settings.logLevel,
 				language: settings.language,
-				clientHost: settings.clientHost,
-				jwtSecret: settings.jwtSecret,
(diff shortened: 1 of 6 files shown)


server/middleware/isAllowed.js
```diff
@@ -41,7 +41,7 @@ const isAllowed = (allowedRoles) => {
 				return;
 			} else {
 				const error = new Error(stringService.insufficientPermissions);
```
Owaiseimdad, line 43: @ajhollid the insufficientPermissions was added here but the getter was never set before?

client/src/Pages/Settings/index.jsx
```diff
@@ -123,31 +123,17 @@ const Settings = () => {
...
+			]);
 
-			if (action.payload.success && settingsAction.payload.success) {
-				createToast({
-					body: t("settingsSuccessSaved"),
-				});
+			if (userAction.payload.success && settingsAction.payload.success) {
+				createToast({ body: t("settingsSuccessSaved") });
 			} else {
-				if (action.payload) {
```
Owaiseimdad, line 136: great! simple return here.

client/src/Features/Settings/settingsSlice.js
```diff
@@ -30,26 +30,9 @@ export const getAppSettings = createAsyncThunk(
 export const updateAppSettings = createAsyncThunk(
```
Owaiseimdad, line 30: @ajhollid , just a question here, what is the reason to remove all the values? Are we not using it any more or just that we will be not editing this at all?

Issue #1803, opened 2025-02-23
Bulk importing of uptime monitoring servers to the system


gorkem-bwl:
Here are some details and use-cases:

To make migrations easier, we can add an automatic server addition feature to Checkmate. This will allow admins to upload a CSV file with server details (URL, display name, and monitor frequency) directly from the UI, saving time and effort.

How it works:

- Admins click the "Add Servers (Uptime)" button on the dashboard.
- They upload a CSV file with server info.
- Checkmate processes the file and adds the servers to the database quickly.

Benefits are faster migration for users switching from other tools and improved user interface.

We already have a function in Settings that adds up to 300 servers at once using a static file. This can be adapted for CSV uploads, making implementation faster and easier. It will also give you an understanding of how adding servers work in general.

The bare minimum of data need to create a `Monitor` is as follows:

```js
{
   name: "Your Monitor Name"
   url: "www.your-url.com"
}
```

Everything else is optional, but can be included.

Important to note that the correct `teamId` must be provided or users will not be able to view their monitors.

Monitor schema for reference:

```js
import mongoose from "mongoose";

const MonitorSchema = mongoose.Schema(
	{
		userId: {
			type: mongoose.Schema.Types.ObjectId,
			ref: "User",
			immutable: true,
			required: true,
		},
		teamId: {
			type: mongoose.Schema.Types.ObjectId,
			ref: "Team",
			immutable: true,
			required: true,
		},
		name: {
			type: String,
			required: true,
		},
		description: {
			type: String,
		},
		status: {
			type: Boolean,
			default: undefined,
		},
		type: {
			type: String,
			required: true,
			enum: [
				"http",
				"ping",
				"pagespeed",
				"hardware",
				"docker",
				"port",
				"distributed_http",
			],
		},
		url: {
			type: String,
			required: true,
		},
		port: {
			type: Number,
		},
		isActive: {
			type: Boolean,
			default: true,
		},
		interval: {
			// in milliseconds
			type: Number,
			default: 60000,
		},
		uptimePercentage: {
			type: Number,
			default: undefined,
		},
		thresholds: {
			type: {
				usage_cpu: { type: Number },
				usage_memory: { type: Number },
				usage_disk: { type: Number },
			},
			_id: false,
		},
		notifications: [
			{
				type: mongoose.Schema.Types.ObjectId,
				ref: "Notification",
			},
		],
		secret: {
			type: String,
		},
	},
	{
		timestamps: true,
	}
);

[...]

PR #2154, opened 2025-04-26
Feature release for bulk upload of uptime monitors

Owaiseimdad:
## Describe your changes
Added Bulk Import Support:
A new page for bulk importing monitors has been created under /uptime/bulk-import. The page allows users to upload a CSV file, parse it, validate it, and submit the data to create multiple monitors at once.

CreateMonitorHeader Update:
Modified CreateMonitorHeader component to support dynamic bulkPath prop, allowing flexibility in specifying the bulk import URL.

CSV Template:
A sample CSV file (bulk_import_monitors_sample.csv) was added, which provides the structure for the bulk import data (type, name, url, interval, etc.).

CSV Parsing & Validation:
Implemented CSV parsing using the papaparse library and added validation logic (bulkMonitorsValidation) to check the correctness of the data before submission.

Network Service:
Added a method createBulkMonitors to the NetworkService for posting bulk monitor data to the backend.

UI Texts & Localization:
All visible strings related to bulk import functionality (like "Select File", "Parsing Failed", "Upload Success") were translated using i18n (useTranslation).

#1803 

```Javascript
<div>{t('add')}</div>
```

Video:

https://github.com/user-attachments/assets/5b374494-438c-4a87-8a7a-3e1779479fe1

## Summary by CodeRabbit

- **New Features**
  - Introduced bulk import functionality for uptime monitors via CSV file upload, accessible from the monitors page.
  - Added a dedicated bulk import page with instructions, downloadable templates, and sample files.
  - Users receive feedback on upload success or errors and are redirected to the overview page upon successful import.
  - Added file upload component with validation for CSV files.

- **Improvements**
  - Enhanced localization with new strings for bulk import features.
  - UI updated to include a "Bulk Import" button alongside "Create new" for easier access.

- **Bug Fixes**
  - Improved validation and error handling for CSV uploads and monitor creation.


gorkem-bwl:
I like how the bulk import has its own page. It could also work as a popup, but this looks good to me.

A few suggestions:

- The button labels could use sentence case. For example, "Bulk Import" can be "Bulk import" and "Select File" can be "Select file."
- I didn't see the `bulk_import_monitors_template.csv` file in the repo. Maybe I missed it, but just double-check.
- Are we validating the CSV file? What happens if the file is missing required columns or is badly formatted?
- If we know how many servers are being imported, it would be better to show it on the button. Like instead of just "Submit," something like "Bulk import N servers" would make it clearer for the admin.
- After the bulk import is complete, it'd be nice if the page updated automatically without needing a manual refresh. Otherwise it might cause some confusion.
- We should probably limit the number of servers per bulk import, maybe 100 at a time to avoid any performance hits. If we do that, it'd be good to mention the limit somewhere near the upload section.

ajhollid (changes requested):
In general it looks fine to me.

Unless there is a copmpelling reason to do parsing on the FE let's drop the library and do the parsing on the BE.  The less work we do and the less bloat we have on the FE the better.

Other than that we can extract the network operation to a hook to keep things nice and clean.

client/src/Pages/Uptime/BulkImport/index.jsx
```diff
@@ -0,0 +1,110 @@
...
+	const navigate = useNavigate();
+	const { t } = useTranslation();
+	const [selectedFile, setSelectedFile] = useState(null);
+
+	const crumbs = [
+		{ name: t("uptime"), path: "/uptime" },
+		{ name: t("bulkImport.title"), path: `/uptime/bulk-import` },
+	];
+
+	const { createBulkMonitors, isLoading: hookLoading, error } = useBulkMonitors();
```
ajhollid, line 29: @Owaiseimdad can you use array destructuring here instead?  It's the standard pattern for React, like `useState` for example:

`const [tab, setTab] = useState(0)`

ajhollid (changes requested):
LGTM for the most part, can we restructure that hook though to use array desctructuring instead of object destructuring?

It's especially handy for reassigning names if you have multiple hooks.

Br0wnHammer (approved):
LGTM!
Quite a large PR. Better to break this down further in the future for faster reviewing and merging.

ajhollid:
> LGTM! Quite a large PR. Better to break this down further in the future for faster reviewing and merging.

@Br0wnHammer can you do some testing with invalid input, malformed CSV etc?

@Owaiseimdad and I did some testing and it behaves unpredictably with unexpected input

ajhollid (changes requested):
Error handling needs some more work as discussed, currently behaves unpredictably with unexpected input

server/controllers/monitorController.js
```diff
@@ -243,56 +244,122 @@ class MonitorController {
...
-								return await this.db.createNotification(notification);
+			const { parse } = pkg;
+	
+			if (!req.file) {
+				return res.status(400).json({
+					errors: [
+						{ message: "No file uploaded" }
+					]
+				});
+			}
```
ajhollid, line 266: We should not return here, we should throw an error.  

We have error handling middleware that creates standard error responses and handles all errors.

server/controllers/monitorController.js
```diff
@@ -243,56 +244,122 @@ class MonitorController {
...
+	
+			const { userId, teamId } = req.body;
+	
+			if (!userId || !teamId) {
+				return res.status(400).json({
+					errors: [
+						{ message: "Missing userId or teamId in form data" }
+					]
+				});
+			}
```
ajhollid, line 276: Same here, errors should always be handled by throwing, not returning

server/controllers/monitorController.js
```diff
@@ -243,56 +244,122 @@ class MonitorController {
...
+	
+					// Handle 'port' and 'interval' fields, check if they're valid numbers
+					if (["port", "interval"].includes(header)) {
+						const num = parseInt(value, 10);
+						if (isNaN(num)) {
+							return res.status(400).json({
+								errors: [
+									{ message: `${header} should be a valid number, got: ${value}` }
+								]
+							}); // Return error immediately with this structure
```
ajhollid, line 297: Same here

server/controllers/monitorController.js
```diff
@@ -243,56 +244,122 @@ class MonitorController {
...
+					return value;
+				},
+				complete: async ({ data, errors }) => {
+					if (errors.length > 0) {
+						return res.status(400).json({
+							errors: [
+								{ message: "Error parsing CSV" }
+							]
+						});
+					}
```
ajhollid, line 311: Same here, also no hardcoded strings please

ajhollid (changes requested):
Error handling should be left up to the middleware, that's why it is there, to produce a standard error response.

Parsing still need some work, if I try to submit this CSV:

```
this,is,a,test
```

it give a success toast.  It should display an error toast. 

Same with an empty CSV with correct column headers.  It displays "success" but it should show an error

ajhollid (approved):
Error handlign appears sufficient now :rocket:

PR #2156, opened 2025-04-27
bug/2108-add-ssl-feature-flag

karins28:
**##draft pr**

**(Please remove this line only before submitting your PR. Ensure that all relevant items are checked before submission.)** 

## Describe your changes

Fixes #123 

- [ X] (Do not skip this or your PR will be closed) I deployed the application locally.
- [ ] (Do not skip this or your PR will be closed) I have performed a self-review and testing of my code.
- [X ] I have included the issue # in the PR.
- [ X] I have added i18n support to visible strings (instead of `<div>Add</div>`, use): 
```Javascript
<div>{t('add')}</div>
```
- [ X] The issue I am working on is assigned to me.
- [ ]X I made sure font sizes, color choices etc are all referenced from the theme. I have no hardcoded dimensions.
- [ X] My PR is granular and targeted to one specific feature.

### DIFF
--- a/server/controllers/distributedUptimeController.js
+++ b/server/controllers/distributedUptimeController.js
@@ -48,8 +48,13 @@ class DistributedUptimeController {
 			}
 
 			// Calculate if server is up or down
+			let isSelfSigned = false;
 			const isErrorStatus = status_code >= 400;
 			const hasError = error !== "";
+ 			if(error === 'DEPTH_ZERO_SELF_SIGNED_CERT' || error === 'SELF_SIGNED_CERT_IN_CHAIN'){
+ 				const monitor = await getMonitorById(id)
+ 				isSelfSigned = monitor.insecureSkipVerify
+ 				}
 
 			const status = isErrorStatus || hasError ? false : true;
 
--- a/client/src/Pages/PageSpeed/Create/index.jsx
+++ b/client/src/Pages/PageSpeed/Create/index.jsx
@@ -51,6 +51,7 @@ const CreatePageSpeed = () => {
 		type: "pagespeed",
 		notifications: [],
 		interval: 3,
+		insecureSkipVerify: false,
 	});
 
 	const [https, setHttps] = useState(true);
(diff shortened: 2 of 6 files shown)


server/controllers/distributedUptimeController.js
```diff
@@ -48,8 +48,13 @@ class DistributedUptimeController {
 			}
 
 			// Calculate if server is up or down
+			let isSelfSigned = false;
 			const isErrorStatus = status_code >= 400;
 			const hasError = error !== "";
+ 			if(error === 'DEPTH_ZERO_SELF_SIGNED_CERT' || error === 'SELF_SIGNED_CERT_IN_CHAIN'){
+ 				const monitor = await getMonitorById(id)
+ 				isSelfSigned = monitor.insecureSkipVerify
+ 				}
```
ajhollid, line 58: This logic should be done later in the process, likely in the status service as we'll arleady have the monitor there.

If we do it here we need to query the db again unnecessarily

karins28: @ajhollid thanks for the help! 
it makes sense to handle it only once,

I am still new to the repo so i'm learning about the logic:
there are two ways i see on how to handle it :

1.One like this - using the error code, in that case by handling the error.code (but on the 
1.1.should all the controllers that call the statusService handle this error? (aka are they all monitor types)?
in that case i'll need to update the response that it sends the error code (it doesn't have a 

2.adding to the config rejectUnauthorized=false (in that case it's easier but can reject other certificates) 

also another question: 
do i need to add the ability to update the flag to an existing monitor (ui + api)?

ajhollid: @karins28 I think the thing to probably do here is when the request is made, check if self signed certificates are allowed, and if so, configure Axios to ignore ssl errors.  No need to wait for an error to occur.

Yes, if a user can set the ignore SSL optiont they should also be able to unset it later.

gorkem-bwl: @karins28 let me know the latest status here please :)

PR #2173, opened 2025-04-30
fix: update build_images.sh to be backwards compatible with older bash ver…

ajhollid:
This PR updates the build images script to be backwards compatible


Br0wnHammer (approved):
LGTM!

Issue #2112, opened 2025-04-22
Retry Loop never resolving Race Condition


jnovack:
**Describe the bug**
If the `server` container starts before the `redis` container (read: before the dns entry of the `redis` container hits the DNS server), the server container will never resolve the IP of the `redis` container and will result in an infinite loop and never resolving the DNS entry.

**To Reproduce**
1. `docker stack deploy checkmate.yml` within Docker Swarm.
2. See error

This occurs naturally as Docker Swarm does not respect `depends-on` from docker-compose as you intend.

**Expected behavior**
I am expecting the `server` container to down itself after a pre-defined limit of failed attempts contacting either the `redis` or `mongo` end-points. 

If 5 backed-off retries (immediately, 0.5 sec, 1 sec, 2 sec, 4 sec) fail, the container should exit with error.  It is the responsibility of the orchestration engine (swarm, kubernetes, etc) to respawn the container, which gives the other containers plenty of time to be functional and registered.

**Screenshots**
```
checkmate_server.1.fa0rb3gwhxyj@swarm-master-04    | Error: getaddrinfo ENOTFOUND redis
checkmate_server.1.fa0rb3gwhxyj@swarm-master-04    |     at GetAddrInfoReqWrap.onlookup [as oncomplete] (node:dns:109:26) {
checkmate_server.1.fa0rb3gwhxyj@swarm-master-04    |   errno: -3008,
checkmate_server.1.fa0rb3gwhxyj@swarm-master-04    |   code: 'ENOTFOUND',
checkmate_server.1.fa0rb3gwhxyj@swarm-master-04    |   syscall: 'getaddrinfo',
checkmate_server.1.fa0rb3gwhxyj@swarm-master-04    |   hostname: 'redis'
checkmate_server.1.fa0rb3gwhxyj@swarm-master-04    | }
```

**Work Around**
1. `docker service scale checkmate_server=0 && docker service scale checkmate_server=1`

gorkem-bwl:
This is resolved so closing.

PR #2199, opened 2025-05-05
feat: redis service, resovles #2112

ajhollid:
This PR adds a redis service to handle redis connection.

- [x] Refactor redis connection into a service class
- [x] Implement a retry strategy with linear backoff
 

## Summary by CodeRabbit

- **New Features**
  - Introduced a managed Redis service for improved connection handling and lifecycle management.
  - Added the ability to clear all Redis data during server shutdown for better resource management.
  - Logger now supports debug-level messages, configurable via environment variables.

- **Improvements**
  - Enhanced job queue initialization by integrating Redis service through dependency injection.
  - Logging is now more flexible with dynamic log level configuration.

### DIFF
--- a/server/service/redisService.js
+++ b/server/service/redisService.js
@@ -1,0 +1,71 @@
+class RedisService {
+	static SERVICE_NAME = "RedisService";
+
+	constructor({ logger, IORedis, SettingsService }) {
+		this.logger = logger;
+		this.IORedis = IORedis;
+		this.SettingsService = SettingsService;
+		this.connection = null;
+	}
+
+	static async createInstance({ logger, IORedis, SettingsService }) {
+		const instance = new RedisService({ logger, IORedis, SettingsService });
+		await instance.connect();
+		return instance;
+	}
+
+	async connect() {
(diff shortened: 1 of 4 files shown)


server/service/redisService.js
```diff
@@ -0,0 +1,72 @@
...
+			retryStrategy: (times) => {
+				if (times >= 5) {
+					throw new Error("Failed to connect to Redis");
+				}
+				this.logger.debug({
+					message: "Retrying Redis connection",
+					service: RedisService.SERVICE_NAME,
+					details: { times },
+				});
+				return Math.min(times * 100, 2000);
```
Owaiseimdad, line 31: @ajhollid , this looks like a liner backoff to me. 

100, 200, 300, 400, 500 and so onn.

If this is the call then its fine, or else as you mentioned the exponential backoff, 

return Math.min(100 * Math.pow(2, times), 2000); this is should be good.

Owaiseimdad (commented):
LGTM except the return statement as you have mentioned as exponential backoff.

Owaiseimdad (approved):

PR #2225, opened 2025-05-09
fix - font size to 13px

### LINKED ISSUE #2056: Fonts in button group components must be 13px
Just like the title says, the fonts here should be 13px and not 14px

Image

Trained-Monkey:
## Describe your changes

Made fontSize for buttons reference the set variable, all buttons should now have a font size of 13px as requested.

Fixes #2056 

Screen Shot 2025-05-09 at 10 24 49 pm

```Javascript
<div>{t('add')}</div>
```

## Summary by CodeRabbit

- **Style**
  - Updated button font size to use a CSS environment variable for improved consistency across the application.
  - Adjusted base font size in the theme to use a dynamic value for better scalability.

### DIFF
--- a/client/src/Utils/Theme/globalTheme.js
+++ b/client/src/Utils/Theme/globalTheme.js
@@ -9,7 +9,7 @@ const shadow =
 const baseTheme = (palette) => ({
 	typography: {
 		fontFamily: fontFamilyPrimary,
-		fontSize: 14,
+		fontSize: typographyLevels.base,
 		h1: {
 			fontSize: typographyLevels.xl,
 			color: palette.primary.contrastText,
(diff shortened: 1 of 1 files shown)


gorkem-bwl:
UI-wise looks good to me.

ajhollid (changes requested):
Hey @Trained-Monkey ,

You're on the right track here by utilizing the theme, a couple of suggestions:

1.  We're trying to move away from using CSS files and using the theme entirely, so it would be best not to reference CSS files from the theme.  We want one source of truth for styling.

2.  Rather than adjusting the font size of the `Button` component, we can adjust the base font size of the entire theme:

```
const baseTheme = (palette) => ({
	typography: {
		fontFamily: fontFamilyPrimary,
		fontSize: 14,
	}
	...
```

This should resolve a number of issues including the button font size.

ajhollid (approved):
@Trained-Monkey you got it, that's what I was thinking! Thanks for the fix, let's go!🚀

PR #2232, opened 2025-05-09
feat: email settings

ajhollid:
This PR adds a settings section for configuring email settings.  It also updates the email service to handle loading settings from DB instead of env

- [x] Add an email setting section
- [x] Handle email settings in email service
- [x] Change settings return type  

image

image


fivemhostingserver:
Excellent contribution to open source! The community will benefit greatly from this.

PR #2246, opened 2025-05-11
fix: gb->en

ajhollid:
This PR sets the default language to en instead of gb

- [x] Remove conversion of `en` to `gb` from workflow script
- [x] Set `en` as default language
- [x] Rename `gb.json` to `en.json`
- [x] Update language selector to select `gb` flag for `en` as there is no `en` flag


Cihatata:
we should update `upload-poeditor.yml` and `poeditor-sync.yml`

PR #2258, opened 2025-05-13
fix: modal centered respective to parent

### LINKED ISSUE #1705: Modals display centered relative to the viewport, not their container

Trained-Monkey:
Modal shifts itself by how much width sidebar is taking

## Describe your changes

Added a new sidebar width constant in themes used to set sidebar sizes.
This value is also used to shift modal depending on whether the sidebar is collapsed or not.

### Reasoning for approach
Reading over previous pull requests on this issue (#1728 and #1950 ), it seemed like a solution primarily focusing on css was preferred. However, as the Material UI modal attaches itself to the document.body, a pure css solution did not look possible. This was the only method I could think of that was primarily css and did not involving passing around a ref of the parent component.

Fixes #1705 

Sidebar Expanded Upptime
Pagespeed
Sidebar Collapsed Uptimes

```Javascript
<div>{t('add')}</div>
```

## Summary by CodeRabbit

- **New Features**
  - Improved dialog positioning to anchor modals relative to specific layout elements for a more consistent user experience.
  - Background scrolling is now prevented when dialogs are open, reducing distractions and maintaining focus.
- **Enhancements**
  - Dialogs now maintain stable layout without shifting content when opened or closed.


client/src/Components/Dialog/genericDialog.jsx
```diff
@@ -20,7 +23,7 @@ const GenericDialog = ({ title, description, open, onClose, theme, children }) =
 					position: "absolute",
 					top: "50%",
 					left: "50%",
-					transform: "translate(-50%, -50%)",
+					transform: `translate(${sidebarWidth / 2}px, 0) translate(-50%, -50%)`,
```
Br0wnHammer, line 26: I like the approach for adjusting the modal position depending on the sidebar menu position. However, it still doesn't fix the issue at hand.

Br0wnHammer (commented):
Dialog Modals usually position themselves at the center of the viewport by default. However, our goal is to center the Dialog Modal relative to its parent component, not the entire viewport. The solution you proposed only adjusts the position based on the sidebar's position.

Trained-Monkey:
Hi, I've modified the code so that it now correctly centered to the parent component. One thing to note is as a result of this, if the users display is too small, it may appear slightly off center since some of the parent component will be off screen.

client/src/Components/Dialog/genericDialog.jsx
vishnusn77, line 1: One small thing I noticed, when the dialog opens, the background layout shifts slightly to the right. 
This is happening because of  ```document.body.style.overflow = 'hidden' ```,  which removes the scrollbar. 

To fix this, I believe we can preserve the scrollbar space by adding some padding-right to the body when we disable scroll.

client/src/Utils/DialogAnchorProvider.jsx
```diff
@@ -0,0 +1,10 @@
+import { useRef } from 'react';
+let DialogAnchorRef;
```
vishnusn77: using a global let for ```DialogAnchorRef``` works for now, but could become fragile if we have more dynamic layouts in the future. A small React Context might be a cleaner solution long-term, especially since we're moving off Redux.

vishnusn77:
Overall, Looks good, and is working as expected.

Trained-Monkey:
Hi, I've implemented changes to fix the highlighted issues.

Opening modal now checks if there is a scrollbar present and allocates space accordingly.

DialogAnchor is now contained in a react context.

gorkem-bwl:
@Trained-Monkey could you check the "format checks" please? We would like this PR go in to the next release :)

Trained-Monkey:
Hi, I have ran the format checks, they should be good now.

gorkem-bwl:
Great, passing now :)

ajhollid:
Hey @Trained-Monkey , apologies for the long delay on this PR.  After discussion with the team we're not 100% comfortable with the DOM manipulation that's needed to make this PR work.  We're going to table this issue for now, but if you come up with an approach that doesn't require manipulation of the DOM please let us know and we'll consider it.

Thank you for your effort here, we do appreciate it, but we're not quite ready to move on this one.

PR #2273, opened 2025-05-14
fix: TypeError in hardware notifications for null disk values

fredrikburmester:
## Describe your changes

Fixed a TypeError that occurs in the hardware notification handling when disk values are null. Used optional chaining (?.) and nullish coalescing (??) operators to safely handle null disk values when checking disk usage thresholds.

## Summary by CodeRabbit

- **Bug Fixes**
	- Improved reliability of disk usage alerts by preventing errors when disk data is missing or unavailable.

### DIFF
--- a/server/service/notificationService.js
+++ b/server/service/notificationService.js
@@ -257,7 +257,7 @@ class NotificationService {
 		const alerts = {
 			cpu: cpuThreshold !== -1 && cpuUsage > cpuThreshold ? true : false,
 			memory: memoryThreshold !== -1 && memoryUsage > memoryThreshold ? true : false,
-			disk: disk?.some(d => diskThreshold !== -1 && d.usage_percent > diskThreshold) ?? false,
+			disk: disk?.some(d => diskThreshold !== -1 && typeof d?.usage_percent === "number" && d?.usage_percent > diskThreshold) ?? false,
 		};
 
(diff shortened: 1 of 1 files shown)


server/service/notificationService.js
```diff
@@ -257,9 +257,7 @@ class NotificationService {
 		const alerts = {
 			cpu: cpuThreshold !== -1 && cpuUsage > cpuThreshold ? true : false,
 			memory: memoryThreshold !== -1 && memoryUsage > memoryThreshold ? true : false,
-			disk: disk.some((d) => diskThreshold !== -1 && d.usage_percent > diskThreshold)
-				? true
-				: false,
+			disk: disk?.some(d => diskThreshold !== -1 && d.usage_percent > diskThreshold) ?? false,
```
ajhollid, line 260: This looks good, why don't we also be defensive around the access to the `disk` object as well?  If `disk` is malformed then we want to evaluate to false.

We can add a `typeof d.usage_percent === "number"` check to this conditional to be a little safer.

ajhollid (changes requested):
Looks good to me, I think we can be defensive around access to `disk` objects as well.

fredrikburmester:
Added the suggested changes @ajhollid

ajhollid (approved):
Thank you for your contribution!

PR #2451, opened 2025-06-14
Better URL Handling on Infrastructure Create Page

Kaotic3:
This is a simple string manipulation involving 4 lines of code:

Clean the URL (already done but also removes any trailing "/")
Add the default Port Number if one does not exist already (:59232) - leaves the ability to have a non-default value.
Append 'api/v1/metrics' to the URL
Changing the default placeholder text to be "e.g. localhost or 192.168.1.100"

The code is designed to discard anything after a "/" so as to enforce the correct "/api/v1/metrics" - so even if it is there, it just discards it and adds it again.

I felt this was a very small change, string manipulation, that would make life easier for people like me running 20 different VMs, I already have to go install Capture on them all, might as well save a few minutes not having to type or copy a default value that never changes.

I know I have not completed the whole rebuild steps - but its 4 lines of string manipulation code, it can't break anything as it is simply a replace function, if its not there, it doesn't do it.

### DIFF
--- a/client/src/Pages/Infrastructure/Create/index.jsx
+++ b/client/src/Pages/Infrastructure/Create/index.jsx
@@ -89,8 +89,17 @@ const CreateInfrastructureMonitor = () => {
 	useEffect(() => {
 		if (isCreate || !monitor) return;
 
+		// Clean URL including any trailing slashes
+		let urlFixed = monitor.url.replace(/^https?:\/\//, "").replace(/\/+$/, "");
+
+		// Create default port if none included
+		if (!/:\d+(\/|$)/.test(urlFixed)) urlFixed += ":59232";
+
+		// Ensure path is exactly /api/v1/metrics
+		urlFixed = urlFixed.replace(/\/.*$/, "") + "/api/v1/metrics";
+
 		setInfrastructureMonitor({
-			url: monitor.url.replace(/^https?:\/\//, ""),
+			url: urlFixed,
 			name: monitor.name || "",
 			notifications: monitor.notifications,
 			interval: monitor.interval / MS_PER_MINUTE,
@@ -231,9 +240,9 @@ const CreateInfrastructureMonitor = () => {
 					...(isCreate
 						? [{ name: "Create", path: "/infrastructure/create" }]
 						: [
-								{ name: "Details", path: `/infrastructure/${monitorId}` },
-								{ name: "Configure", path: `/infrastructure/configure/${monitorId}` },
-							]),
+							{ name: "Details", path: `/infrastructure/${monitorId}` },
+							{ name: "Configure", path: `/infrastructure/configure/${monitorId}` },
+						]),
 				]}
 			/>
 			<Stack
@@ -289,7 +298,7 @@ const CreateInfrastructureMonitor = () => {
 							id="url"
 							name="url"
 							startAdornment={<HttpAdornment https={https} />}
-							placeholder={"localhost:59232/api/v1/metrics"}
+							placeholder={"e.g. localhost or 192.168.1.100"}
 							label={t("infrastructureServerUrlLabel")}
 							https={https}
 							value={infrastructureMonitor.url}


ajhollid (changes requested):
I understand where you are coming from here, but this could make life difficult for people that have Capture behind reverse proxies.  

It's not a given that every user will need to specify the port in their URL, nor is it given that they require the `/api/v1/metrics` path.

Perhaps what we need to do here is have a checkbox that is checked by default that says "use default port and path" which appends the port and path unless unchecked.

I'm open to other ideas too if you have any suggestions!

Kaotic3:
I would have thought a checkbox that was "advanced" would be better as hiding stuff behind reverse proxies, and therefore bypassing the standard language, sounds like advanced stuff to me!

It also allows for a simple "if (!advanced)" for making the proposed string variable changes.

ajhollid:
> I would have thought a checkbox that was "advanced" would be better as hiding stuff behind reverse proxies, and therefore bypassing the standard language, sounds like advanced stuff to me!
> 
> It also allows for a simple "if (!advanced)" for making the proposed string variable changes.

I don't think anything that directly modifies a users input should be hidden from them, that will lead to frustration.

I'm ok with including it as a default setting since it's going to cover the use case for most users, but I feel it should be visible immediately to the user and it should be clear that their input is being modified.

Kaotic3:
I feel like my previous post wasn't clear enough.

The default language is shown as a placeholder.  If you are going to do something different, then that would be advanced.

The code above doesn't change your input, unless it needs to be changed to match the default requirement.

Anything you want to add outside of that can be added as an advanced feature.  I updated because I Noted that the instructions are not clear in this regard, they just say:

Server URL:

Enter the URL of the host you want to monitor. Ensure the server is running the Checkmate Monitoring Agent to start data collection.

It doesn't actually set out that you need to add more - that appears only to exist in the placeholder text?

ajhollid:
> I feel like my previous post wasn't clear enough.
> 
> The default language is shown as a placeholder. If you are going to do something different, then that would be advanced.
> 
> The code above doesn't change your input, unless it needs to be changed to match the default requirement.
> 
> Anything you want to add outside of that can be added as an advanced feature. I updated because I Noted that the instructions are not clear in this regard, they just say:
> 
> Server URL:
> 
> Enter the URL of the host you want to monitor. Ensure the server is running the Checkmate Monitoring Agent to start data collection.
> 
> It doesn't actually set out that you need to add more - that appears only to exist in the placeholder text?

I do understand the point you're making here, however my concerns are mainly

1.  It's a common use case to hide APIs behind reverse proxies.  If I want `www.mysite.com/metrics` to point to 'whatever:59232/api/v1/metrics`, I don't want to enter `www.mysite.com/metrics` as my URL and have `:59232/api/v1/metrics` appended.

2.  The port of Caputre isn't guaranteed to stay at 59232 and the path isn't guaranteed to be `/api/v1/metrics` forever. 

The second point isn't as big a concern as the first, but it's something to keep in mind.

I do definitely see the value in appending the path and port for most users, but there are enough users that are going to be using reverse proxies that I believe it makes sense for it to be front and center that a port and path are appended by default if they are not specified, along with the option to disable that.

Ideally that would be a persistent setting like a checkbox so the user only has to disable it once if they don't want that feature.

In short yes I agree, it's a good idea, but I would like it to be clear that it is happening and an easy opt-out be available for the user.

client/src/Pages/Infrastructure/Create/index.jsx
```diff
@@ -89,8 +89,17 @@ const CreateInfrastructureMonitor = () => {
 	useEffect(() => {
 		if (isCreate || !monitor) return;
 
+		// Clean URL including any trailing slashes
```
Owaiseimdad, line 92: Possible to extract a method here where we can sent the url and it gives us the desired url, you can name it like "normalizeMonitorUrl". It can also be a utility itself, which anyone can use in future.

Owaiseimdad (changes requested):
Looks good but we can make somethings reusable in the component.

Br0wnHammer:
@Kaotic3 any update on this?

ajhollid:
Closing as inactive

Issue #2452, opened 2025-06-14
Related to #684 the Rows Per Page Request Infrastructure Page Missing them.


Kaotic3:
**Is your feature request related to a problem? Please describe.**
In [#684](https://github.com/bluewave-labs/Checkmate/pull/684) - you created the ability to change the number of rows per page.

It is not persistent.

**Describe the solution you'd like**
Persistence in the number of rows, so that once set, it remains that way.

**Describe alternatives you've considered**
Not applicable.

**Additional context**
I feel that this is likely already a global variable - you just need to make it so that when the user updates that variable, it is written as that number, until it changes - the user can always set it back to 5 if they want...

ajhollid:
This is a Redux persistance issue, due to the initial redux state being modified the application now expects a value in the UI state for `rowsPerPage`  

Clearing your browsers application data will reset Redux and load the initial state correctly.  In addition I pushed a fix earler to provide default fallback values in the event of Redux problems.

Thanks for pointing out the issue, it should now be resolved!

PR #2453, opened 2025-06-14
Feature/persist rows setting

Kaotic3:
## Describe your changes

I made a feature request here: #2452 for the rows to be persistent on the Infrastructure Monitor page.

I then noticed that it was in fact persistent on the Uptime Monitor page, which meant that the code already existed to do it it just wasn't implemented.

I applied the code that already existed by creating a table for Infrastructure and then mimicking the code.

It should be pretty straight forward, as technically, it is your code, I just copied some of it and pasted it elsewhere.

### DIFF
--- a/client/src/Features/UI/uiSlice.js
+++ b/client/src/Features/UI/uiSlice.js
@@ -16,6 +16,9 @@ const initialState = {
 	maintenance: {
 		rowsPerPage: 5,
 	},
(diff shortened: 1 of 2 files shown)


Br0wnHammer (approved):
LGTM! Thankyou for your contribution @Kaotic3 🚀

Issue #2454, opened 2025-06-14
Clickable Links on Monitor Pages.


Kaotic3:
**Is your feature request related to a problem? Please describe.**

There is no way to visit a monitored site directly from Checkmate.

**Describe the solution you'd like**

It would be nice to be able to pass a "secondary" url that made some of the labels "clickable" to that URL, opening in a new window.

For example it could be the Green Status box or the IP address as the top, this look is replicated on the Uptime and the Infrastructure look and I created a picture here:

![Image](https://github.com/user-attachments/assets/45ae533f-8e52-4150-9697-cdc92e82ed51)

**Describe alternatives you've considered**

I don't think there is an alternative.  The idea struck me as I use Homarr, and it has a lightweight "uptime" check, by allowing a secondary URL to "ping".

Br0wnHammer:
Thankyou for opening up an issue and creating a PR @Kaotic3.
@gorkem-bwl can you take a look and see if this can be feasible or can elevate Checkmate.

FluffyNumber1:
Can I take on this issue?

anjupathak03:
Hi, I would like to revive this feature against the current develop branch. PR #2456 was closed after the backend refactor, and there has not been a follow-up for many months. I plan to add an optional application link URL to the current TypeScript monitor model and form, validate HTTP(S) URLs on both client and server, and render the link safely in the monitor header with a tooltip. I will keep the change focused and add backend validation coverage.

PR #2456, opened 2025-06-15
Feature: URL links on infrastructure page

Kaotic3:
## Describe your changes

I made it so that you can add an optional URL on the Infrastructure Create or Modify pages, which would then become a clickable link on the status page for that monitor.

#2454 

```Javascript
<div>{t('add')}</div>
```

Here are some screen shots of the changes to the UI:

The infrastructure screen has not changed:

image

This is what it looks like if you do not add an additional URL:

image

This is what it looks like if you edit the monitor that you have:

image

Add a "url" for your application:

image

How it looks after you add it:

image

This link opens up your application on a new page.

I feel the same could be added to the  Uptime page too, but wanted to try and keep this rather large change as direct as possible.  The monitor.link.url now exists though, so could be added elsewhere?

### DIFF
--- a/client/src/Pages/Infrastructure/Create/index.jsx
+++ b/client/src/Pages/Infrastructure/Create/index.jsx
@@ -71,6 +71,7 @@ const CreateInfrastructureMonitor = () => {
 	const [infrastructureMonitor, setInfrastructureMonitor] = useState({
 		url: "",
 		name: "",
+		linkUrl: "",
 		notifications: [],
 		notify_email: false,
 		interval: 0.25,
@@ -92,6 +93,7 @@ const CreateInfrastructureMonitor = () => {
 		setInfrastructureMonitor({
 			url: monitor.url.replace(/^https?:\/\//, ""),
 			name: monitor.name || "",
+			linkUrl: monitor.link?.url || "",
 			notifications: monitor.notifications,
 			interval: monitor.interval / MS_PER_MINUTE,
 			cpu: monitor.thresholds?.usage_cpu !== undefined,
@@ -127,6 +129,7 @@ const CreateInfrastructureMonitor = () => {
 				infrastructureMonitor.name === ""
 					? infrastructureMonitor.url
 					: infrastructureMonitor.name,
+			linkUrl: infrastructureMonitor.linkUrl,
 			interval: infrastructureMonitor.interval * MS_PER_MINUTE,
 			cpu: infrastructureMonitor.cpu,
 			...(infrastructureMonitor.cpu
@@ -172,6 +175,7 @@ const CreateInfrastructureMonitor = () => {
 			usage_disk,
 			temperature,
 			usage_temperature,
+			linkUrl,
 			...rest
 		} = form;
 
@@ -185,6 +189,7 @@ const CreateInfrastructureMonitor = () => {
 		form = {
 			...(isCreate ? {} : { _id: monitorId }),
 			...rest,
+			...(linkUrl ? { link: { url: linkUrl } } : {}),
 			description: form.name,
 			type: "hardware",
 			notifications: infrastructureMonitor.notifications,
@@ -329,6 +334,17 @@ const CreateInfrastructureMonitor = () => {
 							value={infrastructureMonitor.name}
 							onChange={onChange}
 							error={errors["name"]}
+						/>
+						<TextInput
+							type="text"
+							id="linkUrl"
+							name="linkUrl"
+							label={t("infrastructureLinkUrlLabel")}
+							placeholder="https://example.com"
+							isOptional={true}
+							value={infrastructureMonitor.linkUrl}
+							onChange={onChange}
+							error={errors["linkUrl"]}
 						/>
 						<TextInput
 							type="text"
--- a/client/src/Validation/validation.js
+++ b/client/src/Validation/validation.js
@@ -374,6 +374,9 @@ const infrastructureMonitorValidation = joi.object({
 	name: joi.string().trim().max(50).allow("").messages({
 		"string.max": "This field should not exceed the 50 characters limit.",
 	}),
+	linkUrl: joi.string().uri().allow("").messages({
+		"string.uri": "The link URL is not valid.",
+	}),
 	secret: joi.string().trim().messages({ "string.empty": "This field is required." }),
 	usage_cpu: joi.number().messages({
 		"number.base": THRESHOLD_COMMON_BASE_MSG,
--- a/client/src/Components/MonitorStatusHeader/index.jsx
+++ b/client/src/Components/MonitorStatusHeader/index.jsx
@@ -40,7 +40,17 @@ const MonitorStatusHeader = ({ path, isLoading = false, isAdmin, monitor }) => {
 						component="h2"
 						variant="monitorUrl"
 					>
-						{monitor?.url?.replace(/^https?:\/\//, "") || "..."}
+						{monitor?.link?.url ? (
+							<a
+								href={monitor.link.url}
+								target="_blank"
+								rel="noreferrer"
+							>
+								{monitor?.url?.replace(/^https?:\/\//, "") || "..."}
+							</a>
+						) : (
+							monitor?.url?.replace(/^https?:\/\//, "") || "..."
+						)}
 					</Typography>
 					<Dot />
 					<Typography>
--- a/client/src/Hooks/monitorHooks.js
+++ b/client/src/Hooks/monitorHooks.js
@@ -334,6 +334,7 @@ const useUpdateMonitor = () => {
 				...(monitor.type === "hardware" && {
 					thresholds: monitor.thresholds,
 					secret: monitor.secret,
+					...(monitor.link ? { link: monitor.link } : {}),
 				}),
 			};
 			await networkService.updateMonitor({
--- a/client/src/locales/en.json
+++ b/client/src/locales/en.json
@@ -434,6 +434,7 @@
 	"infrastructureServerUrlLabel": "Server URL",
 	"infrastructureDisplayNameLabel": "Display name",
 	"infrastructureAuthorizationSecretLabel": "Authorization secret",
+	"infrastructureLinkUrlLabel": "Link URL",
 	"gb": "GB",
 	"mb": "MB",
 	"mem": "Mem",
--- a/server/db/models/Monitor.js
+++ b/server/db/models/Monitor.js
@@ -87,6 +87,12 @@ const MonitorSchema = mongoose.Schema(
 		secret: {
 			type: String,
 		},
+		link: {
+			type: {
+				url: { type: String },
+			},
+			_id: false,
+		},
 	},
 	{
 		timestamps: true,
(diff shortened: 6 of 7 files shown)


client/src/Pages/Infrastructure/Create/index.jsx
```diff
@@ -330,6 +335,17 @@ const CreateInfrastructureMonitor = () => {
...
+						<TextInput
+							type="text"
+							id="linkUrl"
+							name="linkUrl"
+							label={t("infrastructureLinkUrlLabel")}
+							placeholder="https://example.com"
+							isOptional={true}
+							value={infrastructureMonitor.linkUrl}
+							onChange={onChange}
+							error={errors["linkUrl"]}
```
ajhollid, line 347: `error` takes a `Boolean`, not a `String`, so this value needs to either explicitly be set to a `Boolean` value

`error={errors["linkUrl"] ? true : false}`

It is also missing the `helperText` prop which is needed to display the error message, this does tkae a `String`:

`helperText={errors["linkUrl"]}`

gorkem-bwl:
Quickly glanced on the phone and two things:

- we need to fix the link color.
- we need to identify what a link is. User might be confused why (even optional) they are asked for another url. The explanation can be under the text entry field

Kaotic3:
To hopefully meet the requirements:

"error={errors["linkUrl"] ? true : false}"

This was optional, so I followed the chunk above: "error={errors["name"]  - I thought that was how you were handling optional values, if this is incorrect let me know.

"helperText={errors["linkUrl"]}"

Oh that sent me down a rabbit hole...URL validation wasn't what I thought it was - so I improved it - I moved that long Git regex to the top as an inline const - and I amended it slightly to handle the linkUrl in particular so that it no longer excludes local IPs but otherwise acts the same for a strong validation of the URL.

(I was not clear here, the Regex you had for your validations, is unchanged - I just moved it to the top and made it inline.  I amended it and implemented it later purely for the linkUrl and nothing else - just no-one likes reading Regex so I should be really clear here).

"we need to fix the link color."

That is just the default, I have not changed it, as I don't know which colour you want - but assuming that would be a simple thing to fix.

"we need to identify what a link is. User might be confused why (even optional) they are asked for another url. The explanation can be under the text entry field"

I added some language to the en.json in locales that goes under the "General Settings":

Here you can select the URL of the host. A friendly name and a URL that will open an application's web page and the authorization secret to connect to the server agent.

The server you are monitoring must be running Capture


You will note that I took out "the" as well, because it either needs to be "the Capture agent" or simply "running Capture" - "running the Capture" that is just not right.

gorkem-bwl:
I am good with your proposals here. When it comes to link, it could be the same color as the normal text, but when user hovers on it, the styling can be the same as the one under Settings > About

image

Can you come up with another screenshot when you fix the link? - to see how it looks + the description.

gorkem-bwl:
Rethinking again, why don't we use the same URL and need a different URL to check the service and add another section for this in settings? We can always add the link to the checked URL.

Kaotic3:
This looks like this, it is cleaner, but I am not sure people will recognise it as a link...if you know its there, good but I don't think it is clear that it is a link...I don't have an answer to that - not really a UI guy - but it is an observation.

image


You do get the URL, down in the bottom left corner, but it's not that clear to everyone that it will be there I suspect.  This is what you see when you hover over the link in the image above.


image

Kaotic3:
I did think myself - why not make the Name the link to the service, and leave the URL for Capture as it is - not a link....I just wanted to keep it neat.

gorkem-bwl:
> I did think myself - why not make the Name the link to the service, and leave the URL for Capture as it is - not a link....I just wanted to keep it neat.

Yep, let's remove the extra settings.

Re: image: This is neat, I'll just need to figure out the new underline color and we are done here.

Thanks!

Kaotic3:
Do you think replacing the "url" with the "linkUrl" under the name makes sense?

Don't strip out of the http(s) either and have it as the linkUrl - the api/v1/metrics - doesn't really tell you anything that you didn't already know - as you had to enter it on the page to create the monitor and having the link there with a http references as entered - would make it more obviously a link - and in most cases applications come with a port number and even an /admin /login page (all of which is accepted by the linkUrl). 

I think that might work - and I think leaving it as this colour would be fine then too - as it would match other URL links in the App and would obviously be a link with the http(s) and the application address.

That would look like this:

image

gorkem-bwl:
> Do you think replacing the "url" with the "linkUrl" under the name makes sense?
> 
> Don't strip out of the http(s) either and have it as the linkUrl - the api/v1/metrics - doesn't really tell you anything that you didn't already know - as you had to enter it on the page to create the monitor and having the link there with a http references as entered - would make it more obviously a link - and in most cases applications come with a port number and even an /admin /login page (all of which is accepted by the linkUrl).
> 
> I think that might work - and I think leaving it as this colour would be fine then too - as it would match other URL links in the App and would obviously be a link with the http(s) and the application address.


If you mean to include just the URL part (stripping down the /api/... plus the port) would make sense, yes. The link can go to the URL itself and not the port/directory. When the user hovers on the URL then the URL part would be underlined (meaning it's a link to the server).

Am I on the same page with you? :) 

image

Kaotic3:
Not sure, my view would be that under the Name of the monitor you would have:

http ://192.168.1.123:5050/web

or

https ://myhosted-item.domain.com/web

Which would then be a clickable URL that took you to the UI for whatever application you were monitoring.  It is optional, if you are monitoring a ping to a machine with 10-hosted items, then add nothing and you get what you get now.

I think the whole http at the start provides enough information for the people using the site to understand that this is a link (and separate it from the standard non http language), there is a fair amount of set-up and knowledge required...that I feel that this assumption isn't expecting too much.

gorkem-bwl:
> Not sure, my view would be that under the Name of the monitor you would have:
> 
> http ://192.168.1.123:5050/web
> 
> or
> 
> https ://myhosted-item.domain.com/web
> 
> Which would then be a clickable URL that took you to the UI for whatever application you were monitoring. It is optional, if you are monitoring a ping to a machine with 10-hosted items, then add nothing and you get what you get now.
> 
> I think the whole http at the start provides enough information for the people using the site to understand that this is a link (and separate it from the standard non http language), there is a fair amount of set-up and knowledge required...that I feel that this assumption isn't expecting too much.

My initial thought is not to introduce another settings for this and be able to directly link (part of the) URL but looks like this is apparently a bit complex due to different setups. 

The screenshot below looks ok to me. For this one, can you please modify your PR so that when you hover on it, there is a tooltip saying something along the lines of "Go to this URL on another tab". Our charts have this tooltip but I doubt we have a relevant component, so you may want to add that simple tooltip component as well. This will bring extra clarity. WDYT?

image

client/src/Validation/validation.js
```diff
@@ -374,6 +338,21 @@ const infrastructureMonitorValidation = joi.object({
 	name: joi.string().trim().max(50).allow("").messages({
 		"string.max": "This field should not exceed the 50 characters limit.",
 	}),
+	linkUrl: joi
+		.string()
+		.allow("")
+		.trim()
+		.custom((value, helpers) => {
+			const localUrl =
```
Owaiseimdad, line 346: Why is the so complicated? Why not use the same regex used in url?

If we are going to use this regex alot, lets create a common method which validates the urls.

Owaiseimdad (changes requested):
Can we maintain only one regex for urls? The local-url regex is something I am not able to understand.

client/src/Validation/validation.js
Kaotic3: It isn't the same Regex.

The "const urlRegex" which I moved to be a constant at the top, actually excludes locally hosted URLs.  The "localRegex" is a version which includes that local range, as you require local IPs.

My initial plan was to make this a single const regex at the top and use the cleaner version through-out, but I realised that it simply didn't work for self-hosted links.

Owaiseimdad: Okay, thanks for the explaination @Kaotic3 , possible to explain what the end goal of this validation.

Also can we have a const where we have predefined url patterns something like this.
`
// Extract common URL validation patterns
const URL_PATTERNS = {
  comprehensive: new RegExp(
    "^" +
      "(?:(?:https?|ftp):\\/\\/)?" +
      "(?:" +
        "(?:[1-9]\\d?|1\\d\\d|2[01]\\d|22[0-3])" +
        "(?:\\.(?:1?\\d{1,2}|2[0-4]\\d|25[0-5])){2}" +
        "(?:\\.(?:[1-9]\\d?|1\\d\\d|2[0-4]\\d|25[0-4]))" +
        "|" +
        "(?:" +
          "(?:" +
            "[a-z0-9\\u00a1-\\uffff]" +
            "[a-z0-9\\u00a1-\\uffff_-]{0,62}" +
            ")?" +
            "[a-z0-9\\u00a1-\\uffff]\\." +
          ")+" +
          "(?:[a-z\\u00a1-\\uffff]{2,}\\.?)" +
        ")" +
      "(?::\\d{2,5})?" +
      "(?:[/?#]\\S*)?" +
      "$",
    "i"
  ),
  simple: /^(https?:\/\/)?(([0-9]{1,3}\.){3}[0-9]{1,3}|[\da-z\.-]+)(\.[a-z\.]{2,6})?(:(\d+))?([\/\w \.-]*)*\/?$/i
};

`

Just looking at the possibilty of this.

Kaotic3:
The end goal is really just to accept any valid URL, including local ranges and any additional pointers - in one of my examples above I provided the link to Pi-Hole, a very common self-hosted application, but to reach it, you must have the IP / Domain - but also /admin/login - without those parts, you receive a 404 error on your link - given our validation step requires a valid URL, we needed a Regex that accepted all of it.

You can of course refactor it to a separate utility and use a flag instead to return one or the other Regex, I didn't take it that far as my view of contributing is always to make minimal changes to ensure that people can visualise and accept those changes without the additional concern of breaking changes due to large scale refactoring - refactoring can always come along latter on - and is going to be required (in my view) because this is an option that many people would probably like on the Uptime / Pagespeed monitors as well as the Infrastructure monitors.

Once the standard is nailed down, then it may be a matter of implementing across all sections and thus doing some major refactoring to make that as simple as possible.  There is a lot of repeated code that I saw that I think could be factored out and replaced by a single line with a flag, which would reduce the overall complexity of individual pages.

@gorkem-bwl that looks a lot more like my original PR, without the "about page" formatting and hover tool-tip.

My only questions are:

1. Why do we want to display the URL for the monitor (/api/v1/metric) is that of any value on this page?
My thinking here is that the Capture is a background task, there is no interacting with it from a user's perspective.
2. As mentioned by @ajhollid - people can put the Capture behind proxies so any value here may bear no relation to the application, it may not have the same domain or ip address.
3. Without the http(s) moniker at the start - it isn't immediately obvious that it is a link and it could easily be missed.

That is my thinking about why I proposed it in the way that I did - that the URL would reflect the link to the application, not to the Capture, and that it would be immediately recognisable as a link with the http(s) prefix.

Let me know which way you decide to go and I will make the necessary changes.

ajhollid:
Closing this one as the BE has gone through a significant refactor resulting in numerous merge conflicts.

If you'd like to keep working on this one please rebase onto `develop` and we can go fromt here.  Apologies for the hassle and appreciate the contribution!

PR #2465, opened 2025-06-16
fix: send test notifications

ajhollid:
This PR updates the monitor details header.  It only allows for sending a test email, however a user is more likely to want to test all their notifications

- [x] Replace send test-email with test-notification
- [x] Add controller and route for test all notificaitons
- [x] Add notification service method to test all notifications

image
  

## Summary by CodeRabbit

- **New Features**
  - Added the ability to test all notifications for a monitor at once from the user interface.
  - Introduced a new button labeled "Send test notifications" in the monitor details control header.
  - Success and error messages are now shown when testing all notifications.

- **Bug Fixes**
  - Improved error handling and messaging when testing notifications.

- **Documentation**
  - Updated English text to reflect the new "Send test notifications" feature.

### DIFF
--- a/server/routes/notificationRoute.js
+++ b/server/routes/notificationRoute.js
@@ -14,6 +14,7 @@ class NotificationRoutes {
 		this.router.post("/trigger", this.notificationController.triggerNotification);
 
 		this.router.post("/test", this.notificationController.testNotification);
+		this.router.post("/test/all", this.notificationController.testAllNotifications);
 
 		this.router.post("/", this.notificationController.createNotification);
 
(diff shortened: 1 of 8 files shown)


server/routes/notificationRoute.js
```diff
@@ -14,6 +14,7 @@ class NotificationRoutes {
 		this.router.post("/trigger", this.notificationController.triggerNotification);
 
 		this.router.post("/test", this.notificationController.testNotification);
```
Owaiseimdad, line 16: @ajhollid we can remove this route as we will not be using it.

Owaiseimdad (approved):
@ajhollid we can also clean code which is not being used after this implementation.

server/routes/notificationRoute.js
ajhollid: @Owaiseimdad this route is still used when creating a new notification

PR #2466, opened 2025-06-16
fix: hardware notifications

ajhollid:
Hardware notifications were not updated with regular notificiations when notification channels were implemented.

- [x] Update hardware notificiations to use notification channels

## Summary by CodeRabbit

- **New Features**
  - Added support for Discord as a notification channel, including UI options and configuration for Discord webhooks.
  - Introduced localized text for Discord notification settings.

- **Improvements**
  - Enhanced notification type handling for greater clarity and consistency by switching to string-based logic.
  - Improved validation and configuration logic for notification channels.

- **Bug Fixes**
  - Corrected Slack notification type value for accurate channel selection.

- **Backend Updates**
  - Discord added as a valid notification platform in the backend.
  - Notification creation now automatically associates notifications with the current user and team.
  - Hardware alert notifications now include detailed alert messages and improved formatting.
  - Simplified notification creation validation by removing redundant user and team ID requirements.

### DIFF
--- a/server/service/notificationService.js
+++ b/server/service/notificationService.js
@@ -95,6 +95,19 @@ class NotificationService {
 		return MESSAGE_FORMATTERS[platform](messageText, chatId);
 	}
 
+	formatHardwareNotificationMessage = (
+		monitor,
+		status,
+		platform,
+		chatId,
+		code,
+		timestamp,
+		alerts
+	) => {
+		const messageText = alerts.map((alert) => alert).join("\n");
+		return MESSAGE_FORMATTERS[platform](messageText, chatId);
+	};
+
 	sendTestWebhookNotification = async (notification) => {
 		const config = notification.config;
 		const response = await this.networkService.requestWebhook(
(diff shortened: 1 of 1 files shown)


server/service/notificationService.js
```diff
@@ -247,19 +273,31 @@ class NotificationService {
...
-		const message = this.formatNotificationMessage(
-			monitor,
-			status,
-			platform,
-			null,
-			code, // Pass the code field directly
-			networkResponse.timestamp
-		);
-
+		let message;
```
Owaiseimdad, line 280: @ajhollid are we not repeating a similar pattern here with just one null value. Because if something comes up in future then we have to do monitor.type for it again.

ajhollid: Yes, this notification sertvice needs to be cleaned up and refactored.

Most of this is leftover from the previous implementation prior to notificaiton channels.  

We can refactor this for simplification and maintainability in a new PR.

PR #2469, opened 2025-06-16
Refactor: useFetchChecks

Br0wnHammer:
## Describe your changes

Replaced single `useFetchChecks` hook with conditional logic using `useFetchChecksByMonitor` and `useFetchChecksTeam`, eliminating unnecessary state management and preventing duplicate API requests.

```Javascript
<div>{t('add')}</div>
```

## Summary by CodeRabbit

- **Refactor**
  - Improved how incident check data is fetched and managed, resulting in more streamlined and maintainable code.
  - Updated data fetching in the Uptime Details page for consistency.
- **Bug Fixes**
  - Enhanced reliability in loading check data based on the selected monitor or team, reducing potential errors and improving user experience.


client/src/Pages/Incidents/Components/IncidentTable/index.jsx
```diff
@@ -26,26 +27,40 @@ const IncidentTable = ({
...
+			dateRange,
+			filter: filter,
+			page: page,
+			rowsPerPage: rowsPerPage,
+		});
+
+	const checks = selectedMonitor === "0" ? checksTeam : checksMonitor;
+	const checksCount = selectedMonitor === "0" ? checksCountTeam : checksCountMonitor;
+	const isLoading = selectedMonitor === "0" ? isLoadingTeam : isLoadingMonitor;
+	const networkError = selectedMonitor === "0" ? networkErrorTeam : networkErrorMonitor;
```
ajhollid, line 64: I think we can ditch these and just use logical `or` operator `||` to deal with this.

`const isLoading = isLoadingTeam || isLoadingMonitor`

Easier to follow I think

client/src/Hooks/checkHooksMonitor.js
```diff
@@ -21,40 +20,25 @@ const useFetchChecks = ({
 
 	useEffect(() => {
 		const fetchChecks = async () => {
-			if (!type && !teamId) {
+			if (!type) {
 				return;
 			}
```
ajhollid: Rather than try to return early based on the params given to the hook, why don't we have them take an `enabled` param that is set to `true` by default.  Then we can turn it off as needed, but if nothing is provided then it runs as expected, ie for the `Uptime` page

client/src/Hooks/checkHooksMonitor.js
```diff
@@ -66,20 +50,9 @@ const useFetchChecks = ({
...
-		page,
-		rowsPerPage,
-	]);
+	}, [monitorId, type, status, sortOrder, limit, dateRange, filter, page, rowsPerPage]);
 
 	return [checks, checksCount, isLoading, networkError];
 };
 
-export { useFetchChecks };
+export { useFetchChecksByMonitor };
```
ajhollid: Let's combine both of these hooks into one file for organizational purposes.

Let's have `checkHooks` and have it export both hooks `export {useFetchChecksByMonitor, useFetchChecksByTeam}`

client/src/Pages/Incidents/Components/IncidentTable/index.jsx
```diff
@@ -26,26 +27,40 @@ const IncidentTable = ({
...
+			limit: null,
+			dateRange,
+			filter: filter,
+			page: page,
+			rowsPerPage: rowsPerPage,
+		});
+
+	const [checksTeam, checksCountTeam, isLoadingTeam, networkErrorTeam] =
+		useFetchChecksTeam({
+			teamId: "placeholder",
```
ajhollid: We definitely need to do something about this `useFetchChecksByTeam` should not take a `teamId` at all anymore.

ajhollid (changes requested):
Nice work, this is a good improvement!

I think we can make it a bit better though by passing an `enabled` param to the two hooks with a default value of `true` rather than trying to turn the hook on or off by using the other parameters like side effects.

Let's be explicit and just turn it off when we don't want it to run.

Other than that, can we please move the two hooks into one file and have it export both?  It would be good to have all related hooks organized togehter thusly.

client/src/Pages/Incidents/Components/IncidentTable/index.jsx
Br0wnHammer: Alex, this won't work since checksTeam will have an empty value and || operator will still prefer it as the value is not falsy.

ajhollid: OK let's use logical or just for the loading one then as we definitely want loading to show when either opeartion is in progress.

ajhollid (changes requested):
Uptime details is currently broken, and there is a duplicate network request being made to `checks` endpoint.

Check your network tab and you'll see what's going on.

ajhollid (approved):
Looks good!  Thanks for taking care of this one :rocket:

Issue #2301, opened 2025-05-18
Network Monitoring on the Server


ProxyBlocks:
**Is your feature request related to a problem? Please describe.**
Right now there’s no way to see network usage over time. Would be super helpful to track traffic spikes or weird behavior like we already can with CPU and RAM.

**Describe the solution you'd like**
Add a traffic history graph – in/out bandwidth per Server. Like CPU / RAM 24h, 7d, 30d views.

**Describe alternatives you've considered**
Tianji has this, but only for live traffic, no historical view – so it’s not helpful for spotting past spikes or analyzing trends.

**Additional context**
Thanks for this project, it’s very helpful and lightweight 
I like it :)

gorkem-bwl:
So Capture agent can get those values for us,meaning we can implement this in the next release. Thanks for opening this feature request.

 
![image](https://github.com/user-attachments/assets/eb422e06-72fd-45e7-8603-d5cc0b6aedb9)

gorkem-bwl:
Capture agent can send network details under `/metrics/net`(related PR: https://github.com/bluewave-labs/capture/pull/67)

```
- BytesSent: Total bytes sent from the server.
- BytesRecv: Total bytes received.
- PacketsSent: Number of packets sent.
- PacketsRecv: Number of packets received.
- ErrIn: Incoming packets with errors.
- ErrOut: Outgoing packets with errors.
- DropIn: Incoming packets that were dropped.
- DropOut: Outgoing packets that were dropped.
- FIFOIn: Incoming packets dropped due to full buffer.
- FIFOOut: Outgoing packets dropped due to full buffer.
```


Attached is a mockup screenshot as well.

<img width="858" alt="Image" src="https://github.com/user-attachments/assets/c92a01e4-e34e-48b2-a686-c7ba1d180ffc" />

Owaiseimdad:
@gorkem-bwl do you want a new page for this called network??

gorkem-bwl:
> @gorkem-bwl do you want a new page for this called network??
> 

That's right.

gorkem-bwl:
@Owaiseimdad the new Capture version has network monitoring features. FYI.

Owaiseimdad:
The UI has been done now @gorkem-bwl .

https://github.com/user-attachments/assets/17b758fd-10bf-4849-91a7-62480a6fbf6f

Let me know if there are any changes to be made.

gorkem-bwl:
[...]

PR #2489, opened 2025-06-18
Code for network db settings

Owaiseimdad:
## Describe your changes
1. Created NetworkCheck model to store network interface data (bytes sent/received, packets, errors, etc.)
2. Added networkCheckModule with createNetworkCheck and getNetworkChecksByMonitorId functions

Follows the same pattern as existing HardwareCheck implementation

#2301 

```Javascript
<div>{t('add')}</div>
```
- [] I took a screenshot or a video and attached to this PR if there is a UI change.

## Summary by CodeRabbit

- **New Features**
  - Introduced support for storing and retrieving network check data, including detailed network interface statistics and capture metadata.
  - Users can now view historical network checks filtered by monitor, with results sorted by most recent.


gorkem-bwl:
We can always hop on a call for the FE side :) @Owaiseimdad

server/db/mongo/modules/networkCheckModule.js
```diff
@@ -0,0 +1,46 @@
...
+			.sort({ createdAt: -1 })
+			.limit(limit);
+		return networkChecks;
+	} catch (error) {
+		logger.error({
+			message: error.message,
+			service: SERVICE_NAME,
+			method: "getNetworkChecksByMonitorId",
+			stack: error.stack,
+		});
```
ajhollid: If we log error her and rethrow we will log twice, once here and once in the error handling middleware.

We should set `error.service = SERVICE_NAME` and `err.method=getNetworkChecksByMonitorId` and then throw the error and let the middleware handle logging/

ajhollid (changes requested):
Code looks good, there is a double logging issue, see code review for details :+1:

server/db/mongo/modules/networkCheckModule.js
Owaiseimdad: Done this.

I refered to the hardware code so, there it is like, 

	const n = (await HardwareCheck.countDocuments({ monitorId })) + 1;
		const monitor = await Monitor.findById(monitorId);

		if (!monitor) {
			logger.error({
				message: "Monitor not found",
				service: SERVICE_NAME,
				method: "createHardwareCheck",
				details: `monitor ID: ${monitorId}`,
			});
			return null;
		}

		let newUptimePercentage;
		if (monitor.uptimePercentage === undefined) {
			newUptimePercentage = status === true ? 1 : 0;
		} else {
			newUptimePercentage =
				(monitor.uptimePercentage * (n - 1) + (status === true ? 1 : 0)) / n;
		}

		await Monitor.findOneAndUpdate(
			{ _id: monitorId },

client/src/locales/en.json
```diff
@@ -746,4 +746,4 @@
 	"settingsEmailPool": "Pool - Enable connection pooling",
 	"sendTestNotifications": "Send test notifications",
 	"selectAll": "Select all"
-}
```
Owaiseimdad: @ajhollid this is after I run format

server/db/mongo/modules/networkCheckModule.js
```diff
@@ -0,0 +1,38 @@
...
+
+const createNetworkCheck = async (networkCheckData) => {
+	try {
+		const { monitorId, status } = networkCheckData;
+		const n = (await NetworkCheck.countDocuments({ monitorId })) + 1;
+
+		const networkCheck = await new NetworkCheck({
+			...networkCheckData,
+			n,
+		});
```
ajhollid: Why are keeping a document count in the document?  This seems like an anti-pattern :thinking: 

What are we trying to achieve here?  

This doubles the number of DB operations per request, so there should be a very good reason to do this. n -> 2n is not good if avoidable.

ajhollid (commented):
Can you please walk me through what you're thinking of doing with the document count?

There may be a better way of achieving whatever the goal is than checking document count on every create check operation.

server/db/mongo/modules/networkCheckModule.js
Owaiseimdad: Apologies @ajhollid missed it and thought of putting it in. Took the reflect from hardware file but missed the calculation part.

PR #2494, opened 2025-06-20
fix: disable job on addition if monitor is paused

ajhollid:
Jobs that are created from a monitor in the Paused state should be set to the "disabled" state

- [x] Disable jobs when adding to queue if `isActive === false`, ie the monitor is paused.

## Summary by CodeRabbit

- **Bug Fixes**
  - New jobs are now automatically disabled if the associated monitor is inactive.


shanikauwu1:
I checked on the frontend, and it seems to work correctly.

PR #2496, opened 2025-06-20
Feat: Incidents Resolve BE

### LINKED ISSUE #2373: Button/function to clear the Incidents page
I particularly like to keep things clean, so I would love to see an option/function that would clean/hide old or already resolved incidents, I have had the system running for 2 days and I already have 28 incident pages due to testing

Br0wnHammer:
## Describe your changes

This PR primary focuses on the BE of the resolving incidents.

Fixes #2373 

```Javascript
<div>{t('add')}</div>
```

## Summary by CodeRabbit

- **New Features**
  - Added a "Resolved" filter option to the incidents interface, allowing users to view resolved incidents.
  - Enabled updating the acknowledgment status of individual and multiple incident checks directly from the interface.
- **Bug Fixes**
  - Improved accuracy of incident status filtering to ensure correct display of resolved and unresolved checks.
- **Localization**
  - Added English translation for the new "Resolved" filter label.


client/src/Pages/Incidents/Components/IncidentTable/index.jsx
```diff
@@ -52,7 +52,7 @@ const IncidentTable = ({
 			sortOrder: "desc",
 			limit: null,
 			dateRange,
-			filter: filter,
+			filter: filter === "resolved" ? "all" : filter,
```
ajhollid, line 56: I don't think this one makes sense, if a user selets "resolved" then they should only see checks where `ack === true`.

server/controllers/checkController.js
```diff
@@ -105,6 +109,53 @@ class CheckController {
...
+		try {
+			await updateCheckStatusBodyValidation.validateAsync(req.body);
+		} catch (error) {
+			next(handleValidationError(error, SERVICE_NAME));
+			return;
+		}
+
+		try {
+			const { checkId } = req.params;
+			const { ack } = req.body;
```
ajhollid, line 124: This is currently insecure, a user from a differnet team can update the checks of any other team.

There are two options

1.  Add the the teamID to the check model when it is created (preferrable)
2. Ftech the monitor that the Check belongs too and see if the user is part of that monitor's team.

I think it's better to add the `teamID` to the check, otherwise we need to do an extra database operation to get the teamID from the monitor.

server/controllers/checkController.js
```diff
@@ -105,6 +109,53 @@ class CheckController {
 		}
 	};
 
+	updateCheckStatus = async (req, res, next) => {
```
ajhollid: Let's rename this `ackCheck`, we don't want devs to think this method is for updating the `status` field

server/controllers/checkController.js
```diff
@@ -105,6 +109,53 @@ class CheckController {
...
+			return res.success({
+				msg: this.stringService.checkUpdateStatus,
+				data: updatedCheck,
+			});
+		} catch (error) {
+			next(handleError(error, SERVICE_NAME, "updateCheckStatus"));
+		}
+	};
+
+	updateAllChecksStatus = async (req, res, next) => {
```
ajhollid: Similarly, we can rename this `ackAllChecks`

server/controllers/checkController.js
```diff
@@ -105,6 +109,53 @@ class CheckController {
...
+		} catch (error) {
+			next(handleValidationError(error, SERVICE_NAME));
+			return;
+		}
+
+		try {
+			const { monitorId, target } = req.params;
+			const { ack } = req.body;
+
+			const updatedChecks = await this.db.updateAllChecksStatus(monitorId, ack, target);
```
ajhollid: Same for this DB method, `ackAllChecks`

server/routes/checkRoute.js
```diff
@@ -20,6 +20,10 @@ class CheckRoutes {
 
 		this.router.get("/:monitorId", this.checkController.getChecksByMonitor);
 
+		this.router.put("/:checkId", this.checkController.updateCheckStatus);
+
+		this.router.put("/:monitorId/:target", this.checkController.updateAllChecksStatus);
```
ajhollid: To be REST compliant routes should be `/path/identifier`, here we have `identifier/path`.  Let's switch those around.

We also don't want to pass a teamID to this route, the teamID should be extracted from the user's token.

We should then have two rotues

`PUT /api/v1/checks/monitor/:monitorId`
`PUT /api/v1/checks/team`

If you want to combine those in two one routeypou still can, have the param be optional and rename `target` to `path`:

`this.router.put("/:path/:monitorId?)`

server/db/mongo/modules/checkModule.js
```diff
@@ -236,6 +238,57 @@ const getChecksByTeam = async ({
...
+/**
+ * Update the acknowledgment status of all checks for a monitor or team
+ * @async
+ * @param {string} id - The monitor ID or team ID
+ * @param {boolean} ack - The acknowledgment status to set
+ * @param {string} target - The target type ('monitor' or 'team')
+ * @returns {Promise<number>}
+ * @throws {Error}
+ */
+const updateAllChecksStatus = async (id, ack, target) => {
```
ajhollid: These params can be renamed for clarity, see comment below for details

id -> monitorId
target -> path

server/db/mongo/modules/checkModule.js
```diff
@@ -236,6 +238,57 @@ const getChecksByTeam = async ({
...
+	try {
+		const updatedChecks = await Check.updateMany(
+			target === "monitor" ? { monitorId: id } : { teamId: id },
```
ajhollid: Team ID should be fetched from the user's token, which is already parsed and stored in `req.user`.

We don't want users to be able to specify what team they want to edit, they should only be able to edit documents belongning to their own team.

ajhollid (changes requested):
Looks good conceptually!

Some naming nitpicks here and there.

There's some security updates that need to be done and the the ack all route can be updated to be REST compliant.

With respect to the frontend I don't think the "acknowledged" fitler makes sense right now, please see my comments there for details :+1:

server/controllers/checkController.js
```diff
@@ -105,6 +112,55 @@ class CheckController {
 		}
 	};
 
+	ackCheck = async (req, res, next) => {
+		try {
+			await updateCheckStatusBodyValidation.validateAsync(req.body);
```
ajhollid: Validation should be renamed to match, this should be `ackCheckBodyValidation`

server/controllers/checkController.js
```diff
@@ -105,6 +112,55 @@ class CheckController {
...
+			});
+		} catch (error) {
+			next(handleError(error, SERVICE_NAME, "ackCheck"));
+		}
+	};
+
+	ackAllChecks = async (req, res, next) => {
+		try {
+			await updateCheckStatusParamValidation.validateAsync(req.params);
+			await updateAllChecksStatusBodyValidation.validateAsync(req.body);
```
ajhollid: Ditto for these validations

server/routes/checkRoute.js
```diff
@@ -20,6 +20,10 @@ class CheckRoutes {
 
 		this.router.get("/:monitorId", this.checkController.getChecksByMonitor);
 
+		this.router.put("/:checkId", this.checkController.ackCheck);
```
ajhollid: `verifyTeamAccess` team middleware should be applied here, please merge `develop` into your branch so you are up to date and have access to that middleware

ajhollid (changes requested):
`ackCheck` is still not access secured as far as I can see.

There is a `verifyTeamAccess` middleware that should be applied to this route to make sure that the user requesting `ackCheck` has the right to do so.

Otherwise I can use my token to ack any check I want from any monitor belonging to any team.

The validations should be renamed to match the updated method names :+1:

PR #2515, opened 2025-06-23
feat: logs

ajhollid:
This PR adds a logs component to the logs page

- [x] Implement caching of logs on BE
- [x] Add route/controller to fetch logs
- [x] Add a log component
- [x] Add a hook to fetch logs   

image

### DIFF
--- a/server/utils/logger.js
+++ b/server/utils/logger.js
@@ -74,15 +74,7 @@ class Logger {
 	 * @param {Object} config.details - Additional details.
 	 */
 	info(config) {
-		const logEntry = {
-			level: "info",
-			message: config.message,
-			service: config.service,
-			method: config.method,
-			details: config.details,
-			timestamp: new Date().toISOString(),
-		};
-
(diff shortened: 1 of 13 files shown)


server/utils/logger.js
```diff
@@ -88,11 +96,17 @@ class Logger {
 	 * @param {Object} config.details - Additional details.
 	 */
 	warn(config) {
-		this.logger.warn(config.message, {
+		const logEntry = {
```
Owaiseimdad: @ajhollid we are repeating the logEntry values alot which are same except the level key. Possible to put it out some where and only add the level?

Owaiseimdad (approved):
See if there is a possibility of changes in the logEntry object repetition.

server/utils/logger.js
ajhollid: I don't think this is worth refactoring, this will likely never be touched again :joy: 

**edit** nm did it anyway :P

PR #2522, opened 2025-06-24
feat: add stat boxes and timeframe header

ajhollid:
This PR updates the Uptime monitor details page empty view

If a monitor has no checks for the "recent" time period (the default time period) but _does_ have checks for times > recent, the page currently shows that "no checks exist".  This is not accurate as checks do exist, just not for the selected time period.

This PR adds the time frame header to the empty view so that users can select other time frames to see if there are checks there

image


shanikauwu1:
The stat boxes and timeframe for the empty view look good on my end.

PR #2530, opened 2025-06-24
feat: cascade on superuser delete

ajhollid:
This PR uses a pre delete middleware to cascade delete to

* monitors
* notifications
* teams
* team members

on delete of super admin account.


shanikauwu1:
According to llamapreview  there is a high issue to fixed that Null User Handling in Middleware

PR #2557, opened 2025-06-26
Fix/placeholder center align

### LINKED ISSUE #2548: Placeholder layout is too tall and misaligned on large screens
When there are no monitors in modules such as Uptime, PageSpeed, or Infrastructure, a placeholder is displayed to guide the user. However, on large screens (e.g., external LCD displays), this placeholder appears too tall, stretching all the way down the page and looking visually unbalanced.

karenvicent:
## Describe your changes

Centered the placeholder when there are no monitors in the modules, adjusted vertical spacing, and applied max-height constraints to avoid stretching on large LCD screens. Verified layout on regular laptop screens and larger external displays.

Fixes #2548 

```Javascript
<div>{t('add')}</div>
```
Captura de pantalla 2025-06-26 a la(s) 9 30 10 a m

Captura de pantalla 2025-06-26 a la(s) 9 31 09 a m

## Summary by CodeRabbit

* **Style**
  * Improved CSS readability by adding spacing between rule blocks and removing unnecessary top margin.
  * Updated layout to center the Fallback component content both vertically and horizontally.
  * Refined visual styling by moving border and background styles to a new inner container, enhancing overall appearance and structure.

* **Bug Fixes**
  * Corrected navigation button behavior to use the provided link for redirection.


client/src/Components/Fallback/index.jsx
```diff
@@ -67,113 +67,131 @@ const Fallback = ({
...
+				backgroundColor={theme.palette.primary.main}
+				overflow="hidden"
+				sx={{
+					display: "flex",
+					borderStyle: "dashed",
+					height: "60vh",
+					width: "45%",
+					minWidth: theme.spacing(200),
+					maxHeight: theme.breakpoints.values.sm,
+					maxWidth: theme.breakpoints.values.md,
```
ajhollid: This doesn't seem like the correct way to use break points to me, the breakpoints shoul dbe used to set specific values, ie if screenSize > theme.breakpoints.values.md then set some value for maxWidth.

This will also break if we change the value of the default breakoints, which is a very plausible situation we may face sooner rather than later.

client/src/Components/Fallback/index.jsx
```diff
@@ -67,113 +67,131 @@ const Fallback = ({
...
+				border={1}
+				borderColor={theme.palette.primary.lowContrast}
+				borderRadius={theme.shape.borderRadius}
+				backgroundColor={theme.palette.primary.main}
+				overflow="hidden"
+				sx={{
+					display: "flex",
+					borderStyle: "dashed",
+					height: "60vh",
+					width: "45%",
```
ajhollid: Why 60vh for height and width 45%?  Where did the values come from, and why not 65vh and 40% for example?

Willl these hold up on small screens as well as large screens?

client/src/Components/Fallback/index.jsx
```diff
@@ -67,113 +67,131 @@ const Fallback = ({
...
-					className="background-pattern-svg"
+				<Stack
+					className={`fallback__${title?.trim().split(" ")[0]}`}
+					alignItems="center"
+					gap={theme.spacing(20)}
 					sx={{
-						"& svg g g:last-of-type path": {
-							stroke: theme.palette.primary.lowContrast,
-						},
+						maxHeight: "70vh",
```
ajhollid: Again here, any particular reason for `70vh`?

ajhollid (commented):
Thanks for opening your first PR @karenvicent !

Visually it looks good to me, but I'm not a UI guy so @gorkem-bwl will have more advice here than I do.

The only major issue I see is with the use of breakpoint values as dimensions, this could very likely break down the road when we adjust those breakpoint values.  Please see my comments there for details.

Other than that I do have some questions about the specifc values chosen for some widths and heights, mostly I'm just not sure where they came from.

I'm not a frontend dev so these are my thougths from an outsider, the other FE guys will weigh in with more useful reviews I'm sure :joy:

client/src/Components/Fallback/index.jsx
karenvicent: Thanks so much for the feedback @ajhollid, I really appreciate it.
You're right about the misuse of "theme.breakpoints.values". I use as static dimension, so I'll replace those with "theme.spacing" or fixed values as needed.

client/src/Components/Fallback/index.jsx
karenvicent: As for the 60VH height and 45% width, I use them because the idea was yo make the box responsive while keeping visually centered, the requeriment was specifically focuses on laptops and bigger screens, smaller screens weren't taken into account.
I agree it would better to justify those numbers more clearly.

gorkem-bwl:
@karenvicent I think we better center the middle block horizontally as well, like this (e.g two red lines should be equal in length).

image

client/src/Components/Fallback/index.jsx
```diff
@@ -67,113 +67,127 @@ const Fallback = ({
...
+						<>
 							<Button
 								variant="contained"
 								color="accent"
 								sx={{ alignSelf: "center" }}
-								onClick={() => navigate("/uptime/bulk-import")}
+								onClick={() => navigate(link)}
 							>
-								{t("bulkImport.fallbackPage")}
+								Let's create your first {title}
```
mohicody, line 154: please use translation key.

client/src/Components/Fallback/index.jsx
```diff
@@ -67,113 +67,127 @@ const Fallback = ({
...
+			<Box
+				border={1}
+				borderColor={theme.palette.primary.lowContrast}
+				borderRadius={theme.shape.borderRadius}
+				backgroundColor={theme.palette.primary.main}
+				overflow="hidden"
+				sx={{
+					display: "flex",
+					borderStyle: "dashed",
+					height: { sm: "50vh", md: "60vh" },
```
mohicody: I think it would be better use theme breakpoints system here. 
like this : 
`` 
height: {
    xs: theme.spacing(60),
    sm: theme.spacing(70),
    md: theme.spacing(80)
  },
  ``

client/src/Components/Fallback/index.jsx
```diff
@@ -67,113 +67,127 @@ const Fallback = ({
...
-			backgroundColor={theme.palette.primary.main}
-			overflow="hidden"
 			sx={{
-				borderStyle: "dashed",
-				minHeight: "calc(100vh - var(--env-var-spacing-2) * 2)",
+				height: "100%",
+				width: "100%",
+				display: "flex",
+				justifyContent: "center",
+				alignItems: "center",
```
mohicody, line 78: what about padding?  I think it can be add for better visual balance on different screen sizes.

client/src/Components/Fallback/index.jsx
```diff
@@ -67,113 +67,134 @@ const Fallback = ({
...
 			}}
 		>
-			<Stack
-				className={`fallback__${title?.trim().split(" ")[0]}`}
-				alignItems="center"
-				gap={theme.spacing(20)}
+			<Box
```
Owaiseimdad, line 81: @karenvicent , let me reason for the file has too many changes is because we added a BOX component to wrap it all up?

client/src/Hooks/checkHooks.js
```diff
@@ -122,4 +122,30 @@ const useFetchChecksByMonitor = ({
...
+			} catch (error) {
+				setNetworkError(true);
+				createToast({ body: error.message });
+			} finally {
+				setIsLoading(false);
+			}
+		};
+
+		fetchSummary();
+	}, []);
```
Owaiseimdad: @karenvicent are we sure that we don't want to add dependencies?

client/src/Pages/Incidents/Components/StatusBoxes/StatusBox.jsx
```diff
@@ -0,0 +1,116 @@
...
+	let sharedStyles = {
+		position: "absolute",
+		right: 8,
+		opacity: 0.5,
+		"& svg path": { stroke: theme.palette.primary.contrastTextTertiary },
+	};
+
+	let color;
+	let icon;
+	if (status === "up") {
```
Owaiseimdad: @karenvicent , can we optimise this further, thinking a scenario where I might use the Icons. We can have a Factory pattern based component which returns based on what is the string given in props.

@ajhollid any thoughts?

server/db/mongo/modules/checkModuleQueries.js
```diff
@@ -0,0 +1,48 @@
+const buildChecksSummaryByTeamIdPipeline = ({ matchStage }) => {
```
Owaiseimdad: Formatting looks off, why?

client/src/Components/Fallback/index.jsx
karenvicent: Hi Owaise, thanks for the feedback!
I just want to clarify that the large diff you initially saw was because develop had been updated after I last pushed my changes, so those updates appeared in the PR until I merged develop into my branch.
Now that the branch is up to date, you should only see my actual changes:

I added a <Box> wrapper to help improve layout and spacing.

I adjusted placeholder sizes and spacing to make the layout look more balanced and centered on large screens.
No logic was modified — all changes are purely related to size, layout, and spacing adjustments.
Let me know if there’s anything else you'd like me to update!

client/src/Components/Fallback/index.jsx
```diff
@@ -67,113 +67,134 @@ const Fallback = ({
...
+					borderStyle: "dashed",
+					height: {
+						xs: theme.spacing(200),
+						sm: theme.spacing(300),
+						md: theme.spacing(300),
+					},
+					width: {
+						xs: theme.spacing(100),
+						sm: theme.spacing(210),
+						md: theme.spacing(250),
```
ajhollid: I realize you changed this at the request of another reviewer, but I feel your original approach using `vh` and `vw` was more reliable than this, this will break if we change the base unit of spacing, which is something like 2px at the moment I believe.

I apologize for the back and forth, but could you please revert this to the `vw`/`vh` values?

ajhollid (approved):

PR #2563, opened 2025-06-28
Fix/2471 infrastructure action menu

### LINKED ISSUE #2471: Fix - Infrastrucutre Action Menu
The infrastrucure monitor page action menu only shows "remove":

Image

It should at the very least also have "Details", "Configure", and "Pause"  similar to the Uptime action menu

Image

mohicody:
## Describe your changes

This PR addresses the issue by enhancing the infrastructure monitor action menu to match the functionality of the uptime monitor action menu.

## Changes:
Added "Details" option to navigate to monitor details
Added "Configure" option for admin users
Added "Pause/Resume" toggle functionality
Styled the "Remove" option using theme error color

Fixes #2471  

```Javascript
<div>{t('add')}</div>
```

Screenshot 2025-06-20 at 8 51 40 PM


client/src/Pages/Infrastructure/Monitors/Components/MonitorsTableMenu/index.jsx
```diff
@@ -28,7 +30,21 @@ const InfrastructureMenu = ({ monitor, isAdmin, updateCallback }) => {
...
+	const { t } = useTranslation();
+
+	// Update local state when monitor prop changes
+	useEffect(() => {
+		// Ensure we have a valid monitor object with default values for missing properties
+		if (monitor) {
+			setLocalMonitorState({
+				...monitor,
+				status: monitor.status || "active", // Default to active if status is undefined
+			});
```
ajhollid: A monitor with status "undefined` is in the pending state, not the active state.

The value of `status` is also a `boolean`, not a `string`.

DB model for reference:

```
const MonitorSchema = mongoose.Schema(
	{
		userId: {
			type: mongoose.Schema.Types.ObjectId,
			ref: "User",
			immutable: true,
			required: true,
		},
		teamId: {
			type: mongoose.Schema.Types.ObjectId,
			ref: "Team",
			immutable: true,
			required: true,
		},
		name: {
			type: String,
			required: true,
		},
		description: {
			type: String,
		},
		status: {
			type: Boolean,
			default: undefined,
		},
		type: {
			type: String,
			required: true,
			enum: ["http", "ping", "pagespeed", "hardware", "docker", "port"],
		},
		ignoreTlsErrors: {
			type: Boolean,
			default: false,
		},
		jsonPath: {
			type: String,
		},
		expectedValue: {
			type: String,
		},
		matchMethod: {
			type: String,
			enum: ["equal", "include", "regex", ""],
		},
		url: {
			type: String,
			required: true,
		},
		port: {
			type: Number,
		},
		isActive: {
			type: Boolean,
			default: true,
		},
		interval: {
			// in milliseconds
			type: Number,
			default: 60000,
		},
		uptimePercentage: {
			type: Number,
			default: undefined,
		},
		notifications: [
			{
				type: mongoose.Schema.Types.ObjectId,
				ref: "Notification",
			},
		],
		secret: {
			type: String,
		},
		thresholds: {
			type: {
				usage_cpu: { type: Number },
				usage_memory: { type: Number },
				usage_disk: { type: Number },
				usage_temperature: { type: Number },
			},
			_id: false,
		},
		alertThreshold: {
			type: Number,
			default: 5,
		},
		cpuAlertThreshold: {
			type: Number,
			default: function () {
				return this.alertThreshold;
			},
		},
		memoryAlertThreshold: {
			type: Number,
			default: function () {
				return this.alertThreshold;
			},
		},
		diskAlertThreshold: {
			type: Number,
			default: function () {
				return this.alertThreshold;
			},
		},
		tempAlertThreshold: {
			type: Number,
			default: function () {
				return this.alertThreshold;
			},
		},
	},
	{
		timestamps: true,
	}
);
```

client/src/Pages/Infrastructure/Monitors/Components/MonitorsTableMenu/index.jsx
```diff
@@ -28,7 +30,21 @@ const InfrastructureMenu = ({ monitor, isAdmin, updateCallback }) => {
 	const anchor = useRef(null);
 	const [isOpen, setIsOpen] = useState(false);
 	const [isDialogOpen, setIsDialogOpen] = useState(false);
+	const [localMonitorState, setLocalMonitorState] = useState(monitor);
```
ajhollid: There's no need to declare a local state to track the stauts of the monitor.  When the monitor is paused a new network request is made and the monitor's status is returned correctly:

client/src/Pages/Infrastructure/Monitors/Components/MonitorsTableMenu/index.jsx
```diff
@@ -54,14 +70,40 @@ const InfrastructureMenu = ({ monitor, isAdmin, updateCallback }) => {
...
+	const openConfigure = (id) => {
+		navigate(`/infrastructure/configure/${id}`);
+	};
+
+	const handlePause = async () => {
+		try {
+			// Toggle the local state immediately for better UI responsiveness
+			setLocalMonitorState((prevState) => ({
+				...prevState,
+				status: prevState.status === "paused" ? "active" : "paused",
```
ajhollid: `status` is not the property that is used to determine pause state.  `isActive` is the property that you want ot reference.

`status` is the property that determines if a monitor is up, down, or in the pending state

`true` === up
`false` === down
`undefined` === pending

`isActive` is either

`true` : not paused
`false`:  paused

ajhollid (changes requested):
This looks good visually, but there are some issues with the implementation that need to be resolved

Most important is that `Monitor`'s `status` property is the status of the montior.  

`isActive` is the property that determines whether or not the `Monitor` is paused.

Aside from that there is no need to track the state of the monitor in another state variable, it is already tracked and updated when the pause hook is called.

Please see my comments in the PR review for details on these issues, thanks!

client/src/Pages/Infrastructure/Monitors/Components/MonitorsTableMenu/index.jsx
```diff
@@ -54,14 +58,30 @@ const InfrastructureMenu = ({ monitor, isAdmin, updateCallback }) => {
...
+	};
+
+	const handlePause = async () => {
+		try {
+			// Pass updateCallback as triggerUpdate to the hook
+			await pauseMonitor({ monitorId: monitor.id, triggerUpdate: updateCallback });
+			// Toast is already displayed in the hook, no need to display it again
+		} catch (error) {
+			// Error handling is done in the hook
+		}
```
ajhollid: This try/catch block is not particularly useful since there are very few cases when `pauseMonitor` would throw an error that is not caught by the function itself, and nothing is done with an error if it occurs anyway.

Either we can drop the `try/catch` entirely or we should do something with the unlikely error that may occur.

client/src/Pages/Infrastructure/Monitors/Components/MonitorsTableMenu/index.jsx
```diff
@@ -54,14 +58,30 @@ const InfrastructureMenu = ({ monitor, isAdmin, updateCallback }) => {
...
+	};
+
 	const handleRemove = async () => {
 		try {
 			await networkService.deleteMonitorById({
 				monitorId: monitor.id,
 			});
-			createToast({ body: "Monitor deleted successfully." });
+			createToast({
+				body: t("monitorActions.deleteSuccess", "Monitor deleted successfully"),
```
ajhollid: We don't need the second parameter default string, it can be removed from all `t()`calls

ajhollid (changes requested):
Much improved over the last PR, looking pretty good!  Thanks for doing revisions :+1: 

There is a try/catch block that swallows an error silently, we should either remove that try/catch entirely or do something with the error.

We can also drop the second paramter from the `t()` calls, we don't need to provide a default string.

`t(delete, "delete")` can just be `t(delete)`

Nice work!

PR #2588, opened 2025-07-08
[Frontend]: Refactor PageSpeed monitor Create & Configure components

### LINKED ISSUE #1304: FE - Monitor Config Pages
Monitor Config pages and Create Monitor pages are nearly identical.  It would probably be best to refactor to reuse the "Create" page as "Configure" page as well.

This applies to Uptime, PageSpeed, and Infrastructure monitors

Jesulayomy:
## FE PageSpeed Monitor Create and Config Merger

- Refactored the Create and Configure components to use a single Setup component and reused most of the code
- This PR Refactors the PageSpeed Monitor

Fixes #1304

```Javascript
<div>{t('add')}</div>
```
- [x] I took a screenshot or a video and attached to this PR if there is a UI change. (No visible UI changes)


client/src/Routes/index.jsx
```diff
@@ -107,7 +106,7 @@ const Routes = () => {
 				/>
 				<Route
 					path="pagespeed/configure/:monitorId"
-					element={<PageSpeedConfigure />}
+					element={<PageSpeedCreate configure={true} />}
```
karenvicent: I noticed that PageSpeedSetup always receives configure={true} from the route. It feels like configure has become a static state — just wondering, if we ever need to switch it back to false, how would that be handled? Or is it meant to always stay true based on the route?

client/src/Pages/PageSpeed/Create/index.jsx
```diff
@@ -1,36 +1,41 @@
...
+	useFetchMonitorById,
+	useDeleteMonitor,
+	useUpdateMonitor,
+	usePauseMonitor,
+} from "../../../Hooks/monitorHooks";
 
-const SELECT_VALUES = [
+// Constants
+const MS_PER_MINUTE = 60000;
+const FREQUENCIES = [
```
karenvicent: Just a quick note — should we use the translation function for the name values in the FREQUENCIES array?

client/src/Pages/PageSpeed/Create/index.jsx
```diff
@@ -40,60 +45,98 @@ const SELECT_VALUES = [
...
-		};
-
-		await createMonitor({ monitor: form, redirect: "/pagespeed" });
 	};
 
 	const handleChange = (event) => {
-		const { value, name } = event.target;
+		let { value, name } = event.target;
+
+		if (name === "interval") {
```
karenvicent, line 145: What do you think about adding a new key to FREQUENCIES with the value already in milliseconds? Or maybe even creating a custom hook to handle the conversion if we reuse it elsewhere? It feels like handleChange shouldn't be responsible for that logic.

client/src/Routes/index.jsx
Jesulayomy: @karenvicent I'm using it as a toggle to know when we are in configure mode. Setting `configure={false}` would mean we are creating. It's implied on the `/pagespeed/create` route, but will be undefined - still falsy so the conditionals still work as expected.

client/src/Pages/PageSpeed/Create/index.jsx
Jesulayomy: I left this as is based on `handleChange`, `onChange` and `onSubmit` from the old files. I think it's great to consider but it was much simpler to implement while in handleChange, since it only affects the value sent to the monitor.

client/src/Pages/PageSpeed/Create/index.css
```diff
@@ -0,0 +1,9 @@
+.configure-pagespeed button {
+	height: var(--env-var-height-2);
+}
+
+.configure-pagespeed .field,
+.configure-pagespeed .section-disabled,
+.configure-pagespeed .select-wrapper {
+	flex: 1;
+}
```
ajhollid: We are moving way from using local CSS in the app and using MUIs theme system and inline styles when necessary.  

Can we please remove these and set the `flex` property on the components directly?

We don't want developers having to look in more than one place to figure out what styles are being appleid.

client/src/Pages/PageSpeed/Create/index.jsx
```diff
@@ -40,60 +45,98 @@ const SELECT_VALUES = [
...
-	// State
-	const [monitor, setMonitor] = useState({
-		url: "",
-		name: "",
-		type: "pagespeed",
-		notifications: [],
-		interval: 3,
-	});
+const PageSpeedSetup = ({ configure }) => {
+	const isConfigure = configure;
```
ajhollid: Is a prop here necessary at all?  We are fetching the `monitorId` param anyways, so if it is undefined then we are "creating", if it is defined, we are configuring:

```
const { monitorId } = useParams();
const isCreate = typeof monitorId === "undefined";
```

Other pages use the `isCreate` convention as opposed to `isConfigure`, so lets follow that convention to avoid confusion.

client/src/Pages/PageSpeed/Create/index.jsx
```diff
@@ -40,60 +45,98 @@ const SELECT_VALUES = [
...
 		event.preventDefault();
-		let form = {
-			url: `http${https ? "s" : ""}://` + monitor.url,
-			name: monitor.name === "" ? monitor.url : monitor.name,
-			type: monitor.type,
-			interval: monitor.interval * MS_PER_MINUTE,
-		};
+		if (isConfigure) {
+			await updateMonitor({ monitor, redirect: "/pagespeed" });
+		} else {
```
ajhollid: Why is validation bypassed when configuring?  Validation for configuration should be the same as for creating :thinking:

client/src/Pages/PageSpeed/Create/index.jsx
```diff
@@ -40,60 +45,98 @@ const SELECT_VALUES = [
...
-			abortEarly: false,
-		});
+			const { error } = monitorValidation.validate(form, { abortEarly: false });
+			if (error) {
+				const newErrors = {};
+				error.details.forEach((err) => {
+					newErrors[err.path[0]] = err.message;
+				});
+				setErrors(newErrors);
+				createToast({ body: "Please check the form for errors." });
```
ajhollid: This string needs to be extacted to the localization file `en.json`.  

Please be sure to also sort the file alphabetically before submitting :+1:

client/src/Pages/PageSpeed/Create/index.jsx
ajhollid: Good point @karenvicent , I think we can address this in a separate PR since it's an issue that would touch other pages as well.

client/src/Routes/index.jsx
ajhollid: @Jesulayomy we should be able to drop this prop entirely as we can determine whether or not the page is in configuration mode by the presence or absence of the `monitorId` route param

client/src/Pages/PageSpeed/Create/index.jsx
Jesulayomy: @ajhollid  Technically every field change is validated within `handleChange` so I'm not even sure it needs a revalidation `onSubmit` when creating

Within `handleChange()`
```js
let { value, name } = event.target;

...

const { error } = monitorValidation.validate(
  { [name]: value },
  { abortEarly: false }
);

setErrors(...);
```

Errors are displayed and since values are checked when changed, submitting should be safe.

client/src/Pages/PageSpeed/Create/index.jsx
Jesulayomy: I could take it out and cleanup the routes using something like this is more direct:
```js
const { monitorId } = useParams();
const isCreate = !!monitorId;
```

client/src/Pages/PageSpeed/Create/index.jsx
Jesulayomy: Do I also need to modify every other localization file to add the field there?

Jesulayomy: I extracted it to the `en.json` file, also did the same for the frequency times so other pages could use it.

```json
{
  "time": {
    "threeMinutes": "3 minutes",
    "fiveMinutes": "5 minutes",
    "tenMinutes": "10 minutes",
    "twentyMinutes": "20 minutes",
    "oneHour": "1 hour",
    "oneDay": "1 day",
    "oneWeek": "1 week"
  }
}
```

client/src/Routes/index.jsx
Jesulayomy: I've dropped the prop in the most recent commit. It now depends on the monitorId

Jesulayomy:
Made some changes to address your comments @karenvicent @ajhollid 

### On the most recent commit I have:
- [x] Removed the lines in the css file
- [x] Dropped the configure prop, now creating or configuring is dependent on `monitorId`
- [x] Added localization for time, checkFormError and urlMonitor to the `en.json` file

#### Not Changed
- [x] Validation bypass - left as is for now, validation still occurs on `handleChange`
- [x] Interval hooks - Separate issue
- [x] Synchronizing localization files - sounds like another good first issue

client/src/Pages/PageSpeed/Create/index.jsx
ajhollid: > Do I also need to modify every other localization file to add the field there?

No this is not necessary, just the `en.json` for the baseline, localization folks will handle the other files.

client/src/Pages/PageSpeed/Create/index.jsx
ajhollid: Validation should be done when submitting as well, it is the final gate keeper before data is sent off.

There's no guarantee that the behavior of `onChange` will stay as it is, but `onSubmit` will always be the step before submission so we want validation there always.

client/src/Pages/PageSpeed/Create/index.jsx
ajhollid: `isCreate = !!monitorId` is not correct here, that would mean that `isCreate` is true when a monitor ID is supplied.  When a monitor ID is supplied `isCreate` should be `false`.

Let's stick with `isCreate = tyepof monitorId === "undefined"`

This is explicit and the convention used elsewhere.

client/src/Pages/PageSpeed/Create/index.jsx
```diff
@@ -1,99 +1,141 @@
...
-	const [monitor, setMonitor] = useState({
-		url: "",
-		name: "",
-		type: "pagespeed",
-		notifications: [],
-		interval: 3,
-	});
+const PageSpeedSetup = () => {
+	const { monitorId } = useParams();
+	const isConfigure = !!monitorId;
```
ajhollid: Let's use the `isCreate` convention as in other pages please.

client/src/Pages/PageSpeed/Create/index.jsx
```diff
@@ -1,99 +1,141 @@
...
 	const [https, setHttps] = useState(true);
 	const [errors, setErrors] = useState({});
-	const { user } = useSelector((state) => state.auth);
-	const [notifications, notificationsAreLoading, error] = useGetNotificationsByTeamId();
+
+	const [isOpen, setIsOpen] = useState(false);
+	const [updateTrigger, setUpdateTrigger] = useState(false);
+	const [isDeleting, setIsDeleting] = useState(false);
+	const [isUpdating, setIsUpdating] = useState(false);
+	const [isPausing, setIsPausing] = useState(false);
```
ajhollid: There's no need to track loading state here, I believe all of these hooks supply an `isLoading` state.

ajhollid (changes requested):
Thanks for making requested changes here, it's looking good.

We would like to use `isCreate` as a convention rather than `isConfigure`.

I noticed that you're tracking loading state locally within the component, this is not necessary as the hooks provide a loading state, please use those states.

Thanks for your contribution!

client/src/Pages/PageSpeed/Create/index.jsx
Jesulayomy: Sorry I mixed that up with `isConfigure` in that example.

client/src/Pages/PageSpeed/Create/index.jsx
Jesulayomy: Yes, added `time.threeMinutes, ...` to the en.json

client/src/Pages/PageSpeed/Create/index.jsx
Jesulayomy: Added validation of the specific params to the update conditional
```js
const monitorParams = {
    url: monitor.url,
    name: monitor.name === "" ? monitor.url : monitor.name,
    type: monitor.type,
    interval: monitor.interval,
}
const { error } = monitorValidation.validate(monitorParams, { abortEarly: false });
```

Can't pass the entire monitor into the validate() function because it will have other parameters when in configure mode. (Data that comes with fetching the monitor)

client/src/Pages/PageSpeed/Create/index.jsx
Jesulayomy: Now using:

```js
const { monitorId } = useParams();
const isCreate = typeof monitorId === "undefined";
```

The other parts of the code will be adjusted accordingly to replace isConfigure variable

client/src/Pages/PageSpeed/Create/index.jsx
Jesulayomy: Removed, `updateTrigger` is used when fetching the monitor based on an id, so it's kept in.

Jesulayomy:
Made another [commit](https://github.com/bluewave-labs/Checkmate/pull/2588/commits/eaa85bb9c46aa61527a43ff2f6f45e1135102bb8) for the recent comments
@ajhollid

client/src/Pages/PageSpeed/Create/index.jsx
```diff
@@ -1,99 +1,154 @@
...
+		useGetNotificationsByTeamId();
+
+	// Hooks for API actions
 	const [createMonitor, isCreating] = useCreateMonitor();
+	const [isLoading] = isCreate
+		? [false]
+		: useFetchMonitorById({ monitorId, setMonitor, updateTrigger });
+	const [deleteMonitor] = useDeleteMonitor();
+	const [updateMonitor] = useUpdateMonitor();
+	const [pauseMonitor] = usePauseMonitor();
```
ajhollid: We need the loading states from these hooks.  Buttons that perform network actions should all be disabled when another action is taking place.

For example you should not be able to update a monitor while it is in the process of being paused

client/src/Pages/PageSpeed/Create/index.jsx
```diff
@@ -1,99 +1,154 @@
...
+	const { user } = useSelector((state) => state.auth);
+	const { statusColor, pagespeedStatusMsg, determineState } = useMonitorUtils();
+	const [notifications, notificationsAreLoading, notificationsError] =
+		useGetNotificationsByTeamId();
+
+	// Hooks for API actions
 	const [createMonitor, isCreating] = useCreateMonitor();
+	const [isLoading] = isCreate
+		? [false]
+		: useFetchMonitorById({ monitorId, setMonitor, updateTrigger });
```
ajhollid: This will a violation of the fundamental [rules of hooks](https://react.dev/warnings/invalid-hook-call-warning), let's avoid conditionally calling hooks. 

If you really need to call this conditionally then you should add a condition to the hook itself to return early if the condition is/is not met.

client/src/Pages/PageSpeed/Create/index.jsx
```diff
@@ -103,210 +158,359 @@ const CreatePageSpeed = () => {
...
-						component="span"
-						fontSize="inherit"
+			{Object.keys(monitor).length === 0 ? (
+				<SkeletonLayout />
+			) : (
+				<>
+					<Breadcrumbs list={CRUMBS} />
+					<Stack
+						component="form"
+						className={!isCreate ? "" : "create-monitor-form"}
```
ajhollid: We should not have classnames any longer

ajhollid (changes requested):
It's getting there :+1:  A coupel of issues I see at the moment

1. Unused classnames, these should be removed
2. Conditional calling of a hook, this should be avoided and conditional early returns added to the hook itself
3.  Loading states need to be used to set the `loading` prop of buttons so they show loading state and are disabled when an operation is in place.  If you are doing a netwokr operation all buttons that allow network operations to take place should be set to `loading`

Thanks for revisions!

client/src/Pages/PageSpeed/Create/index.jsx
Jesulayomy: I'll place a check for undefined monitorId in the hook itself so it doesn't try to fetch the monitor when it does not exist.
```js
if (monitorId !== undefined) {
// Fetch monitor
...
} else {
  // set loading to false before return
  setIsLoading(false);
}
```

client/src/Pages/PageSpeed/Create/index.jsx
Jesulayomy: Fetched the loading states from the hooks, and placed them on Pause/Resume, Remove, & CreateMonitor/Save buttons.

client/src/Pages/PageSpeed/Create/index.jsx
Jesulayomy: [Eliminated> ( -_•)▄︻デ══━一

Jesulayomy:
### Summary of latest commits

- [x] Removed unused classnames in the component
- [x] Moved conditional calling of the hook to inside said hook
- [x] Retrieved states from hooks and placed loading states on buttons that would perform network operations (`pausing`, `removing`, `saving`, `creating`)
- [x] Accidently left a console.log statement

client/src/Pages/PageSpeed/Create/index.jsx
```diff
@@ -103,210 +155,361 @@ const CreatePageSpeed = () => {
...
-				mt={theme.spacing(6)}
-			>
-				<Typography
-					component="h1"
-					variant="h1"
-				>
-					<Typography
-						component="span"
-						fontSize="inherit"
+			{Object.keys(monitor).length === 0 ? (
```
Br0wnHammer: Ensure that the logic for rendering the skeleton layout is handled prior to the actual rendering of the component.

client/src/Pages/PageSpeed/Create/index.jsx
```diff
@@ -103,210 +155,361 @@ const CreatePageSpeed = () => {
...
+									alignSelf="flex-end"
+									ml="auto"
 								>
-									{t("http")}
-								</Button>
-							</ButtonGroup>
+									<Button
+										onClick={handlePause}
+										loading={
+											isLoading || isCreating || isDeleting || isUpdating || isPausing
```
Br0wnHammer: I believe this can also be omitted and managed before the component rendering.

client/src/Pages/PageSpeed/Create/index.jsx
```diff
@@ -103,210 +155,361 @@ const CreatePageSpeed = () => {
...
+										<Typography
+											position="relative"
+											variant="body2"
+											ml={theme.spacing(6)}
+											mt={theme.spacing(1)}
+											sx={{
+												"&:before": {
+													position: "absolute",
+													content: `""`,
+													width: 4,
```
Br0wnHammer: These are hardcoded values. I am assuming they are from legacy. Try to use `theme.spacing` here aswell.

client/src/Pages/PageSpeed/Create/index.jsx
```diff
@@ -103,210 +155,361 @@ const CreatePageSpeed = () => {
...
-							label="Check frequency"
-							value={monitor.interval || 3}
-							onChange={handleChange}
-							items={SELECT_VALUES}
-						/>
+							<Button
+								type="submit"
+								variant="contained"
+								color="accent"
+								loading={isLoading || isCreating || isDeleting || isUpdating || isPausing}
```
Br0wnHammer: Same here. The logic can be reused.

Br0wnHammer (changes requested):
Looks good for the most part. Please review my comments. Some modifications are required.

client/src/Pages/PageSpeed/Create/index.jsx
Jesulayomy: Got it, the condition is checked just before the main return statement

client/src/Pages/PageSpeed/Create/index.jsx
Jesulayomy: I can combine it to an isBusy boolean so it's easier to read, it affects when a specific button (`pause`, `remove`, `save`) is clicked so the loading state will still belong to the button. While the network is performing the request, it has a spinner + disable effect on all buttons to prevent another request while loading.

```js
const isBusy = isLoading || isCreating || isDeleting || isUpdating || isPausing;

return ...
```

Screenshot 2025-07-11 at 11 03 07

client/src/Pages/PageSpeed/Create/index.jsx
Jesulayomy: I see the styling is for the circle that appears next to the Editing text. Do you think replacing this with a • would be reasonable?

Jesulayomy: I'm replacing the sizes with the theme spacings

Jesulayomy:
### Latest changes ♻️ 

- [x] Moved loading checks into a single `isBusy` variable
- [x] Replaced hard coded spacing with theme spacing
- [x] Moved skeleton layout render outside of main render

client/src/Hooks/monitorHooks.js
```diff
@@ -205,18 +205,22 @@ const useFetchStatsByMonitorId = ({
...
-				const res = await networkService.getMonitorById({ monitorId: monitorId });
-				setMonitor(res.data.data);
-			} catch (error) {
-				createToast({ body: error.message });
-			} finally {
-				setIsLoading(false);
-			}
-		};
-		fetchMonitor();
+		if (monitorId !== undefined) {
```
ajhollid: I find the early return pattern to be much more readable:

```
if (typeof monitorId === "undefined") {
    setIsLoading(false);
    return;
}
```

This lets us dispense with the else block entirely.

We should also use `typeof monitorId === undefined` vs `monitorId === undefined` as a rule as it handles the case of an undeclared variable as well which strict equality does not.

Jesulayomy: Done ✅

Br0wnHammer (approved):
LGTM! Thanks for your contributions 🚀

Br0wnHammer:
@Jesulayomy There are some format checks failing. Can you take care of that. Make sure to run `npm run format`.

Jesulayomy:
@Br0wnHammer Thanks for noticing that, I've fixed it now. 🚀

PR #2595, opened 2025-07-11
feat: diagnostics tab

ajhollid:
This PR adds a diagnostic tab to the Logs page.  This page will allow users to assess the health of their Checkmate instance.

- [x] Add endpoint for gathering system stats
- [x] Add a diagnostic page
- [x] Add gauges for relevant data
- [x] Add statistic boxes for non graphic data
- [x] Unify log page design

image
    

## Summary by CodeRabbit

* **New Features**
  * Introduced a Diagnostics tab in the Logs page, providing real-time system and process health metrics with visual gauges and statistics.
  * Added detailed diagnostic metrics including CPU usage, heap memory allocation and utilization, event loop delay, uptime, and OS memory limit.
  * Enhanced gauge and stats components for customizable precision, units, and improved loading indicators.

* **Improvements**
  * Updated UI layouts for Logs and Queue pages with clearer headings, dividers, and improved spacing.
  * Refined localization with new strings for diagnostics and updated logs descriptions.

* **Bug Fixes**
  * Improved duration formatting to include milliseconds for greater accuracy in short durations.

### DIFF
--- a/client/src/Pages/Logs/Diagnostics/index.jsx
+++ b/client/src/Pages/Logs/Diagnostics/index.jsx
@@ -2,6 +2,8 @@ import Stack from "@mui/material/Stack";
 import Box from "@mui/material/Box";
 import Typography from "@mui/material/Typography";
 import Gauges from "./components/gauges";
+import Stats from "./components/stats";
+import Divider from "@mui/material/Divider";
 
 import { useTheme } from "@emotion/react";
 import { useState } from "react";
@@ -15,14 +17,26 @@ const Diagnostics = () => {
 	const theme = useTheme();
 	const { t } = useTranslation();
 	const [diagnostics, isLoading, error] = useFetchDiagnostics();
-	console.log(diagnostics);
 	// Setup
 	return (
 		<Stack gap={theme.spacing(4)}>
 			<Box>
-				<Typography variant="h2">{t("diagnosticsPage.description")}</Typography>
+				<Typography variant="h2">{t("diagnosticsPage.diagnosticDescription")}</Typography>
 			</Box>
-			<Gauges diagnostics={diagnostics} />
+			<Divider color={theme.palette.accent.main} />
+			<Stack
+				gap={theme.spacing(20)}
+				mt={theme.spacing(10)}
+			>
+				<Gauges
+					diagnostics={diagnostics}
+					isLoading={isLoading}
+				/>
+				<Stats
+					diagnostics={diagnostics}
+					isLoading={isLoading}
+				/>
+			</Stack>
 		</Stack>
 	);
 };
--- a/client/src/Components/Charts/CustomGauge/index.jsx
+++ b/client/src/Components/Charts/CustomGauge/index.jsx
@@ -1,8 +1,10 @@
 import { useTheme } from "@emotion/react";
 import { useEffect, useState, useMemo } from "react";
+import Stack from "@mui/material/Stack";
 import { Box, Typography } from "@mui/material";
 import PropTypes from "prop-types";
 import "./index.css";
+import CircularProgress from "@mui/material/CircularProgress";
 
 const MINIMUM_VALUE = 0;
 const MAXIMUM_VALUE = 100;
@@ -16,6 +18,8 @@ const MAXIMUM_VALUE = 100;
  * @param {number} [props.radius=60] - Radius of the gauge circle
  * @param {number} [props.strokeWidth=15] - Width of the gauge stroke
  * @param {number} [props.threshold=50] - Threshold for color change
+ * @param {number} [props.precision=1] - Precision of the progress percentage
+ * @param {string} [props.unit="%"] - Unit of progress
  *
  * @example
  * <CustomGauge
@@ -27,7 +31,15 @@ const MAXIMUM_VALUE = 100;
  *
  * @returns {React.ReactElement} Rendered CustomGauge component
  */
-const CustomGauge = ({ progress = 0, radius = 70, strokeWidth = 15, threshold = 50 }) => {
+const CustomGauge = ({
+	isLoading = false,
+	progress = 0,
+	radius = 70,
+	strokeWidth = 15,
+	threshold = 50,
+	precision = 1,
+	unit = "%",
+}) => {
 	const theme = useTheme();
 	// Calculate the length of the stroke for the circle
 	const { circumference, totalSize, strokeLength } = useMemo(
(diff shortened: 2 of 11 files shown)


client/src/Pages/Logs/Diagnostics/index.jsx
```diff
@@ -0,0 +1,44 @@
...
+import { useTranslation } from "react-i18next";
+import { useFetchDiagnostics } from "../../../Hooks/logHooks";
+
+const Diagnostics = () => {
+	// Local state
+
+	// Hooks
+	const theme = useTheme();
+	const { t } = useTranslation();
+	const [diagnostics, isLoading, error] = useFetchDiagnostics();
```
Owaiseimdad, line 19: @ajhollid , any possibility of getting errors here. If yes can we add that in the toast here.

client/src/Hooks/logHooks.js
```diff
@@ -32,4 +32,79 @@ const useFetchLogs = () => {
...
 
-export { useFetchLogs };
+const useFetchQueueData = (trigger) => {
+	const [jobs, setJobs] = useState(undefined);
+	const [metrics, setMetrics] = useState(undefined);
+	const [isLoading, setIsLoading] = useState(false);
+	const [error, setError] = useState(undefined);
+
+	useEffect(() => {
+		const fetchJobs = async () => {
```
Owaiseimdad, line 42: @ajhollid there is no way to fetch the hooks again rather than manually refresh or hop on to another page and come back. Can we have a capability to add a refetch functionality? 

Like (just a demo):
`const useFetchDiagnostics = () => {
     // ...state
     const fetchDiagnostics = async () => { ... };
     useEffect(() => { fetchDiagnostics(); }, []);
     return [diagnostics, isLoading, error, fetchDiagnostics];
   };
`

server/controllers/diagnosticController.js
```diff
@@ -44,5 +52,85 @@ class DiagnosticController {
...
+			let eventLoopDelay = 0;
+			performance.mark("start");
+			await new Promise((resolve) => setTimeout(resolve, 0));
+			performance.mark("end");
+			performance.measure("eventLoopDelay", "start", "end");
+			const entries = performance.getEntriesByName("eventLoopDelay");
+			if (entries.length > 0) {
+				eventLoopDelay = entries[0].duration;
+			}
+
```
Owaiseimdad, line 113: ### Prevent Memory Leak in Diagnostics Endpoint
can we clear the memory of performance here instead, to avoid the memory leak.

client/src/Hooks/logHooks.js
```diff
@@ -32,4 +32,79 @@ const useFetchLogs = () => {
...
+			try {
+				setIsLoading(true);
+				const response = await networkService.getQueueData();
+				if (response.status === 200) {
+					setJobs(response.data.data.jobs);
+					setMetrics(response.data.data.metrics);
+				}
+			} catch (error) {
+				setError(error);
```
Owaiseimdad, line 51: I see we add toast here itself, so we can add it here then.

server/controllers/diagnosticController.js
```diff
@@ -44,5 +52,85 @@ class DiagnosticController {
 			next(handleError(error, SERVICE_NAME, "getDbStats"));
 		}
 	}
+
+	async getCPUUsage() {
+		try {
+			const startUsage = process.cpuUsage();
+			await new Promise((resolve) => setTimeout(resolve, 1000));
+			const endUsage = process.cpuUsage(startUsage);
+			const cpuUsage = {
```
Owaiseimdad, line 61: Can we add comments here like why are we doing /1000.

`
 // Convert microseconds to milliseconds
    userUsageMs: endUsage.user / 1000,
    systemUsageMs: endUsage.system / 1000,
    // Convert total CPU time (μs) to percentage of 1 second interval
    usagePercentage: ((endUsage.user + endUsage.system) / 1_000_000) * 100,
  `

client/src/Hooks/logHooks.js
```diff
@@ -32,4 +32,79 @@ const useFetchLogs = () => {
...
+	const [metrics, setMetrics] = useState(undefined);
+	const [isLoading, setIsLoading] = useState(false);
+	const [error, setError] = useState(undefined);
+
+	useEffect(() => {
+		const fetchJobs = async () => {
+			try {
+				setIsLoading(true);
+				const response = await networkService.getQueueData();
+				if (response.status === 200) {
```
Owaiseimdad, line 46: I was exploring something called Unmount Safety, recently and i think we can use it here. 

Like:
`
useEffect(() => {
  let isMounted = true;
  const fetchJobs = async () => {
    try {
      setIsLoading(true);
      const response = await networkService.getQueueData();
      if (isMounted) {
        setJobs(response.data.data.jobs);
        setMetrics(response.data.data.metrics);
      }
    } catch (error) {
      if (isMounted) setError(error);
    } finally {
      if (isMounted) setIsLoading(false);
    }
  };
  fetchJobs();
  return () => { isMounted = false; };
}, [trigger]);
`

Let me if I am wrong here. We can use it for other hooks as well.

karenvicent (commented):
Hi @ajhollid  everything works well locally and overall the feature looks great!
Just a small UX detail for the future: we could slightly improve the statistic boxes for non graphic data. Maybe by capitalizing and bolding the title, centering the text, adding some border-radius, and giving them a slightly different background to highlight the data as those with the graphics.

ajhollid:
> Hi @ajhollid everything works well locally and overall the feature looks great! Just a small UX detail for the future: we could slightly improve the statistic boxes for non graphic data. Maybe by capitalizing and bolding the title, centering the text, adding some border-radius, and giving them a slightly different background to highlight the data as those with the graphics.

Thanks for the suggestion!  How about we handles styling in another PR?  This is not my specialty and perhaps one of the FE team could take this on?  I'm affraid I'll just make it worse than it is :joy:

client/src/Hooks/logHooks.js
ajhollid: @Owaiseimdad What problem is this meant to solve?  Do you have some reference material I can take a look at?  I'm not a React expert so I don't really feel qualified to say whether this is useful or not :joy:

server/controllers/diagnosticController.js
ajhollid: This is done in the observer 

```
const obs = new PerformanceObserver((items) => {
	const entry = items.getEntries()[0];
	performance.clearMarks();
});
```

`clearMarks()` should take care of this

client/src/Hooks/logHooks.js
ajhollid: Can do

client/src/Hooks/logHooks.js
Owaiseimdad: Even I am not an expert @ajhollid 😅. Actually recently I was going through a article ( I think in medium) about mounting (basically about react architecture). If we are using fetch(async) and waiting for data then by that time if user navigates to a diff page then even though we are not in that page, we still set the states in react. In order to avoid that we can wait if the mounting can happen and then set the states. Its not required, but it does improve how we change the states. 

If you feel its not required. No need to do it.

ajhollid: That sounds a lot like oldschool react with `componentWillMount` etc :joy:   I don't know if this is necessary or not, but if we run into isseus because of this we can look further into it. 

This would be applicable to all places we are fetching data, so most certainly worthy of its own separate issue and PRs

PR #2601, opened 2025-07-12
Implement Exponential Backoff for Incidents (#2366)

### LINKED ISSUE #2366: Implement exponential backoff for incidents
Currently when a service is down, it waits for 5 incidents to happen. We can implement a exponential backoff strategy where the time window increases until the service is back up again.

mohicody:
### Summary
This PR implements an exponential backoff strategy for incident notifications to reduce alert fatigue during prolonged service outages. When a service remains down, the time between notifications increases exponentially until the service recovers.

### Changes
- Created configuration file with customizable backoff parameters
- Added backoff fields to the Notification model
- Implemented backoff logic with jitter in notification handling
- Added recovery detection to reset backoff on service restoration
- Added proper error handling and logging
### Technical Details
- Initial delay: 60 seconds (configurable)
- Max delay: 1 hour (configurable)
- Backoff multiplier: 2x (configurable)
- Jitter: 20% to prevent thundering herd problem
- Automatic reset on service recovery

Fixes #2366 

- [ ] (Do not skip this or your PR will be closed) I deployed the application locally.
- [ ] (Do not skip this or your PR will be closed) I have performed a self-review and testing of my code.
```Javascript
<div>{t('add')}</div>
```

## Summary by CodeRabbit

* **New Features**
  * Introduced configurable exponential backoff settings for notifications, including options for delay, multiplier, maximum delay, and jitter.
  * Added monitor-level controls and tracking for backoff behavior in notifications.
  * Implemented logic to automatically reset notification backoff when a monitor recovers.

* **Improvements**
  * Enhanced error logging during notification dispatch for better traceability.
  * Centralized and streamlined notification backoff configuration.


Br0wnHammer (approved):
LGTM!

ajhollid (changes requested):
The concept here is on the right track, however this would apply a backoff to a notification channel.  That means that any monitor that uses that notification channel would update the backoff.

Rather, the backoff should be applied to individual monitors for the desired behaviour here.

mohicody:
> The concept here is on the right track, however this would apply a backoff to a notification channel. That means that any monitor that uses that notification channel would update the backoff.
> 
> Rather, the backoff should be applied to individual monitors for the desired behaviour here.

Thanks for the feedback! I see the issue now. I'll refactor to move the backoff fields and logic to the Monitor model instead, so each monitor has its own independent backoff state.

server/utils/notificationConfig.js
```diff
@@ -0,0 +1,20 @@
...
+
+	// Maximum delay between notifications (1 hour in ms)
+	MAX_BACKOFF_DELAY_MS: 3600000,
+
+	// Multiplier for exponential growth of delay
+	BACKOFF_MULTIPLIER: 2,
+
+	// Jitter factor to prevent thundering herd (20% randomness)
+	JITTER_FACTOR: 0.2,
+};
```
ajhollid: Are these meant to be global settings for all notifications?

If so, these should be part of the `AppSettings` model and should be handled by the settings service

server/utils/notificationConfig.js
```diff
@@ -0,0 +1,20 @@
+/**
+ * Configuration for notification behavior including exponential backoff
+ * These values determine how notifications are delayed when a service is down
+ */
+export default {
+	// Whether backoff is enabled by default for new notifications
+	BACKOFF_ENABLED_DEFAULT: true,
```
ajhollid: Let's have this as disabled by default since we don't have any UI for setting any of these values right now

server/db/models/Monitor.js
```diff
@@ -75,6 +76,31 @@ const MonitorSchema = mongoose.Schema(
 				ref: "Notification",
 			},
 		],
+		// Exponential backoff configuration
+		backoffEnabled: {
+			type: Boolean,
+			default: notificationConfig.BACKOFF_ENABLED_DEFAULT,
```
ajhollid: default to `false` please until we have some UI for handling backoff settings

ajhollid:
Closing this as a simpler approach is possible

PR #2690, opened 2025-07-28
refactor: centralize hooks and add reusable `useFetchData` hook (#1867)

### LINKED ISSUE #1867: FE - Hook organizatin
There are a number of hooks used to make network requests.  Some of them are reused between components, some have been implemented in more than one place and are nearly identical.  This was done in a hurry to quickly refactor some pages, but they should now be organized.

- [ ] Hooks should be moved out of Page directories and to the root Hook directory.
- [ ] Duplicate hooks should be removed
- [ ] Hooks that are very similar should be evaluated for refactoring, if reasonable they should be made generic enought to use in more than one component/page

AbhishekG-Codes:
## Describe your changes

Refactored the folder structure to improve code organization and maintainability.  
changes were made to the logic or functionality of the application, moved repeated logic to single file called ./Hooks/useFetchData.js and was used in ./Hooks/checkHooks.js

Fixes #1867

- [x]  (Do not skip this or your PR will be closed) I deployed the application locally.
- yes. only folder structure and moving hook to centralized file for reuse.
-
- [x]  (Do not skip this or your PR will be closed) I have performed a self-review and testing of my code.
- yes

- yes
- [x] I have added i18n support to visible strings (N/A - no UI strings were touched).
- [x] I have **not** included any files that are not related to my pull request, including package-lock and package-json if dependencies have not changed.
- [x] I made sure font sizes, color choices etc are all referenced from the theme. (N/A)

- [x]  My PR is granular and targeted to one specific feature (folder structure refactor).
- yes

- [x]  I ran `npm run format` in server and client directories, which automatically formats your code.
- yes

- [x] I took a screenshot or a video and attached to this PR if there is a UI change. (N/A)
- (N/A)


ajhollid (commented):
I approve of the idea of writing a more generic reusable hook, but why go through the extra trouble fo them wrapping that generic hook in another hook?

```
const useFetchChecksTeam = (params) => {
	const config = { ...params };
	const deps = Object.values(config); // Safe enough for now

	const requestFn = () => networkService.getChecksByTeam(config);

	return useFetchData({
		requestFn,
		enabled: config.enabled,
		deps,
	});
};
```

This seems redundant to me, why not just call `useFetchData` directly?

Additionally, accepting arbitrary `params` without any JSdoc makes using these hooks more error prone in my opinion.  We should define what parameters we can accept and enforce as best we can so users of the hook don't have to work too hard.

If we're going to refactor the hooks, we may as well do it thoroughly and remove the wrapper hooks entirely unless there's a compelling reason not to.

If there is such a reason, please do let me know and we can discuss!

AbhishekG-Codes:
You're right that useFetchChecksTeam essentially wraps useFetchData, and that may seem redundant at first glance. But, I added this wrapper intentionally to maintain semantic clarity and abstraction at the call site. 
By having named hooks like useFetchChecksTeam, we make the purpose of the data fetching explicit and isolate the API specifics inside the hook.

This also gives us flexibility going forward, for example, if the data transformation, error handling, or side-effects specific to this fetch change in the future, we can manage that within the wrapper without touching consuming components.

That said, I do agree it’s important not to over-abstract it. If the main idea is to call useFetchData directly and remove these wrappers for simplicity, I’m happy to update accordingly.

Also noted on the JSDoc comment, I’ll add proper documentation to clarify expected params and improve dev experience. Thanks again for pointing that out.....

Br0wnHammer:
@AbhishekG-Codes  I believe Alex's suggestion is valid. We can directly call the wrapper hooks with the dependency array where the function needs to be executed, which simplifies the process. Additionally, sending the dependencies as an array appears to be a good approach. However, we should include a dependencies checker function that can serve as an extra validation layer to ensure that the hooks do not re-render infinitely due to a specific property.

Also, please make sure to run `npm run format` on the client dir.

Br0wnHammer:
@AbhishekG-Codes any update?

AbhishekG-Codes:
> @AbhishekG-Codes any update?

Thanks for the follow-up @Br0wnHammer!

I’m currently wrapping up the suggested changes:
- Working on adding a dependency checker utility to prevent infinite renders.
- Will run `npm run format` on the `client` dir.
- Will update or refactor the wrapper hooks (`useFetchChecksTeam` etc.) based on team decision, leaning towards keeping them with better documentation for now.

Will push the changes and update here by EOD. Appreciate the patience!

AbhishekG-Codes:
Refactored hooks to accept dependency arrays as suggested. Added a dependency checker for validation. Also ran npm run format in the client directory as requested. Ready for review.

client/src/Hooks/useFetchData.js
```diff
@@ -0,0 +1,115 @@
...
+import { createToast } from "../Utils/toastUtils";
+
+/**
+ * Checks dependencies for potential issues in development mode.
+ * Warns if dependencies are non-memoized functions or objects, which may cause unnecessary re-renders.
+ *
+ * @param {Array<any>} deps - Dependency array passed to useEffect
+ * @param {Array<string>} depsNames - Optional array of names corresponding to dependencies for better warnings
+ */
+const validateDeps = (deps, depsNames = []) => {
```
Br0wnHammer: What I meant by validating was making sure that no prop was sent that could re-render the component unnecessarily. I am not sure about the validation of the dependency array itself. You can remove this function as it seems too generic.

client/src/Hooks/useFetchData.js
```diff
@@ -0,0 +1,115 @@
...
+	depsNames = [],
+	extractFn = null,
+}) => {
+	const [data, setData] = useState(undefined);
+	const [count, setCount] = useState(undefined);
+	const [isLoading, setIsLoading] = useState(false);
+	const [networkError, setNetworkError] = useState(false);
+
+	// Dev-only dependency validation
+	useEffect(() => {
```
Br0wnHammer: This can be removed too in such case.

client/package-lock.json
```diff
@@ -7220,20 +7220,6 @@
 			"integrity": "sha512-a4UGQaWPH59mOXUYnAG2ewncQS4i4F43Tv3JoAM+s2VDAmS9NsK8GpDMLrCHPksFT7h3K6TOoUNn2pb7RoXx4g==",
 			"license": "ISC"
 		},
-		"node_modules/yaml": {
```
Br0wnHammer: Can you please omit the `package-lock.json` file from your PR?

AbhishekG-Codes:
> Looks good for the most part. Please see my comments on the changes that are needed.

Thanks for the feedback! @Br0wnHammer 

I've already addressed the following as part of the latest changes:
- Reverted the package-lock.json from develop using git checkout and amended the last commit to clean up history.
- Cleaned up useFetchData and relocated hooks to client/src/Hooks/ for better organization.
- Pushed the changes with --force-with-lease to keep the PR history clean.

ajhollid:
Hey @AbhishekG-Codes ,

After looking at this further I'm not convinced this is the direction we want to be taking this in.

We are actively trying to move business logic out of the UI components, and having to import and use the network service in UI components feels like a step backwards to me.

```
networkService.getChecksByMonitor({
	monitorId: selectedMonitor === "0" ? undefined : selectedMonitor,
	type: selectedMonitorType,
	status: false,
	sortOrder: "desc",
	limit: null,
	dateRange,
	filter: filter === "resolved" ? "all" : filter,
	ack: filter === "resolved" ? true : false,
	page,
	rowsPerPage,
}),
```

I don't want devs to have to do this in the UI component, they should simply be able to use the hook as they do now and get a value back.

I'm all for refactoring the hooks to make them more reusable and predictable, and certainly for moving them to the correct directory, but we should aim to decrease complexity in the UI components.

PR #2707, opened 2025-07-29
Feature/global thresholds

### LINKED ISSUE #2628: Global Threshold
Not specifically, but it does ease the management of the servers.

Setting a global threshold for CPU, memory, disk and temperature that is automatically applied when a server is added.

A default setting so it doesn't have to be enabled individually on each server.

**Additional context**
Discord thread here:
https://discordapp.com/channels/1316935690377760788/1394592059188445194

singh-kanwarpreet:
## Describe your changes

I implemented a Global Thresholds section in the Settings page. Users can define CPU, Memory, Disk, and Temperature threshold values and save them. These values are stored and automatically pre-filled when the user navigates to the Infrastructure page, ensuring consistent threshold usage across the app.

Fixes #2628 

Video link - https://drive.google.com/file/d/1_I9UfGoO76JwyGgmO4MIQIlVZsaV2BKG/view?usp=sharing

## Summary by CodeRabbit

* **New Features**
  * Added a new settings section for configuring global CPU, Memory, Disk, and Temperature thresholds.
  * Introduced a user interface for admins to set and update global threshold values.
  * Global threshold values are now applied when creating new infrastructure monitors.
  * Added a hook to fetch global application settings for use in the UI.

* **Improvements**
  * Enhanced validation to support global threshold settings.
  * Added English localization for the new global thresholds section in the settings page.

* **Bug Fixes**
  * Ensured proper initialization of form fields when creating new monitors based on global thresholds.


client/src/Pages/Infrastructure/Create/index.jsx
```diff
@@ -87,35 +91,46 @@ const CreateInfrastructureMonitor = () => {
 	});
 
 	// Populate form fields if editing
+
 	useEffect(() => {
-		if (isCreate || !monitor) return;
```
Br0wnHammer, line 91: Can you explain the logic behind removing monitor from the if statement also why is !isCreate replacing isCreate?

client/src/Pages/Infrastructure/Create/index.jsx
```diff
@@ -87,35 +91,46 @@ const CreateInfrastructureMonitor = () => {
...
+		if (!isCreate || globalSettingsLoading) return;
+		const globalThresholds = globalSettings?.data?.settings?.globalThresholds || {};
 
-		setInfrastructureMonitor({
-			url: monitor.url.replace(/^https?:\/\//, ""),
```
Br0wnHammer, line 94: url, name, notifications and interval have been omitted from here. Any reason?

client/src/Pages/Infrastructure/Create/index.jsx
```diff
@@ -87,35 +91,50 @@ const CreateInfrastructureMonitor = () => {
 	});
 
 	// Populate form fields if editing
+
 	useEffect(() => {
-		if (isCreate || !monitor) return;
+		if (monitor == undefined && (!isCreate || globalSettingsLoading)) return;
```
Br0wnHammer: The logic seems to be off still. This means that if and only if the monitor is undefined it will check further. This will result in a lot of errors. 

Please stick to the previous logic and add in the globalSettingsLoading

client/src/Pages/Infrastructure/Create/index.jsx
```diff
@@ -87,35 +91,50 @@ const CreateInfrastructureMonitor = () => {
...
+			temperature: "",
+		};
 
-			memory: monitor.thresholds?.usage_memory !== undefined,
-			usage_memory: monitor.thresholds?.usage_memory
-				? monitor.thresholds.usage_memory * 100
-				: "",
+		setInfrastructureMonitor((prev) => ({
+			...prev,
+			url: monitor != undefined ? monitor.url.replace(/^https?:\/\//, "") : "",
```
Br0wnHammer: Also no need to explicitly check if the monitor is undefined each time. If it is already handled above. Omit the logic part here in all properties.

client/src/Pages/Infrastructure/Create/index.jsx
singh-kanwarpreet: sir i will create 2 diff part one for create another for update

client/src/Pages/Infrastructure/Create/index.jsx
singh-kanwarpreet: sir it will be solved on creating 2 diff if else

client/src/Pages/Infrastructure/Create/index.jsx
ajhollid: @singh-kanwarpreet please avoid resolving conversations as all reviewers need to read them to see what's been happening.

What was the consensus on this issue?

client/src/Pages/Infrastructure/Create/index.jsx
ajhollid: resolution here?

client/src/Pages/Infrastructure/Create/index.jsx
```diff
@@ -87,35 +91,55 @@ const CreateInfrastructureMonitor = () => {
...
+				name: monitor.name || "",
+				notifications: monitor.notifications || [],
+				interval: monitor.interval / MS_PER_MINUTE,
+				cpu: !!monitor.thresholds?.usage_cpu ?? false,
+				usage_cpu: monitor.thresholds?.usage_cpu * 100?.toString() || "",
+				memory: !!monitor.thresholds?.usage_memory ?? false,
+				usage_memory: monitor.thresholds?.usage_memory * 100?.toString() || "",
+				disk: !!monitor.thresholds?.usage_disk ?? false,
+				usage_disk: monitor.thresholds?.usage_disk * 100?.toString() || "",
+				temperature: !!monitor.thresholds?.usage_temperature ?? false,
```
ajhollid: This is complex and misleading due to a misunderstanding of how the nullish coalescing operator works.

Double negation always returns a `boolean`, so the righthand of the nullish coalescing operator is always ignored.

The [nullish coalescing operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing) only returns when the left hand operand is  `null` or `undefined`.

Additionally I am hard pressed to think of a situation where double negation is easier to read or more correct than some other logical operator.  Please avoid their use unless there is a very good reason to do so.

client/src/Pages/Infrastructure/Create/index.jsx
```diff
@@ -87,35 +91,55 @@ const CreateInfrastructureMonitor = () => {
...
+				url: monitor.url?.replace(/^https?:\/\//, "") || "",
+				name: monitor.name || "",
+				notifications: monitor.notifications || [],
+				interval: monitor.interval / MS_PER_MINUTE,
+				cpu: !!monitor.thresholds?.usage_cpu ?? false,
+				usage_cpu: monitor.thresholds?.usage_cpu * 100?.toString() || "",
+				memory: !!monitor.thresholds?.usage_memory ?? false,
+				usage_memory: monitor.thresholds?.usage_memory * 100?.toString() || "",
+				disk: !!monitor.thresholds?.usage_disk ?? false,
+				usage_disk: monitor.thresholds?.usage_disk * 100?.toString() || "",
```
ajhollid: Why is nullish coalescing operator used sometimes and logical or used sometimes?  Let's be consistent with our operators.

ajhollid (changes requested):
The `useEffect` hook in the infrastruture page needs some work.  There are some logical errors and the whole `if/else` block has considerable repeated logic and should be refactored.

singh-kanwarpreet:
Ok sir

client/src/Pages/Infrastructure/Create/index.jsx
singh-kanwarpreet: ok sir

ajhollid (approved):

PR #2771, opened 2025-08-13
Fix: Resolve login error handling issues

### LINKED ISSUE #2768: Login error handling issues
**Describe the bug**
Description:
There are a couple of problems in the login flow:

1. Backend is not returning the correct error message — it sends authIncorrectPassword, which looks like a localization key, not the actual user-facing message.

2.  Frontend doesn’t properly display errors — it should display whatever error the backend returns, and if there’s nothing from the backend, show a default error message. It shouldn’t default to “incorrect password” unless that’s the actual error.

**To Reproduce**
1. Go to the Login page.
2. Enter a invalid email and a  password.
3. Observe that the backend returns authIncorrectPassword as the message.

**Expected behavior**
Backend should return a valid, translatable error key or a proper message.

Frontend should display the exact message from the backend.

gorkem-bwl:
## Summary
Fixes critical login flow issue where backend was returning raw localization keys (`authIncorrectPassword`) instead of proper user-facing error messages. This issue made error handling brittle and caused poor user experience during authentication failures.

## Problem Description
- **Backend**: Translation service was looking for locale files in wrong directory, causing translations to fail and return raw keys
- **Frontend**: Error handling was hardcoded to only work with exact string matches, making it fragile for different error scenarios
- **Security**: Login errors revealed whether a user existed vs wrong password

### Backend Fixes
1. **Fixed translation service path** (`translationService.js`):
   - Changed from `locales/` to `src/locales/` directory
   - Now properly loads translation files and returns translated messages

2. **Improved login security** (`userService.js`):
   - Added try-catch to handle both "user not found" and "incorrect password" scenarios
   - Returns generic "Incorrect password" message for security (doesn't reveal if user exists)

### Frontend Fixes  
3. **Made error handling generic** (`Login/index.jsx`):
   - Removed brittle hardcoded string matching (`=== "Incorrect password"`)
   - Now displays whatever error message backend sends with proper fallbacks
   - Uses HTTP status codes (401) for field-level errors instead of string matching

## Testing Results
✅ **Before Fix**: Backend returned `"authIncorrectPassword"` (raw key)  
✅ **After Fix**: Backend returns `"Incorrect password"` (translated message)

- Invalid email scenarios properly handled
- Valid email with wrong password returns same generic message for security
- Frontend displays backend error messages correctly with fallbacks
- Network errors handled gracefully

## Impact
- Fixes user-facing error messages in login flow
- Improves security by not revealing user existence
- Makes error handling robust and extensible for future error scenarios
- Resolves translation system issues across the application

Fixes #2768

### DIFF
--- a/server/src/service/business/userService.js
+++ b/server/src/service/business/userService.js
@@ -83,24 +83,30 @@ class UserService {
 
 	loginUser = async (email, password) => {
 		// Check if user exists
-		const user = await this.db.userModule.getUserByEmail(email);
-		// Compare password
-		const match = await user.comparePassword(password);
-		if (match !== true) {
+		try {
+			const user = await this.db.userModule.getUserByEmail(email);
+			// Compare password
+			const match = await user.comparePassword(password);
+			if (match !== true) {
+				throw this.errorService.createAuthenticationError(this.stringService.authIncorrectPassword);
+			}
+			
+			// Remove password from user object.  Should this be abstracted to DB layer?
+			const userWithoutPassword = { ...user._doc };
+			delete userWithoutPassword.password;
+			delete userWithoutPassword.avatarImage;
+
+			// Happy path, return token
+			const appSettings = await this.settingsService.getSettings();
+			const token = this.issueToken(userWithoutPassword, appSettings);
+			// reset avatar image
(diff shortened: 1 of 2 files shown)


gorkem-bwl:
@karenvicent fixed conflicts.

PR #2800, opened 2025-08-15
Fix: resolve erroneous PageSpeed settings validation

### LINKED ISSUE #2799: PageSpeed Monitor URL Validation Prevents Creation
**Describe the bug**
Running beta 3, I’m unable to add a PageSpeed monitor because the _URL to monitor_ field’s validation says that a port is required even though providing one has no impact.

**To Reproduce**
1. Go to _Pagespeed_
2. Click _Let’s create your first PageSpeed monitor!_
3. In the _URL to monitor_ field, enter a domain name like `example.com`.
4. See “Port is required for port and game monitors.”
5. Append a port number like `example.com:443`.
6. See that the validation error from step 4 remains.

**Expected behavior**
I would expect to be able to add a domain name and create a monitor, and based on the validation error at least provide an explicit port number in order to proceed. But _Create monitor_ is continually disabled.

**Screenshots**

Image

 - Browser: Safari
 - Version: Version 18.6

**Additional context**

I also tried prepending the field value with `http://` and `https://`, which made no difference. This seems like an issue with how validation errors are applied to the fields, because the _Monitor display name_ field will also display the same error (“Port is required for port and game monitors.”) when presumably any string will do.

Minor thing: Pagespeed in the sidebar has a lowercase S, when it should be PageSpeed like it is everywhere else.

mattstein:
## Describe your changes

Fixes #2799 

The field-level validation in `handleChange` was only passing the single changed field to the validator, missing the type context needed for conditional port validation rules. After that, the field validation read “"notifications" is not allowed”, so this also ensures that `notifications` are declared as an optional array.

With both of these changes made, I’m able to successfully create a PageSpeed monitor as expected, with and without selecting notifications.

## Checklist

- ~~I have added i18n support to visible strings (instead of `<div>Add</div>`).~~ No new strings added.
- ~~I made sure font sizes, color choices etc are all referenced from the theme. I don't have any hardcoded dimensions.~~ No visual changes applied.
- ~~I took a screenshot or a video and attached to this PR if there is a UI change.~~ No UI change unless you count the absence of validation errors.

## Summary by CodeRabbit

* **Bug Fixes**
  * Form validation now considers the selected monitor type when editing fields, ensuring rules are applied correctly for each type.
  * Reduces false validation errors and prevents invalid field/type combinations during setup.

### DIFF
--- a/client/src/Pages/PageSpeed/Create/index.jsx
+++ b/client/src/Pages/PageSpeed/Create/index.jsx
@@ -152,7 +152,7 @@ const PageSpeedSetup = () => {
 		});
 
 		const { error } = monitorValidation.validate(
-			{ ...monitor, [name]: value },
+			{ [name]: value, type: monitor.type },
 			{ abortEarly: false }
 		);
 		setErrors((prev) => ({
--- a/client/src/Pages/PageSpeed/Create/index.jsx
+++ b/client/src/Pages/PageSpeed/Create/index.jsx
@@ -152,7 +152,7 @@ const PageSpeedSetup = () => {
 		});
 
 		const { error } = monitorValidation.validate(
-			{ [name]: value },
+			{ ...monitor, [name]: value },
 			{ abortEarly: false }
 		);
 		setErrors((prev) => ({
--- a/client/src/Validation/validation.js
+++ b/client/src/Validation/validation.js
@@ -213,7 +213,6 @@ const monitorValidation = joi.object({
 		}),
 		otherwise: joi.string().allow(null, ""),
 	}),
-	notifications: joi.array().optional(),
 });
 
 const imageValidation = joi.object({
--- a/client/src/Validation/validation.js
+++ b/client/src/Validation/validation.js
@@ -213,6 +213,7 @@ const monitorValidation = joi.object({
 		}),
 		otherwise: joi.string().allow(null, ""),
 	}),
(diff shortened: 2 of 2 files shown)


client/src/Pages/PageSpeed/Create/index.jsx
```diff
@@ -152,7 +152,7 @@ const PageSpeedSetup = () => {
 		});
 
 		const { error } = monitorValidation.validate(
-			{ [name]: value },
+			{ ...monitor, [name]: value },
```
ajhollid: @mattstein This will vaidate the entire monitor, which isn't what we want to do for the field level validation.

It should be sufficient to just provide the type here to resolve the validation issue.

client/src/Validation/validation.js
```diff
@@ -213,6 +213,7 @@ const monitorValidation = joi.object({
 		}),
 		otherwise: joi.string().allow(null, ""),
 	}),
+	notifications: joi.array().optional(),
```
ajhollid: I believe this won't be necessary after adding only "type" to the validation above.

mattstein:
Thanks @ajhollid! I just made those updates and tested locally to confirm joy.

ajhollid (approved):
Thanks for the revision and for resolving this issue so quickly! Much appreciated

PR #2807, opened 2025-08-18
feat: unify check models

ajhollid:
This PR removes `PagespeedCheck` and `HardwareCheck` models in favaour of a single unified `Check` type.

This will simplify queries and reduce complexity throughout the application.  It also opens the door to implementing an "Incident" model to better handle incidents and alerts.

- [x] Remove Pagespeed and Hardware check models
- [x] Update queries and job templates.
- [x] Update all associated services  

## Summary by CodeRabbit

- Bug Fixes
  - Fixed fallback background rendering to ensure transparent background displays correctly.
  - Prevented crashes in the Pie Chart Legend when audit data is missing.

- Refactor
  - Unified all check data into a single checks stream; specialized hardware, network, and PageSpeed check types removed.
  - History and dashboards now order/filter by last-updated time.
  - Simplified buffering to a single batch flush for more reliable data ingestion.

- Metrics
  - Removed the “uptime burn” metric and streamlined status tracking.

### DIFF
--- a/server/src/db/mongo/modules/monitorModuleQueries.js
+++ b/server/src/db/mongo/modules/monitorModuleQueries.js
@@ -987,7 +987,7 @@ const buildMonitorsWithChecksByTeamIdPipeline = ({ matchStage, filter, page, row
 
 	// Add checks
 	if (limit) {
-		let checksCollection = "checks";
+		const checksCollection = "checks";
 		monitorsPipeline.push({
 			$lookup: {
 				from: checksCollection,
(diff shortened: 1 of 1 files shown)


server/src/db/mongo/modules/monitorModuleQueries.js
```diff
@@ -987,11 +988,6 @@ const buildMonitorsWithChecksByTeamIdPipeline = ({ matchStage, filter, page, row
 	// Add checks
 	if (limit) {
 		let checksCollection = "checks";
```
Br0wnHammer: since checksCollection won't change now. We can use convert this to a const.

server/src/db/mongo/modules/monitorModuleQueries.js
```diff
@@ -1057,11 +1053,7 @@ const buildFilteredMonitorsByTeamIdPipeline = ({ matchStage, filter, page, rowsP
...
```
Br0wnHammer: same here.

PR #2870, opened 2025-08-28
Add direct user registration for team management

### LINKED ISSUE #2822: Add team member without invite
Some teams run self-hosted instances with no outbound email. They still need a quick way to create user accounts.

Allow an admin to add a team member by entering an email and a password directly. No invitation email, no magic link. The user can sign in immediately with those credentials.

Admin UI: 

- New button in “Team” page: “Add team member”.
- Modal with fields: Email (required), password (required, show eye toggle) and role select which defaults to “Member”, and which only “Admin” can create.

Submit creates the account and shows a success toast. “User created. Share credentials with the member securely.”

Email must be RFC compliant and reject if email already exists.

karenvicent:
## Describe your changes
Allows administrators to register team members directly without sending invitation emails. Admins can create user accounts immediately and share credentials directly.

### What's Added
- **New `AddTeamMember` component** - Modal form for direct user registration
- **Enhanced team workflow** - Choose between email invitations or direct registration
- **Auto team list refresh** - Updates team panel after new member creation
- **Complete form validation** - All fields validated 

### Benefits
- ✅ Faster team member onboarding
- ✅ No dependency on email delivery  
- ✅ Immediate credential sharing
- ✅ Enhanced admin control

Fixes #2822 

```Javascript
<div>{t('add')}</div>
```

Captura de pantalla 2025-08-28 a la(s) 1 49 17 p m

Captura de pantalla 2025-08-28 a la(s) 1 49 39 p m

### DIFF
--- a/client/src/Hooks/inviteHooks.js
+++ b/client/src/Hooks/inviteHooks.js
@@ -33,6 +33,7 @@ const useGetInviteToken = () => {
 			}
 
 			setToken(inviteLink);
+			return token;
 		} catch (error) {
 			setError(error);
 		} finally {
--- a/client/src/Pages/Account/components/AddMemberMenu/index.jsx
+++ b/client/src/Pages/Account/components/AddMemberMenu/index.jsx
@@ -1,0 +1,68 @@
+import { useState } from "react";
+import Button from "@mui/material/Button";
+import Menu from "@mui/material/Menu";
+import MenuItem from "@mui/material/MenuItem";
+import { useTheme } from "@emotion/react";
+import ArrowDropDownIcon from "@mui/icons-material/ArrowDropDown";
+import { useTranslation } from "react-i18next";
+import Proptypes from "prop-types";
+
+const AddMemberMenu = ({ handleInviteOpen, handleIsRegisterOpen }) => {
+	const [anchorEl, setAnchorEl] = useState(null);
+	const open = Boolean(anchorEl);
+	const { t } = useTranslation();
+	const theme = useTheme();
+	const handleClick = (event) => {
+		setAnchorEl(event.currentTarget);
+	};
+	const handleClose = () => {
+		setAnchorEl(null);
+	};
+
+	return (
+		<>
+			<Button
+				variant="contained"
+				color="accent"
+				endIcon={<ArrowDropDownIcon sx={{ color: theme.palette.secondary.light }} />}
+				onClick={handleClick}
+			>
+				Add Team Member
+			</Button>
+			<Menu
+				anchorEl={anchorEl}
+				open={open}
+				onClose={handleClose}
+				sx={{
+					"& .MuiPaper-root": {
+						minWidth: anchorEl?.offsetWidth || "auto",
+					},
+				}}
+			>
+				<MenuItem
+					onClick={() => {
+						handleClose();
+						handleInviteOpen();
+					}}
+				>
+					{t("teamPanel.inviteTeamMember")}
+				</MenuItem>
+				<MenuItem
+					onClick={() => {
+						handleClose();
+						handleIsRegisterOpen(true);
+					}}
+				>
+					{t("teamPanel.register")}
+				</MenuItem>
+			</Menu>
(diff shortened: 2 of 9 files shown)


client/src/Hooks/inviteHooks.js
```diff
@@ -33,6 +33,7 @@ const useGetInviteToken = () => {
 			}
 
 			setToken(inviteLink);
+			return token;
```
ajhollid, line 35: @karenvicent we don't need to return the token here, it is in the return values from the hook:

```
	return [getInviteToken, clearToken, isLoading, error, token];
```

karenvicent: Hi Alex, is fixed I think that the solution is much better now

client/src/Hooks/inviteHooks.js
```diff
@@ -14,33 +14,52 @@ const useGetInviteToken = () => {
...
+			const toSubmit = {
+				...formData,
+				inviteToken: token,
+			};
+			delete toSubmit.confirm;
+			const responseRegister = await networkService.registerUser(toSubmit);
+			return responseRegister;
 		} catch (error) {
 			setError(error);
+			throw error;
```
Br0wnHammer, line 56: Instead of throwing the whole error message. Should we send a custom message?

client/src/Pages/Account/components/AddMemberMenu/index.jsx
```diff
@@ -0,0 +1,68 @@
...
+
+	return (
+		<>
+			<Button
+				variant="contained"
+				color="accent"
+				endIcon={<ArrowDropDownIcon sx={{ color: theme.palette.secondary.light }} />}
+				onClick={handleClick}
+			>
+				Add Team Member
```
Br0wnHammer: This string can be localized.

client/src/Pages/Account/components/AddTeamMember/hooks/useAddTeamMember.jsx
```diff
@@ -0,0 +1,44 @@
...
+			context: { password: formData.password },
+		});
+		const formErrors = {};
+		if (error) {
+			for (const err of error.details) {
+				formErrors[err.path[0]] = err.message;
+			}
+		}
+		if (!role[0] || role.length === 0) {
+			formErrors.role = "Role is required";
```
Br0wnHammer: This too can be localized.

client/src/Pages/Account/components/AddTeamMember/index.jsx
```diff
@@ -0,0 +1,212 @@
...
+
+		if (name === "password" || name === "confirm") {
+			handlePasswordFeedback(updatedForm, name, value, form, errors, setErrors);
+		}
+	};
+
+	const onsubmitAddMember = async (event) => {
+		event.preventDefault();
+		if (!validateForm(form, role)) return;
+		setIsLoadingSubmit(true);
```
Br0wnHammer: We can use this state variable inside the try catch block for a better robustness

client/src/Pages/Account/components/TeamPanel.jsx
```diff
@@ -65,7 +67,10 @@ const TeamPanel = () => {
 			render: (row) => row.role,
 		},
 	];
-
+	const [refreshTrigger, setRefreshTrigger] = useState(0);
```
Br0wnHammer: Generally for flag triggers we use `false` as the state.

Br0wnHammer (changes requested):
Looks good for the most part. As always clean code from @karenvicent . I have left some comments please review them.

client/src/Hooks/inviteHooks.js
karenvicent: @Br0wnHammer Thanks for the feedback this is a  good point! The reason I kept it like this is because networkService.registerUser already returns a custom error message. Here I just catch it to update the hook state and then re-throw it, so the component can handle that same message consistently.

PR #2895, opened 2025-08-31
fix: resolve Sharp module Alpine Linux compatibility

gorkem-bwl:
## Summary
- Fix Sharp module loading error on Alpine Linux containers
- Add required dependencies for Sharp compilation on Alpine
- Resolve "Could not load sharp module using linuxmusl-x64 runtime" error

## Changes Made
- **Added Alpine dependencies**: `vips-dev`, `vips-tools`, `python3`, `make`, `g++`
- **Added Sharp rebuild**: `npm rebuild sharp` after install
- **Updated Dockerfiles**: prod, staging, dist-mono, dist-arm server containers

## Root Cause
Sharp's prebuilt binaries weren't compatible with Alpine's musl libc. The fix ensures Sharp builds correctly for the Alpine environment by:
1. Installing native dependencies required for Sharp/libvips
2. Rebuilding Sharp to compile against the correct runtime

## Testing
This resolves the deployment issue where containers would crash with:
```
Error: Could not load the "sharp" module using the linuxmusl-x64 runtime
```

Fixes deployment failures on both `checkmate-backend-mono:latest` and `checkmate-backend-mono-multiarch:latest`.

## Summary by CodeRabbit

- Bug Fixes
  - Resolved image processing failures by ensuring native modules are correctly built, improving reliability across Alpine/ARM and multi-arch environments.

- Chores
  - Updated server container builds (staging, prod, mono, arm) to include required system libraries and install optional dependencies.
  - Rebuilt native dependencies during image build to align with system libraries.
  - Set a Node.js memory limit to reduce out-of-memory issues during builds and runtime.

PR #2897, opened 2025-08-31
remove: eliminate Sharp dependency for simpler Docker deployment

gorkem-bwl:
## Summary
Alternative solution to the Sharp Alpine compatibility issue. Instead of fixing Docker configuration, this PR removes the Sharp dependency entirely and achieves the same visual result with CSS.

## Changes Made
- **Removed Sharp dependency** from package.json (eliminates ~100MB from Docker images)
- **Updated `GenerateAvatarImage`** to return original image as base64 without resizing
- **CSS handles display scaling**: Frontend can use `object-fit: cover` with `width: 64px, height: 64px`
- **Updated tests** to reflect new behavior (no server-side resizing)

## Benefits vs Alpine Fix Approach
- ✅ **Eliminates Sharp Alpine issues permanently** - no platform-specific compilation
- ✅ **Reduces Docker image size** by ~100MB (no native deps)
- ✅ **Faster builds** - no compilation of native binaries
- ✅ **Works on all platforms** - pure JavaScript solution
- ✅ **Same visual result** - CSS handles the 64x64 display

## Technical Details
**Before**: Sharp resizes images server-side to 64x64 pixels
**After**: Original images stored, CSS handles display sizing:
```css
img.avatar { 
  width: 64px; 
  height: 64px; 
  object-fit: cover; 
}
```

## Comparison to PR #2895
- **This PR**: Remove dependency entirely (cleaner solution)
- **PR #2895**: Fix Alpine compatibility by adding build tools
- Both solve the deployment issue - team can choose approach

Fixes Sharp Docker deployment errors without adding complexity.

## Summary by CodeRabbit

- Refactor
  - Simplified avatar image handling by removing server-side resizing; avatars are now displayed via client-side styling for consistent 64×64 presentation.
- Documentation
  - Clarified that avatar rendering is handled on the client.
- Chores
  - Removed an image-processing dependency to reduce server footprint.
- Tests
  - Updated tests to assert base64 encoding of original image data and simplified error case handling.

### DIFF
--- a/client/src/Components/Avatar/index.jsx
+++ b/client/src/Components/Avatar/index.jsx
@@ -41,6 +41,10 @@ const Avatar = ({ src, small, sx, onClick = () => {} }) => {
 				color: theme.palette.accent.contrastText,
 				backgroundColor: theme.palette.accent.main, // Same BG color as checkmate BG in sidebar
 				display: "inline-flex",
+				// Ensure images scale properly when server no longer resizes to 64x64
+				"& img": {
+					objectFit: "cover",
+				},
 				/* 
 				TODO not sure what this is for*/
 				"&::before": {
--- a/server/package-lock.json
+++ b/server/package-lock.json
@@ -37,7 +37,6 @@
 				"nodemailer": "^6.9.14",
 				"papaparse": "^5.5.2",
 				"ping": "0.4.4",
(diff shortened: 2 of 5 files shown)


ajhollid:
I believe this has been resovled now

PR #2916, opened 2025-09-01
Fix skeleton flash on empty monitor pages

gorkem-bwl:
## Summary
Prevents skeleton/table flashing before empty state info pages are displayed.

## Changes
- Infrastructure monitors: Added initialization tracking to prevent skeleton flash
- PageSpeed monitors: Added initialization tracking to prevent skeleton flash  
- Uptime monitors: Added initialization tracking to prevent skeleton flash
- Maintenance windows: Added initialization check to prevent table flash

## Behavior
Empty monitor pages now show info content immediately without visual flashes.

### DIFF
--- a/client/src/Pages/Infrastructure/Monitors/index.jsx
+++ b/client/src/Pages/Infrastructure/Monitors/index.jsx
@@ -33,6 +33,7 @@ const InfrastructureMonitors = () => {
 	const [toFilterStatus, setToFilterStatus] = useState(undefined);
 	const [search, setSearch] = useState(undefined);
 	const [isSearching, setIsSearching] = useState(false);
+	const [hasInitialized, setHasInitialized] = useState(false);
 
 	// Utils
 	const theme = useTheme();
@@ -81,6 +82,17 @@ const InfrastructureMonitors = () => {
 		updateTrigger,
 	});
 
+	// Track initialization to prevent skeleton flash
+	useEffect(() => {
+		if (!isLoading && (summary !== undefined || monitors !== undefined)) {
+			setHasInitialized(true);
+		}
+	}, [isLoading, summary, monitors]);
+
+	// Show empty state when no monitors exist
+	const hasNoMonitors = hasInitialized && typeof summary?.totalMonitors === "undefined";
+	const isEmpty = hasInitialized && summary?.totalMonitors === 0;
+
 	if (networkError === true) {
 		return (
 			<GenericFallback>
@@ -96,7 +108,7 @@ const InfrastructureMonitors = () => {
 		);
 	}
 
-	if (!isLoading && typeof summary?.totalMonitors === "undefined") {
+	if (hasNoMonitors || isEmpty) {
 		return (
 			<Fallback
 				type="infrastructureMonitor"
@@ -106,6 +118,11 @@ const InfrastructureMonitors = () => {
 				isAdmin={isAdmin}
 			/>
 		);
+	}
+
+	// Don't render anything until we've initialized to prevent skeleton flash
+	if (!hasInitialized) {
+		return null;
 	}
 
 	return (
--- a/client/src/Components/Charts/StatusPageBarChart/index.jsx
+++ b/client/src/Components/Charts/StatusPageBarChart/index.jsx
@@ -22,7 +22,7 @@ const Bar = ({ width, height, backgroundColor, borderRadius, children }) => {
 
 	return (
 		<Box
-            width={width}
+			width={width}
 			position="relative"
 			height={height}
 			backgroundColor={backgroundColor}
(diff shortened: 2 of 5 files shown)


ajhollid:
We should implement a consistent loading pattern for all pages in the application rather than implementing piecemeal on each page.

A base component is probably the ideal solution here, @karenvicent thoughts here?

client/src/Pages/Infrastructure/Monitors/index.jsx
```diff
@@ -33,6 +33,7 @@ const InfrastructureMonitors = () => {
 	const [toFilterStatus, setToFilterStatus] = useState(undefined);
 	const [search, setSearch] = useState(undefined);
 	const [isSearching, setIsSearching] = useState(false);
+	const [hasInitialized, setHasInitialized] = useState(false);
```
karenvicent, line 36: Hi Gorkem, what do you think about deriving the value directly instead of using useState?
`const hasData = summary !== undefined || monitors !== undefined;`
`const hasInitialized = !isLoading && hasData;` 

This could reduce the complexity of the code for future maintenance

client/src/Pages/Infrastructure/Monitors/index.jsx
```diff
@@ -81,6 +82,17 @@ const InfrastructureMonitors = () => {
 		updateTrigger,
 	});
 
+	// Track initialization to prevent skeleton flash
+	useEffect(() => {
```
karenvicent, line 86: Instead of using useEffect, we could define derived variables like:
 `	const hasNoMonitors = summary?.totalMonitors == null;`
	`const isEmpty = summary?.totalMonitors === 0;`
	`const showEmpty = hasNoMonitors || isEmpty;`

client/src/Pages/Infrastructure/Monitors/index.jsx
```diff
@@ -96,7 +108,7 @@ const InfrastructureMonitors = () => {
 		);
 	}
 
-	if (!isLoading && typeof summary?.totalMonitors === "undefined") {
+	if (hasNoMonitors || isEmpty) {
```
karenvicent, line 111: And here we could simplify the condition by directly returning the fallback when either the data is initialized or explicitly empty:
`if (hasInitialized || showEmpty) {
		return (
			<Fallback
				type="infrastructureMonitor"
				title={t("infrastructureMonitor.fallback.title")}
				checks={t("infrastructureMonitor.fallback.checks", { returnObjects: true })}
				link="/infrastructure/create"
				isAdmin={isAdmin}
			/>
		);
	}`

karenvicent:
> We should implement a consistent loading pattern for all pages in the application rather than implementing piecemeal on each page.
> 
> A base component is probably the ideal solution here, @karenvicent thoughts here?

@ajhollid  I think a reusable hook could be a good option here. Since all pages follow the same “wait until initial data is ready” logic, a hook would allow us to centralize it and keep the loading pattern consistent across the application.

ajhollid:
@karenvicent do you want to take this on?

We can do it in a series of PRs

1.  Create the hook
2.  Apply the hook to pages, one PR for each page

That should keep it nice and simple

karenvicent:
> @karenvicent do you want to take this on?
> 
> We can do it in a series of PRs
> 
> 1. Create the hook
> 2. Apply the hook to pages, one PR for each page
> 
> That should keep it nice and simple

@ajhollid Sure, I’d be happy to take this on! I’ll start with the hook and then move forward applying it to each page in separate PRs.

gorkem-bwl:
> > @karenvicent do you want to take this on?
> > We can do it in a series of PRs
> > 
> > 1. Create the hook
> > 2. Apply the hook to pages, one PR for each page
> > 
> > That should keep it nice and simple
> 
> @ajhollid Sure, I’d be happy to take this on! I’ll start with the hook and then move forward applying it to each page in separate PRs.

@karenvicent did you want me to fix the pr in any way, and/or restructure it to your needs here?

karenvicent:
> > > @karenvicent do you want to take this on?
> > > We can do it in a series of PRs
> > > 
> > > 1. Create the hook
> > > 2. Apply the hook to pages, one PR for each page
> > > 
> > > That should keep it nice and simple
> > 
> > 
> > @ajhollid Sure, I’d be happy to take this on! I’ll start with the hook and then move forward applying it to each page in separate PRs.
> 
> @karenvicent did you want me to fix the pr in any way, and/or restructure it to your needs here?

@gorkem-bwl No worries, you can leave it as it is. This gives me a good foundation to continue, and I’ll take care of moving the logic into a reusable hook in the upcoming work.

ajhollid:
Resolved in another PR by @karenvicent

PR #3028, opened 2025-10-20
Remove unused CSS variables

### LINKED ISSUE #1025: FE - Refactor unused CSS Variables
There are unused CSS variables. We need to understand if we are going to need them, where the project's theme lives, and keep everything in the same place.

gorkem-bwl:
## Summary
Removes 13 unused CSS variables that were leftovers from when the project switched to the MUI theme system.

## Changes
- Removed unused CSS variables from `client/src/index.css`
- Kept 19 variables that are still actively used

Fixes #1025

## Summary by CodeRabbit

* **Refactor**
  * Removed unused CSS design token variables from the styling system.

### DIFF
--- a/client/src/index.css
+++ b/client/src/index.css
@@ -15,15 +15,11 @@ html {
 	-moz-osx-font-smoothing: grayscale;
 
 	/* Generalized Stylings */
-	--env-var-radius-1: 4px;
 	--env-var-radius-2: 8px;
 
-	--env-var-width-1: 100vw;
 	--env-var-width-2: 360px;
-	--env-var-width-3: 250px;
 	--env-var-width-4: 100px;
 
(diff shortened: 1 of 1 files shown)


client/src/index.css
```diff
@@ -15,15 +15,11 @@ html {
 	-moz-osx-font-smoothing: grayscale;
```
Owaiseimdad, line 15: @gorkem-bwl  any reason to remove it. Are we planning to use them at all or is there a framework we have internally which will not allow it.
I mean yes its not being used but that should be a reason to remove it from code. If its not up to the css framework we have then we can remove. But its allowed then we can keep it and not wait for a pr which will add some of it again. Extra work, let me know if I am missing anything.

PR #3045, opened 2025-10-29
Feat: Pagespeed Graph Pagination Backend

### LINKED ISSUE #606: FE&BE - Scrollable Pagespeed Graph
Pagespeed checks need to be limited to a reasonable number, and it doesn't make a whole lot of sense to pick every `nth` check like we did with Monitor Details.

A visually intesresting implementation might be a scrollable graph which fetches data in paginated chunks.

            [Preloaded Chunk] <-> [ Displayed Chunk ] <-> [Preloaded Chunk]
            
We would fetch one page of offscreen data in each direction to ensure smooth scrolling

mospell:
## Describe your changes

This PR adds pagination support for the PageSpeed chart on the backend by introducing an optional variable to FetchStatsByMonitorId/GetMonitorStatsById called `page`, which shifts back the date range depending on it's value.

For example: If `dateRange` is set to "day" and `page` is 4, then it will query data from 4 days ago.

The returned `monitor` object now also includes the number of remaining checks to be used for the frontend implementation.

Fixes #606  

```Javascript
<div>{t('add')}</div>
```

## Summary by CodeRabbit

* **New Features**
  * Added page-based pagination for monitor statistics so users can navigate older/newer data when reviewing monitor performance.
  * Monitor statistics now include a "remaining checks" count showing how many checks exist beyond the current page.
  * Paging is preserved across requests and respects selected date ranges, improving navigation through historical data.

* **Validation**
  * Query accepts a non-negative integer page parameter to control paging.

### DIFF
--- a/client/src/Hooks/v1/monitorHooks.js
+++ b/client/src/Hooks/v1/monitorHooks.js
@@ -170,7 +170,7 @@ const useFetchStatsByMonitorId = ({
 	dateRange,
 	numToDisplay,
 	normalize,
-	pageOffset,
+	page,
 	updateTrigger,
 }) => {
 	const [monitor, setMonitor] = useState(undefined);
@@ -188,7 +188,7 @@ const useFetchStatsByMonitorId = ({
 					dateRange,
 					numToDisplay,
 					normalize,
-					pageOffset,
+					page,
 				});
 				setMonitor(res?.data?.data ?? undefined);
 				setAudits(res?.data?.data?.checks?.[0]?.audits ?? undefined);
@@ -200,7 +200,16 @@ const useFetchStatsByMonitorId = ({
 			}
 		};
 		fetchMonitor();
-	}, [monitorId, dateRange, numToDisplay, normalize, sortOrder, limit, pageOffset, updateTrigger]);
+	}, [
+		monitorId,
+		dateRange,
+		numToDisplay,
+		normalize,
+		sortOrder,
+		limit,
+		page,
+		updateTrigger,
+	]);
 	return [monitor, audits, isLoading, networkError];
 };
 
--- a/client/src/Hooks/v1/monitorHooks.js
+++ b/client/src/Hooks/v1/monitorHooks.js
@@ -170,6 +170,7 @@ const useFetchStatsByMonitorId = ({
 	dateRange,
 	numToDisplay,
 	normalize,
+	pageOffset,
 	updateTrigger,
 }) => {
 	const [monitor, setMonitor] = useState(undefined);
@@ -187,6 +188,7 @@ const useFetchStatsByMonitorId = ({
 					dateRange,
 					numToDisplay,
 					normalize,
+					pageOffset,
 				});
 				setMonitor(res?.data?.data ?? undefined);
 				setAudits(res?.data?.data?.checks?.[0]?.audits ?? undefined);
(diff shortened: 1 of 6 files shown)


ajhollid (changes requested):
Hey @mospell ,

Thanks for your contribution here. 

If the backend only takes a page offset that leaves calculating the page offset to the frontend.  I believe that this should be a backend operation, and the frontend shouldn't know or care what the page offset is.

The frontend should only care about what page it is on, and how many rows per page to display.

In light of that, the backend should take two variables, the `page` and `rowsPerPage`, and multiplying the two together will yield the page offset.

You can see how pagination is handled for other monitor types on both the frontend and the backend for reference.

mospell:
Hi @ajhollid, 

`pageOffset` would basically increment by 1 until it runs out of `remainingChecks` blocking you from scrolling further when there is no more data. I probably should've just named it `page` instead now that I think about it.

> The backend should take two variables, the page and rowsPerPage, and multiplying the two together will yield the page offset.

My initial idea was time-based pagination, as querying an x amount of checks doesn't really make sense on a graph. How would rowsPerPage apply here? Could you give an example?

gorkem-bwl:
@mospell you have a CI/CD check issue here.

mospell:
Hi @ajhollid, just pinging you again

The way it currently works is that there is no pageOffset calculation in the frontend. All offset calculations happen in the backend as you requested. To clarify this I have renamed the `pageOffset` variable to just `page`. 

I believe it would make more sense to have the user scroll through previous dates instead of displaying a number of rows per page. I think the way it currently works should satisfy your requirements. If you'd like me to change it anyway, please let me know.

mospell:
@gorkem-bwl this should be resolved now, thanks for the notice!

ajhollid:
@mospell perhaps I'm misunderstanding somethig here, pagination of results should depend on a `page` and a `rowsPerPage` input from the front end should it not?

ie on the frontend, the user selects "5 rows per page" and then paginates thorugh results, starting from page 0.

Offset is then calcualted by multiplying` rowsPerPage * page` on the backend.

Example:

```
let skip = 0;
if (page && rowsPerPage) {
	skip = page * rowsPerPage;
}

const checks = await this.Check.aggregate([
	{ $match: matchStage },
	{ $sort: { createdAt: sortOrder } },
	{
		$facet: {
			summary: [{ $count: "checksCount" }],
			checks: [{ $skip: skip }, { $limit: rowsPerPage }],
		},
	},
	{
		$project: {
			checksCount: {
				$ifNull: [{ $arrayElemAt: ["$summary.checksCount", 0] }, 0],
			},
			checks: {
				$ifNull: ["$checks", []],
			},
		},
	},
]);
```

mospell:
@ajhollid Usually yes, but I decided to go for time-based pagination here. In this case dateRange would kind of act as rowsPerPage here.

On the frontend the user would pick a date range and then paginate through results that way, starting from page 0 (today).

Instead of calculating an offset based on rowsPerPage and page, the backend shifts back the selected dateRange by the  page number. 

So for example:
`dateRange:"day"` and `page:5` - returns data from 5 days ago
`dateRange:"week"` and `page:2` - returns data from 2 weeks ago.
and so on, until there are no more checks.

This is done by the `offsetDatesByPage` helper function in monitorService.js, feel free to take a look at it if this reply still ends up being a bit confusing.

mospell:
@ajhollid This implementation was partially inspired by other projects like Thanos, where you can select the exact date range to display data on a time based chart. Pagination is generally used for tables or text, and while it can be done for charts it's a little unintuitive from the perspective of a user. 

I think that "displaying data from (date) to (date)" is a lot more readable to the average person than "displaying x checks" on a time based chart. 
It's also way more consistent because it wouldn't be affected by, for example, pausing the monitor or editing the check interval, both of which could create gaps between checks.  

Please reconsider this request. I'd be happy to make any other tweaks to the way it's currently implemented, but if you'd still like me to add regular pagination I can do that too.

PR #3061, opened 2025-11-10
fix: error resolve for signup user when password contains <> sign

### LINKED ISSUE #3010: 'Password cannot be empty' user signup bug on fresh install
**Describe the bug**
When using the following password on a completely fresh install, a faulty error message 'password cannot be empty' appears while the password is not empty and follows all required rules.

**To Reproduce**
The password in question is `<BIBpIiDXbgdp6M0vr6SO8T7`
I suppose the `<` is the troublemaker, because when I'm changing it to `$`, the error dissapears and the user is created succesfully.

Cfr. screenshot with devtools request in question.

**Expected behavior**
The user is created with the intended password (thus `<` as first character)

**Screenshots**
Image

I also noticed the following error message in the server logs:

```
2025-10-08T21:28:36.443Z error: [authController](registerUser) "password" is not allowed to be empty (stack: [
  {
    "function": "ErrorService.createValidationError",
    "file": "file:///app/server/src/service/infrastructure/errorService.js",
    "line": 69,
    "column": 10
  },
  "at file:///app/server/src/controllers/baseController.js:26:49",
  {
    "function": "process.processTicksAndRejections",
    "file": "node:internal/process/task_queues",
    "line": 95,
    "column": 5
  }
])
```

 - Browser Chrome, latest
 - Version 3.2.0 (Latest at time of writing)

This definelty is a low priority bug, but a rather odd one suggesting an underlying issue in the password checking logic hence I'm reporting it.

Darshan98Solanki:
Hey @gorkemcetin,

Fixes #3010 

I just have encoded the password before hit request (cause it is reserved keywords) and decode that on backend before validating and insert option in db.

**Note**:- from frontend we saw that password is there in payload but when we log password in backend we see empty string like ''. So this method I have added.

Have given the **SS** for validating the changed on real time.

Signup form:-
image

Login form:-
image

After login:-
image

Please check the pr and let me know is there is any updates.

## Summary by CodeRabbit

* **Refactor**
  * Updated authentication data handling to enhance password transmission security during register and login processes.

### DIFF
--- a/client/src/Features/Auth/authSlice.js
+++ b/client/src/Features/Auth/authSlice.js
@@ -12,7 +12,8 @@ const initialState = {
 
 export const register = createAsyncThunk("auth/register", async (form, thunkApi) => {
 	try {
-		const res = await networkService.registerUser(form);
+		const reqObj = { ...form, password: encodeURIComponent(form.password) };
+		const res = await networkService.registerUser(reqObj);
 		return res.data;
 	} catch (error) {
 		if (error.response.data) {
@@ -28,7 +29,8 @@ export const register = createAsyncThunk("auth/register", async (form, thunkApi)
 
 export const login = createAsyncThunk("auth/login", async (form, thunkApi) => {
 	try {
-		const res = await networkService.loginUser(form);
+		const reqObj = { ...form, password: encodeURIComponent(form.password) };
+		const res = await networkService.loginUser(reqObj);
 		return res.data;
 	} catch (error) {
 		if (error.response && error.response.data) {
--- a/server/src/controllers/v1/authController.js
+++ b/server/src/controllers/v1/authController.js
@@ -91,6 +91,7 @@ class AuthController extends BaseController {
 			if (req.body?.email) {
 				req.body.email = req.body.email?.toLowerCase();
 			}
+			req.body.password = decodeURIComponent(req.body.password);
 			await registrationBodyValidation.validateAsync(req.body);
 			const { user, token } = await this.userService.registerUser(req.body, req.file);
 			res.success({
@@ -127,6 +128,7 @@ class AuthController extends BaseController {
 			if (req.body?.email) {
 				req.body.email = req.body.email?.toLowerCase();
 			}
+			req.body.password = decodeURIComponent(req.body.password);
 			await loginValidation.validateAsync(req.body);
 			const { user, token } = await this.userService.loginUser(req.body.email, req.body.password);
 


ajhollid:
Hey @Darshan98Solanki ,

While it works, this seems like an overly complex solution to the problem.

Moreover, it creates a tight dependency between the front and back ends, which is to be avoided.

Is there some reason the validation on the frontend can't be modified to allow for the password in question?

Darshan98Solanki:
hey, @ajhollid 

I understood your concern give me some time I will find some solution with decoupled for same.

Owaiseimdad:
> Hey @Darshan98Solanki ,
> 
> While it works, this seems like an overly complex solution to the problem.
> 
> Moreover, it creates a tight dependency between the front and back ends, which is to be avoided.
> 
> Is there some reason the validation on the frontend can't be modified to allow for the password in question?

I think we need to think in terms of regex here, let me know what solution you have thought of.

Darshan98Solanki:
@ajhollid @Owaiseimdad 

can we restrict < > this two letter for password  or else pass the payload with stringify version but that is also tightly coupled.

I think restrict this letter will be good solution ?

ajhollid:
> @ajhollid @Owaiseimdad
> 
> can we restrict < > this two letter for password or else pass the payload with stringify version but that is also tightly coupled.
> 
> I think restrict this letter will be good solution ?

Stringifying or otherwise transforming the input defeats the purpose of input validation, so let's avoid that.

The input vaidation schema should be modified to allow for the `<` and `>` characters.  Regex is probably a reasonable solution as @Owaiseimdad mentioned, why not give that a shot?

Darshan98Solanki:
New Regex:- /^(?=.*[a-z])(?=.*[A-Z])(?=.*\d)(?=.*[!?@#$%^&*()\-_=+\[\]{};:'",.~`|\\/])[A-Za-z\d!?@#$%^&*()\-_=+\[\]{};:'",.~`|\\/]{8,}$/;

Old Regex:- /^(?=.*[a-z])(?=.*[A-Z])(?=.*[0-9])(?=.*[!?@#$%^&*()\-_=+[\]{};:'",.<>~`|\\/])[A-Za-z0-9!?@#$%^&*()\-_=+[\]{};:'",.<>~`|\\/]+$/;

**Difference Note** :- 
-         More secure (blocks < and >)
-         Stronger (minimum 8 characters)

I think this regex is good for validation should I change this for validation of password ?
please give me you feedback @ajhollid @Owaiseimdad

Darshan98Solanki:
@ajhollid @Owaiseimdad  @gorkemcetin  

as per we discussed last any inputs ? should I remove <> sign from the password regex to prevent XXS attack and make it more secure also change the frontend relevant if needed.

Darshan98Solanki:
@ajhollid @Owaiseimdad @gorkemcetin 

i'm looking inputs from your what should do further?

ajhollid:
@Darshan98Solanki the new regex must be only more premissive than the original one, it cannot be any more restrictive or it could possibly disallow existing users.  So you can remove the min 8 char and it should be good to go.

Darshan98Solanki:
@ajhollid 
we also need to change the frontend as well. and give the list of special char list so the user will have clear idea about it.

gorkem-bwl:
> @ajhollid we also need to change the frontend as well. and give the list of special char list so the user will have clear idea about it.

We can do that in another PR.

gorkem-bwl:
Thank you Darshan. Closing this and we're going to implement another approach.

PR #3104, opened 2026-01-09
Redesign uptime monitor table UI components

gorkem-bwl:
## Summary
- Add SegmentedControl component for time range selection with clean, modern design
- Redesign response time visualization with stacked bars showing variable heights
- Add TypeChip component for color-coded monitor type display (HTTP, HTTPS, Ping, etc.)
- Update StatusLabel to match new consistent chip design style
- Add subtle gradient background to status summary cards

## Components added
- **SegmentedControl**: Clean segmented button group (34px height, 4px radius, generous padding)
- **TypeChip**: Color-coded chips for monitor types (HTTP=blue, HTTPS=green, Ping=info, TCP=warning, SMTP=red)

## Visual improvements
- Response time bars now show variable height based on response time duration
- Bottom status row indicates up/down state with color-coded indicators
- Each bar has individual gray background container with rounded corners
- Consistent chip styling across Status and Type columns in the table
- Subtle gray gradient on Up/Down/Paused/Initializing status cards

### DIFF
--- a/.github/workflows/check-format.yml
+++ b/.github/workflows/check-format.yml
@@ -8,6 +8,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
 
       - name: Set up Node
         uses: actions/setup-node@v4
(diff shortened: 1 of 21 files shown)


ajhollid (changes requested):
This appears to have broken some things:

image


image

image


image

Heatmap doesn't display properly anymore and switching between histogram and heamtap doesn't work :thinking: 

How about we break this PR up into more focused ones so we can track down the bugs?

1.  One PR for the replacement of button group
2.  One PR for adjustments to heatmap/histogram

And I don't think there should be any backend changes in this PR at all so we can drop the server changes

Issue #3120, opened 2026-01-10
Monitor IP/URL on Status Page is discoverable from page even when disabled


web-engineer:
**Describe the bug**

Admin page clearly states -

Monitor IP/URL on Status Page

Display the IP address or URL of monitor on the public Status page. If it's disabled, only the monitor name will be shown to protect sensitive information.

However with this option disabled the sensitive information is still in the JSON data the page reads. This should be filtered.

**To Reproduce**
Steps to reproduce -
1. Go to /settings
2. Set Monitor IP/URL on Status Page to disabled
3. Visit a status page
4. Open inspector and view the downloaded JSON for the page, data contains sensitive URLS

**Expected behavior**

The URLS **should not be in the data** they are not needed to render the report and may represent sensitive data about how someones infrastructure is arranged.

**Example**

```json
 {
                "_id": "692b1ecdba5dbe7b7ef93e13",
                "userId": "692b1e73ba5dbe7b7ef93e0b",
                "teamId": "692b1e73ba5dbe7b7ef93e09",
                "name": "Web-Engineer",
                "description": "Web-Engineer",
                "type": "http",
                "ignoreTlsErrors": false,
                "url": "https://www.web-engineer.co.uk",
                "isActive": true,
                "interval": 60000,
                "notifications": [
                    "692b25faba5dbe7b7ef94705"
                ],
```

In the above url **should not be present** if you have decided not to display this publicly,  I would have assumed that the user would not expect this to be easily found.

web-engineer:
PR made on [#2121](https://github.com/bluewave-labs/Checkmate/pull/3121)

web-engineer:
From my tests I don't think the settings are properly saved - they seem to only effect the current session when applied which maybe intentional but the Ui doesn't make this clear - just updating my patch for front and backend - to persist the data properly and enforce the rules on the data.

If you open a private browser the settings do not effect that browser when you toggle this setting.

web-engineer:
as far as I'm aware the PR I sent in has been merged

PR #3121, opened 2026-01-10
3120 remove URL's from data when they are supposed to be disabled

web-engineer:
## Fix: Hide monitor URLs from public status page when setting is disabled

**Issue:** #3120 Data leak: Monitor URLs exposed on public status page

### What changed
- Modified `StatusPageModule` to respect the global "Display IP/URL on status page" setting
- Monitor URLs are now conditionally removed from public API responses based on `AppSettings.showURL`
- Refactored to use dependency injection for `AppSettings` (consistent with existing patterns)

### Files changed
- `server/src/db/v1/modules/statusPageModule.js` – Added conditional URL stripping in response assembly
- `server/src/config/services.js` – Inject `AppSettings` into `StatusPageModule`

### Testing
- Public status page API respects the admin setting
- Disabled: URLs hidden, only monitor names visible
- Enabled: Full details shown as before
- No schema changes or migrations

## Summary by CodeRabbit

* **New Features**
  * Added optional "Show URL" setting to application settings
  * Settings save now displays a success confirmation message

* **Improvements**
  * Status pages now hide sensitive monitor details (URLs, ports, secrets, notifications) by default unless "Show URL" is enabled


web-engineer:
Tested in our cluster and verified this looks to be working as expected - had to change the docker/prod/server.Dockerfile entrypoint  to build the image 

```
CMD ["node", "dist/index.js"]
```

then I could build and push to our registry, manually patched the live service to use the new image and verified the status pages worked and that the URL's were omitted from the data.

Hope this works for you guys - had to rush this out as we want to use this status page for our live services but didn't want to expose our endpoints.

web-engineer:
This relates to the issue i opened #3120

web-engineer:
Jus to update, this PR is no good currently - I've spotted an issue with it and been working the problem today, hope to have an update shorlty.

web-engineer:
OK - Had to rethink this a bit, turned out the setting was not being saved - in the current release it is  only for the current user session, so doesn't do what it says anyway!

I've updated this so that this now saves to the database so the settings do now persist and corrected the filtering of the data.

Summary of changes -

## Implementation Details

### 1. Database Schema (Backend)
**File:** `server/src/db/v1/models/AppSettings.js`

Added `showURL` field to the Mongoose AppSettings schema:
```javascript
showURL: {
    type: Boolean,
    default: false,
}
```

This ensures the setting can be persisted to MongoDB and prevents Mongoose from silently ignoring the field (strict mode).

### 2. Server Validation (Backend)
**File:** `server/src/validation/joi.js`

Added `showURL` to the server-side validation schema:
```javascript
const updateAppSettingsBodyValidation = joi.object({
    // ... other fields
    showURL: joi.bool().optional(),
    // ... other fields
});
```

This validates incoming requests and allows the showURL field to be accepted by the API.

### 3. Status Page Data Filtering (Backend)
**File:** `server/src/db/v1/modules/statusPageModule.js`

Modified `getStatusPage` method to conditionally strip sensitive fields:
```javascript
const appSettings = await this.AppSettings.findOne({ singleton: true }).lean();
const showURL = appSettings?.showURL === true;

const normalizedMonitors = monitors.map((monitor) => {
    const normalizedChecks = this.NormalizeData(monitor.checks, 10, 100);
    if (showURL !== true) {
        const { url, port, secret, notifications, ...rest } = monitor;
        return { ...rest, checks: normalizedChecks };
    }
    return { ...monitor, checks: normalizedChecks };
});
```

When `showURL` is `false` or undefined, the following fields are removed from each monitor object:
- `url` - Monitor endpoint URL
- `port` - Monitor port number
- `secret` - Secret tokens/keys
- `notifications` - Notification configuration

### 4. Client Validation (Frontend)
**File:** `client/src/Validation/validation.js`

Added `showURL` to the client-side validation schema:
```javascript
const settingsValidation = joi.object({
    // ... other fields
    showURL: joi.boolean().optional(),
    // ... other fields
});
```

Without this, Joi validation would reject the settings object as having an unknown field, preventing the save button from working.

### 5. Settings Form Handler (Frontend)
**File:** `client/src/Pages/v1/Settings/index.jsx`

Added proper boolean coercion for the `showURL` field:
```javascript
// Ensure showURL is a proper boolean
if (name === "showURL") {
    newValue = value === true || value === "true";
}
```

This handles potential type coercion issues from MUI Select components and ensures the value is always a proper boolean before saving.

web-engineer:
thats me done for now - hope this is good for you guys, it works well for me when testing in our cluster.

web-engineer:
updated validation as per recomendation.

Regarding the nitpick about Redux being updated before persistence - this is there for immediate UI feedback. The showURL toggle updates Redux immediately so users see the change reflected in the UI right away, then persists when they click Save. This is different from timezone/mode/language which save immediately, but matches the Save button pattern used by other settings on the page.

ajhollid:
Hey there @web-engineer ,

Thanks for the PR!  Can you please remove package-lock.json from your PR?  We'll get to reviewing as soon as we can :+1:

web-engineer:
I noticed the package-lock.json was automatically updated when I worked on this PR. Looking at the git history, it's been tracked in the repo since 2024, and the changes in my commit are just dependency resolution updates (Babel packages being bumped from 7.27.x to 7.28.x) that happened when I ran npm install.

I can remove the package-lock changes from this commit if that's preferred, but wanted to check - what's the team's preference here?

Generally package-lock.json should be tracked for reproducible builds and dependency consistency, but I understand if you want these changes in a separate commit to keep this PR focused on just the Settings page fix. Let me know what you'd like me to do.

web-engineer:
lmk if you want me to pull latest from you and resolve / resubmit - hopefully this is a trivial merge.

ajhollid:
Yep that would be ideal

web-engineer:
there you go, reverted the package lock to where it was so this should now auto-merge

server/src/db/modules/statusPageModule.js
```diff
@@ -260,11 +260,16 @@ class StatusPageModule {
...
+			const appSettings = await this.AppSettings.findOne({ singleton: true }).lean();
+			const showURL = appSettings?.showURL === true;
+
 			const normalizedMonitors = monitors.map((monitor) => {
-				return {
-					...monitor,
-					checks: this.NormalizeData(monitor.checks, 10, 100),
-				};
+				const normalizedChecks = this.NormalizeData(monitor.checks, 10, 100);
+				if (showURL !== true) {
```
Br0wnHammer: Why the redundant check?
Line 264 already defines `showURL` as a `boolean`.
In the if block we can simply check `if (!showURL)`

client/src/Pages/Settings/index.jsx
```diff
@@ -147,29 +148,22 @@ const Settings = () => {
...
 	};
 
 	const handleSave = () => {
+		// Validate
 		const { error } = settingsValidation.validate(settingsData.settings, {
 			abortEarly: false,
 		});
 		if (!error || error.details.length === 0) {
 			setErrors({});
+			saveSettings(settingsData?.settings);
```
Br0wnHammer, line 166: Nice catch!

client/src/Pages/Settings/index.jsx
```diff
@@ -90,6 +85,12 @@ const Settings = () => {
 		) {
 			newValue = checked;
 		}
+
+		// Ensure showURL is a proper boolean
+		if (name === "showURL") {
```
Br0wnHammer, line 90: Nit: This is getting into if-statement hell territory, could be cleaner with a lookup map pattern. Happy to discuss, or we can mark this for a future cleanup.

Br0wnHammer (changes requested):
Solid fix for the URL leak! Left a minor nit on code structure, but otherwise LGTM.

web-engineer:
I'm actually just testing this as tried to address some of the other things the rabbit picked up on, validating the changes work against our install before updating this PR - I'll go over your comments too once done. Apologies for the mass of commits, I'm using another remote to push to our CI so I can run tests ...

web-engineer:
@Br0wnHammer took your point onboard re condition, tested on our local stack and merge looks happy to me. I did try implement some of the coderabbit suggestions but that opened a can of worms so reset back to the initial merge as they didn't seem practical right now.

client/src/Pages/Settings/index.jsx
web-engineer: happy to assist in the future where I can, I've some other ideas I'd like to float with you guys - happy to contribute ideas/fixes but my time is a bit limited - I needed this data-leak plugged as we wanted to use the data publicly on our site, hence the PR.

Br0wnHammer (approved):
LGTM!

client/src/Pages/Settings/index.jsx
web-engineer: This is using my fork - https://www.web-engineer.co.uk/status created a custom status page from the same JSON the public status page uses, I think there is milage in thinking about static simplified data such as a general overall status value for users who want to get an instant low impact status on their site - currently we have to pull the whole thing - anyway - its not that bigger overhead but I'd be happy to look at this and providing some sightly reduced responses and or greater control over what information an enterprise may want to leak to the public - i.e. you may want to be open about general health but more sparse with some details - particularly if it gives those trying to test for impacts on your application... ;) loving this product though and it's been fun to work with :) I'll try get some discussions going in the forums with some ideas when I get a chance.

PR #3136, opened 2026-01-16
fix: various UI/UX improvements for infrastructure pages

gorkem-bwl:
## Summary
- Fixed infrastructure monitor configure page not loading due to missing `dateRange` parameter
- Fixed URL validation error blocking edits on existing infrastructure monitors
- Fixed monitor update request using wrong property name (`_id` → `id`)
- Improved gauge box labels: "Device" (capitalized) and "Mounted on" instead of "mountpoint"
- Show full device path (e.g., `/dev/vda1`) instead of truncated version
- Implemented CSS Grid layout for gauge boxes ensuring consistent widths across all rows

## Test plan
- [ ] Navigate to infrastructure monitor configure page and verify it loads without errors
- [ ] Edit an existing infrastructure monitor and save changes successfully
- [ ] Check gauge boxes on infrastructure details page display correct labels
- [ ] Verify gauge boxes maintain consistent widths when multiple rows are displayed
- [ ] Resize browser window and confirm gauge boxes respond correctly

## Summary by CodeRabbit

* **New Features**
  * Responsive metric layout that adapts grid sizing based on the number of metrics displayed.

* **Improvements**
  * Device and mountpoint names now display in full without truncation for better visibility.
  * Updated metric label naming for improved clarity.
  * Date range filtering now includes sensible defaults for convenience.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

PR #3140, opened 2026-01-16
fix: make dateRange optional for hardware details endpoint

gorkem-bwl:
## Summary
- Make `dateRange` query parameter optional for `/monitors/hardware/details/{id}` endpoint
- Default to `"recent"` when not provided

## Problem
The Joi validation schema marks `dateRange` as optional, but the controller used `requireString()` which throws a 400 error when `dateRange` is missing. This caused the infrastructure configure page (`/infrastructure/configure/{id}`) to fail with:

```json
{"success":false,"msg":"dateRange is required","data":null}
```

## Solution
Changed from `requireString()` to `optionalString()` with `"recent"` as the default value, matching the Joi validation schema and the behavior of similar endpoints.

## Summary by CodeRabbit

* **Bug Fixes**
  * Hardware details retrieval no longer requires specifying a date range; the system now defaults to displaying recent data when a date range is not provided.

### DIFF
--- a/server/src/controllers/monitorController.ts
+++ b/server/src/controllers/monitorController.ts
@@ -97,7 +97,7 @@ class MonitorController {
 			await getHardwareDetailsByIdQueryValidation.validateAsync(req.query);
 
 			const monitorId = requireString(req?.params?.monitorId, "Monitor ID");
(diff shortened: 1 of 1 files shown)


gorkem-bwl:
Closing - fix pushed to PR #3136 instead

PR #3144, opened 2026-01-16
Fix incidents page issues

gorkem-bwl:
## Summary
- Fix incidents "Avg. Resolution Time" displaying decimal hours instead of human-readable format
- Fix incidents monitor filter dropdown showing no results when selecting a specific monitor

### Incidents resolution time format
Added `formatResolutionTime()` function that converts decimal hours to human-readable format:
- "0.86 hours" → "52m"
- "2.25 hours" → "2h 15m"
- Shows "N/A" when there are no incidents

### Incidents monitor filter fix
Fixed a race condition where selecting a monitor from the dropdown would fetch with the wrong page number. Consolidated page reset and fetch logic into a single useEffect to ensure page 0 is used immediately when filters change.

---

**Note:** The PageSpeed stats fix was removed from this PR as it has already been fixed differently on develop by Alex Holliday (commit 6447c315c).

## Summary by CodeRabbit

* **Bug Fixes**
  - Pagination now only resets to page 1 when filters or date ranges actually change.
  - Status boxes for PageSpeed now show updated "checks since" and last-check timings derived from check history.

* **Refactor**
  - Average resolution time is shown in a human-readable format (e.g., "2h 15m" instead of decimal hours).

### DIFF
--- a/client/src/Pages/Incidents2/Components/StatisticsPanel/index.jsx
+++ b/client/src/Pages/Incidents2/Components/StatisticsPanel/index.jsx
@@ -26,6 +26,18 @@ import SummaryCard from "../SummaryCard/index.jsx";
  * @param {boolean} props.isLoading - Loading state
  * @param {Object} props.error - Error object if any
  */
+const formatResolutionTime = (hours) => {
+	if (!hours || hours === 0) return "0m";
+
+	const totalMinutes = Math.round(hours * 60);
+	const h = Math.floor(totalMinutes / 60);
+	const m = totalMinutes % 60;
+
+	if (h === 0) return `${m}m`;
+	if (m === 0) return `${h}h`;
+	return `${h}h ${m}m`;
+};
+
 const StatisticsPanel = ({ isLoading = false, error = null, summary = {} }) => {
 	const theme = useTheme();
 	const { t } = useTranslation();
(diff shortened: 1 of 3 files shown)


client/src/Pages/Incidents2/index.jsx
```diff
@@ -53,16 +53,33 @@ const Incidents2 = () => {
 
 	const theme = useTheme();
 
-	useEffect(() => {
-		setPage(0);
-	}, [selectedMonitor, filter, dateRange]);
+	// Track previous filter values to detect changes and reset page
+	const prevFiltersRef = useRef({ selectedMonitor, filter, dateRange });
```
Br0wnHammer, line 57: We don't need two separate `useEffect` hooks or the `ref` tracking here.
Just merge them into one: fetch with page: 0 when filters change, then call setPage(0) at the end. Remove page from the dependencies.

Updated code:

```
useEffect(() => {
    const config = {
        monitorId: selectedMonitor !== "0" ? selectedMonitor : undefined,
        sortOrder: "desc",
        dateRange,
        page: 0,
        rowsPerPage,
    };
    
    if (filter === "active") {
        fetchActiveIncidents(config);
    } else if (filter === "resolved") {
        fetchResolvedIncidents(config);
    } else {
        fetchIncidents(config);
    }
    
    setPage(0);
}, [selectedMonitor, filter, dateRange, rowsPerPage, updateTrigger]);
```

client/src/Pages/Incidents2/Components/StatisticsPanel/index.jsx
```diff
@@ -26,6 +26,18 @@ import SummaryCard from "../SummaryCard/index.jsx";
  * @param {boolean} props.isLoading - Loading state
  * @param {Object} props.error - Error object if any
  */
+const formatResolutionTime = (hours) => {
```
Br0wnHammer, line 29: Not sure if there's already a utility function for this elsewhere in the codebase?

Br0wnHammer (changes requested):
Looks fine for most part. See my comments.

ajhollid:
I believe these have all been resolved

PR #3147, opened 2026-01-17
Fix sidebar gap when scrolling on pages with tall content

gorkem-bwl:
## Summary
- Fixed the sidebar showing a visible gap at the bottom when scrolling on pages with tall tables (e.g., /checks page)
- Changed sidebar positioning from sticky to fixed to avoid issues with parent container's overflow-x: hidden
- Added a spacer element to maintain proper layout spacing for the content area
- Disabled overscroll bounce effect and set black background on html/body

## Changes
- `Sidebar/index.jsx`: Changed from `position: sticky` to `position: fixed`, added background and border styles
- `HomeLayout/index.jsx`: Added spacer Box element that responds to sidebar collapsed state  
- `HomeLayout/index.css`: Removed unused selectors and cleaned up CSS
- `index.css`: Added `background-color: #000000` and `overscroll-behavior: none` to html/body

## Summary by CodeRabbit

* **New Features**
  * Sidebar now stays fixed while scrolling.
  * Layout adds a responsive spacer so content shifts when the sidebar is collapsed or expanded.

* **Style**
  * Updated sidebar background color, border, and stacking order for consistent appearance.
  * Enforced a dark page background and adjusted home layout spacing to match the new structure.

### DIFF
--- a/client/src/Components/v1/Layouts/HomeLayout/index.css
+++ b/client/src/Components/v1/Layouts/HomeLayout/index.css
@@ -14,15 +14,7 @@
 	}
 } */
 
-/* .home-layout aside {
-	position: sticky;
-	top: 0;
-	left: 0;
-	height: 100vh;
-	max-width: var(--env-var-side-bar-width);
-} */
-
-.home-layout > div {
+.home-layout > .home-content-wrapper {
 	min-height: calc(100vh - var(--env-var-spacing-2) * 2);
 	flex: 1;
 }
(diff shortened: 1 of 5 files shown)


ajhollid (changes requested):
The notes about duplicated logic for handling spacing is correct, this should only be set in one place and referenced as needed, or they will eventually become out of sync

gorkem-bwl:
> The notes about duplicated logic for handling spacing is correct, this should only be set in one place and referenced as needed, or they will eventually become out of sync

Can you check again?

Br0wnHammer (approved):
LGTM!

PR #3180, opened 2026-01-23
fix: prevent advanced matching checkbox from resetting on input

gorkem-bwl:
## Summary
- Fix issue where the advanced matching checkbox would uncheck when typing in the expected value field without first changing the match method dropdown

## Changes
- `client/src/Pages/Uptime/Create/index.jsx`: Add ref to track initialization state and only sync advanced matching checkbox once when monitor data is first loaded


ajhollid:
This issue has been resolved in another PR

PR #3232, opened 2026-02-03
fix: use correct route param in deleteStatusPage

### LINKED ISSUE #3228: Bug: DELETE /status-page/:url route param mismatch causes 400 error
## Bug Description

The DELETE status page endpoint doesn't work - it always returns `{"status":400,"msg":"Status page ID is required"}` regardless of how the request is made.

## Root Cause

There's a mismatch between the route parameter name and what the controller expects:

**Route definition** (`server/src/routes/statusPageRoute.ts`):
```typescript
this.router.delete("/:url(*)", verifyJWT, this.statusPageController.deleteStatusPage);
```

**Controller** (`server/src/controllers/statusPageController.ts`):
```typescript
const statusPageId = req.params.id as string;  // ❌ expects 'id'
if (!statusPageId) {
    throw new AppError({ message: "Status page ID is required", status: 400 });
}
```

The route defines the parameter as `:url` but the controller reads `req.params.id`, which will always be `undefined`.

## Steps to Reproduce

1. Create a status page with URL slug "test-page"
2. Attempt to delete it:
```bash
curl -X DELETE "http://localhost:52345/api/v1/status-page/test-page" \
  -H "Authorization: Bearer $TOKEN"
```
3. Response: `{"status":400,"msg":"Status page ID is required"}`

## Expected Behavior

The status page should be deleted successfully.

## Suggested Fix

Either:
1. Change the route to use `:id(*)` instead of `:url(*)`
2. Or change the controller to use `req.params.url`

The OpenAPI spec (`server/openapi.json`) documents this as taking the URL slug, so option 2 would match the documented API:
```json
{
  "name": "url",
  "in": "path",
  "required": true,
  "description": "Custom URL of the status page"
}
```

## Environment

- Checkmate version: v3.1.7 (develop branch, commit 2e7aba788)
- Deployment: Docker (dist config)

niveshdandyan:
## Summary

Fixes #3228

The `DELETE /status-page/:url` route defines the parameter as `:url(*)`, but the `deleteStatusPage` controller method was incorrectly reading `req.params.id` instead of `req.params.url`, causing a 400 error when trying to delete status pages.

## Changes

- Changed `req.params.id` to `req.params.url` in the `deleteStatusPage` method
- Renamed variable from `statusPageId` to `statusPageUrl` for clarity
- Updated error message to reference "URL" instead of "ID"

## Root Cause

This regression was introduced in PR #3112 during the JavaScript to TypeScript migration.

## Test Plan

1. Navigate to a status page
2. Click delete
3. Verify the status page is deleted successfully (no 400 error)

### DIFF
--- a/server/src/controllers/statusPageController.ts
+++ b/server/src/controllers/statusPageController.ts
@@ -116,12 +116,12 @@ class StatusPageController {
 
 	deleteStatusPage = async (req: Request, res: Response, next: NextFunction) => {
 		try {
-			const statusPageId = req.params.id as string;
-			if (!statusPageId) {
-				throw new AppError({ message: "Status page ID is required", status: 400 });
+			const statusPageUrl = req.params.url as string;
+			if (!statusPageUrl) {
+				throw new AppError({ message: "Status page URL is required", status: 400 });
 			}
 			const teamId = requireTeamId(req.user?.teamId);
-			await this.statusPageService.deleteStatusPage(statusPageId, teamId);
+			await this.statusPageService.deleteStatusPage(statusPageUrl, teamId);
 			return res.status(200).json({
 				success: true,
 				msg: "Status page deleted successfully",


ajhollid:
Hi @niveshdandyan, the change to deleteById is in fact the correct change and the route should be updated to reflect that.

This is actually resolved in an upcoming PR, so I'll close this one as the fix has already been implemented.

Thanks for the help here, though!

PR #3256, opened 2026-02-09
Fix Server URL field to be editable

### LINKED ISSUE #3100: Infrastructure Server URL impossible to edit
**Describe the bug**
The field `Server URL` in `Infrastructure` entries can not be edited. This makes it impossible to update entries to reflect servers who got a new IP or domain

**To Reproduce**
1. Go to 'Infrastructure'
2. Click on 'any entry'
3. Scroll down to 'Server URL'
4. Try to edit the field
5. See error mouse cursor

**Expected behavior**
I would expect not to have to recreate a monitor and lose all historic data over a changed IP

 - Firefox
 - 146.0.1

Aryan1718:
## Describe your changes

Fixes inability to edit the "Server URL" field.

Fixes #3100

```Javascript
<div>{t('add')}</div>
```
Screenshot 2026-02-08 at 6 28 14 PM


client/src/Pages/CreateMonitor/index.tsx
```diff
@@ -313,13 +314,17 @@ const CreateMonitorPage = () => {
...
 										fullWidth
-										disabled={isEditMode}
 										error={!!fieldState.error}
 										helperText={fieldState.error?.message ?? ""}
 									/>
 								)}
 							/>
 						)}
+						{isEditMode &&
+							generalSettingsConfig.showUrl &&
```
ajhollid: You can remove the `showUrl` condition, that is not used on the the config page.

client/src/Pages/CreateMonitor/index.tsx
```diff
@@ -313,13 +314,17 @@ const CreateMonitorPage = () => {
...
+							watchedUrl !== existingMonitor?.url && (
```
ajhollid: This causes the alert to show and then hide when the page loads, as the watchedUrl does not equal the existingMonitor.url initially but does once loaded.

client/src/Pages/CreateMonitor/index.tsx
```diff
@@ -313,13 +314,17 @@ const CreateMonitorPage = () => {
...
+								<Alert severity="warning">{t("urlUpdateWarning")}</Alert>
```
ajhollid: Pleaes put this string under `pages.createMonitor.form.general.url.alert` in `en.json`

ajhollid (changes requested):
Thanks for your contribution thusfar!

A couple of changes needed on the frontend, noted in review.

There is a bigger issue on the server side, which isn't caused by your PR but will require changes to your PR once resolved, so we may as well do it all here.
```
await editMonitorBodyValidation.validateAsync(req.body);
const editedMonitor = await this.monitorService.editMonitor({ teamId, monitorId, body: req.body });
```

The body is validated, but then the original unvalidated body is passed to the monitor service.

We should pass the validated body along:

```
const validatedBody = await editMonitorBodyValidation.validateAsync(req.body);
const editedMonitor = await this.monitorService.editMonitor({ teamId, monitorId, body: validatedBody });
```

And that will require updating the validation schema to allow the URL field, which is not currently allowed.

The reason the PR works currently is that the body is effectively not validated; if we did validate it properly, it would fail to update the URL as the value would be stripped.

We should be good to go after that 👍

Aryan1718:
Sure, @ajhollid. I will update it accordingly. Thank you.

Br0wnHammer:
@Aryan1718 any update?

Aryan1718:
Sorry about that @Br0wnHammer , I will do it.

client/src/Pages/CreateMonitor/index.tsx
Aryan1718: Updated. I removed the redundant showUrl condition from the alert logic.

client/src/Pages/CreateMonitor/index.tsx
Aryan1718: Fixed. The alert now only compares once existingMonitor is loaded, so it no
  longer flashes on initial page load.

client/src/Pages/CreateMonitor/index.tsx
Aryan1718: Updated. I moved the string to pages.createMonitor.form.general.url.alert in en.json and updated the component to use that key.

Aryan1718:
Hey @Br0wnHammer, please review it and let me know if anything needs to be changed.

Br0wnHammer (approved):
LGTM!

Br0wnHammer:
@Aryan1718 Can you please take care of the failing format check on client dir.

Aryan1718:
Hey @Br0wnHammer, I have updated the code to make sure the format check does not fail. Can you please review it?

Br0wnHammer (approved):
LGTM!

gorkem-bwl:
@Br0wnHammer can you check the conflicts here?

PR #3259, opened 2026-02-09
fix: restore isMaintenance field in status page API response

### LINKED ISSUE #3255: Status Page API no longer includes maintenance window status
**Describe the bug**
In v3.2, the API endpoint for status pages would include information about if a maintenance window existed for a monitor, with a property of `isMaintenance` under each monitor. After updating to v3.3, this field is no longer returned in the API response.

**To Reproduce**
1. Send a `GET` request to `https://checkmate-demo.bluewavelabs.ca/api/v1/status-page/179630?type=uptime`
2. Refer to `data.monitors[].isMaintenance`

**Expected behavior**
In v3.2, the field will exist as a boolean, indicating if a maintenance window has been created for this monitor. In v3.3, the field is missing.

 - Browser: cURL
 - Version: 8.18.0

scrollingreel:
## Describe your changes

Fixes the regression in v3.3 where `isMaintenance` field was missing from the status page API response (`/api/v1/status-page/:url`).

**Root cause:** The `getStatusPageByUrl` controller method was using `findByIds()` which performs a simple query, instead of `findByIdsWithChecks()` which runs the aggregation pipeline that calculates `isMaintenance` from maintenance windows.

**Changes:**
- Changed `findByIds` to `findByIdsWithChecks` in `statusPageController.ts`
- Added `isMaintenance?: boolean` property to `Monitor` type
- Included `isMaintenance` in `toEntityWithChecks` entity mapping

# Related Issue
Closes #3255

### DIFF
--- a/server/src/controllers/statusPageController.ts
+++ b/server/src/controllers/statusPageController.ts
@@ -81,7 +81,7 @@ class StatusPageController {
 			const settings = await this.settingsService.getDBSettings();
 			const showURL = settings.showURL;
 
-			const monitors = await this.monitorsRepository.findByIds(statusPage.monitors);
+			const monitors = await this.monitorsRepository.findByIdsWithChecks(statusPage.monitors);
 			// Sort monitors according to the order in statusPage.monitors
 			const monitorOrder = new Map(statusPage.monitors.map((id, index) => [id, index]));
 			const sortedMonitors = [...monitors].sort((a, b) => {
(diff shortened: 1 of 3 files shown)


Br0wnHammer (approved):
LGTM! Thanks for the fix 🚀

ajhollid:
Thanks for taking a look at this, however the change is intentional. The findByIdsWithChecks aggregation is too slow for use on the status page. Additionally, recent checks are now denormalized and included in the monitor itself, so the old aggregation is no longer needed here.

If the maintenance status is required, we'll have to implement a different method of handling that.  Please feel free to discuss further in the [relevant issue!](https://github.com/bluewave-labs/Checkmate/issues/3255)

PR #3279, opened 2026-02-13
enhancement: add average and max response time to histogram

### LINKED ISSUE #3263: Show the max response time in homepage
Could be great if we could see the max response time in the homepage graphs to have an idea of values in graph : 

Image

akashmannil:
**Summary**
Added average and max response time labels to the `HistogramResponseTime` component. The stats are displayed beside the graph (left or right side, configurable via statsPosition prop) to help users understand the scale of values shown in the visualization.

Changes
Added Avg: X ms and Max: X ms labels beside the response time graph
New optional props: `showStats` and `statsPosition`

image

image

Fixes #3263

- [ x ] (Do not skip this or your PR will be closed) I deployed the application locally.
- [ x ] (Do not skip this or your PR will be closed) I have performed a self-review and testing of my code.
- [ x ] I have included the issue # in the PR.
- [ x ] I have added i18n support to visible strings (instead of `<div>Add</div>`, use): 
```Javascript
<div>{t('add')}</div>
```
- [ x ] I have **not** included any files that are not related to my pull request, including package-lock and package-json if dependencies have not changed
- [ x ] I didn't use any hardcoded values (otherwise it will not scale, and will make it difficult to maintain consistency across the application).
- [ x ] I made sure font sizes, color choices etc are all referenced from the theme. I don't have any hardcoded dimensions.
- [ x ] My PR is granular and targeted to one specific feature.
- [ x ] I ran `npm run format` in server and client directories, which automatically formats your code.
- [ x ] I took a screenshot or a video and attached to this PR if there is a UI change.


akashmannil:
latest UI ss
image

client/src/Components/v2/common/charts/HistogramResponseTime.tsx
```diff
@@ -1,102 +1,181 @@
...
 const DEFAULT_HEIGHT = 50;
 
+const calculateResponseTimeStats = (checks: CheckSnapshot[]): ResponseTimeStats => {
+	if (!Array.isArray(checks) || checks.length === 0) {
+		return { max: 0, avg: 0 };
+	}
+
+	const validChecks = checks.filter(
+		(check) =>
+			(check as any).status !== "placeholder" &&
```
ajhollid: Let's avoid 'any' casts, there should never be a reason to cast to the `any` type really.  We lose all type safety if we do so.

client/src/Components/v2/common/charts/HistogramResponseTime.tsx
```diff
@@ -1,102 +1,181 @@
...
+			(check.originalResponseTime != null || check.responseTime != null)
+	);
+
+	if (validChecks.length === 0) {
+		return { max: 0, avg: 0 };
+	}
+
+	const responseTimes = validChecks.map(
+		(check) => check.originalResponseTime ?? check.responseTime ?? 0
+	);
```
ajhollid: We should only use `originalResponseTime` here.  `responseTime` is normalized and won't provide useful data for average or max response time.

Additionally, we should not provide a default value of 0, as that will skew the average response time downwards.  If response time is not present for some reason, the value should be omitted entirely.

TLDR we should filter the array and remove any items where `originalResponseTime` is not present.

client/src/Components/v2/common/charts/HistogramResponseTime.tsx
```diff
@@ -1,102 +1,181 @@
...
+
+	const statsContent = showStats && (stats.max > 0 || stats.avg > 0) && (
+		<Stack
+			justifyContent="center"
+			alignItems={statsPosition === "left" ? "flex-end" : "flex-start"}
+			sx={{
+				minWidth: 70,
+				pr: statsPosition === "left" ? theme.spacing(8) : 0,
+				pl: statsPosition === "right" ? theme.spacing(8) : 0,
+			}}
```
ajhollid: These properties can all be set on the `Stack` via exposed props, which should always be the priority.  `sx` is the next choice if the component does not expose the needed props.

client/src/Components/v2/common/charts/HistogramResponseTime.tsx
```diff
@@ -1,102 +1,181 @@
...
+				variant="caption"
 				sx={{
-					width: "100%",
-					display: "grid",
-					gridTemplateColumns: "repeat(25, 1fr)",
-					gap: gridGap,
-					alignItems: "end",
-					height: chartHeight,
+					color: theme.palette.text.secondary,
+					fontSize: "0.65rem",
```
ajhollid: Let's avoid custom font sizes, please. You can use the different variants like `body2` if you need a smaller font size.

client/src/Components/v2/common/charts/HistogramResponseTime.tsx
```diff
@@ -1,102 +1,181 @@
...
-								overflow: "hidden",
-							}}
-						>
+				{t("common.charts.histogram.avg", { value: stats.avg })}
+			</Typography>
+			<Typography
+				variant="caption"
+				sx={{
+					color: theme.palette.text.secondary,
+					fontSize: "0.65rem",
```
ajhollid: Same here

client/src/Components/v2/common/charts/HistogramResponseTime.tsx
```diff
@@ -1,102 +1,181 @@
...
+		</Stack>
+	);
+
+	return (
+		<Stack
+			direction="row"
+			alignItems="center"
+			sx={{ width: "100%" }}
+		>
+			{statsPosition === "left" && statsContent}
```
ajhollid: Rendering this conditionally on either side feels like a hack to me, what about changing the direction of the stack based on the statsPosition prop?  Toggling between `row` and `row-reverse` should achieve the desired effect 🤔

client/src/locales/en.json
```diff
@@ -231,6 +231,10 @@
 				"high": "High",
 				"low": "Low",
 				"uptime": "Uptime"
+			},
+			"histogram": {
+				"avg": "Avg: {{value}} ms",
+				"max": "Max: {{value}} ms"
```
ajhollid, line 53: Cool, good use of interpolation here 👍

ajhollid (changes requested):
Pretty good for the most part!

Some changes are needed around the response time data, and some minor adjustments here and there.

Thanks for the PR!

ajhollid (approved):
Looks good, thanks for revisions!

PR #3310, opened 2026-02-19
Fix: Authentication Middleware in StatusPage

Br0wnHammer:
## Describe your changes

This PR fixes a security vulnerability in the status page URL.

```Javascript
<div>{t('add')}</div>
```

Screenshot 2026-02-19 at 23 55 07


server/src/middleware/verifyStatusPageAccess.ts
```diff
@@ -0,0 +1,22 @@
...
+		try {
+			const url = req.params.url;
+			if (!url) {
+				throw new AppError({ message: "Status page URL is required", status: 400 });
+			}
+			const statusPage = await statusPagesRepository.findByUrl(url);
+			if (statusPage.isPublished) {
+				next(); // Published — proceed to controller (no JWT)
+			} else {
+				next("route"); // Unpublished — skip to next route (which has verifyJWT)
```
ajhollid: This is quite brittle; it depends on the route order never changing, which is not obvious.  Maintainers down the line won't know that this middleware depends on route order.

I think a more robust solution would be to have one route and just skip the JWT validation if the page is published

client/src/Utils/ApiClient.ts
```diff
@@ -61,7 +61,10 @@ export const initApiClient = (store: StoreType): void => {
 		}
 
 		if (error.response?.status === 401) {
-			if (window.location.pathname !== "/login") {
+			if (
+				window.location.pathname !== "/login" &&
+				!window.location.pathname.startsWith("/status/public")
```
ajhollid: This shouldn't be necessary I don't think 🤔   /status/public should never throw a 401 should it?  If the page isn't published we should get a 403, if it is published we should get a 200?

ajhollid (changes requested):
This solution depends on route order, which isn't immediately obvious if you didn't review this PR and understand how it was implemented.  It's not difficult to figure out, but we should try to avoid non-obvious solutions wherever possible, and I think we can do that here.

ajhollid (approved):
Thanks for revisions!

PR #3312, opened 2026-02-19
feat: add buffer lookup and location results update to services

gorkem-bwl:
## Summary
- Add `findInBuffer()` to BufferService for looking up checks still in the write buffer by ID
- Add `updateLocationResults()` to checks repository interface and MongoDB implementation
- These methods are not yet wired into the check pipeline — purely additive, no behavior changes

## Test plan
- [ ] `npm run build` in `/server` passes with zero errors
- [ ] Existing check pipeline is unaffected
- [ ] New methods are available but not called yet


Br0wnHammer (approved):
LGTM!

PR #3313, opened 2026-02-19
refactor: wire backfill pipeline and remove per-location Check documents

gorkem-bwl:
## Summary
- Switch from creating separate Check documents per Globalping location to backfilling `locationResults` into the same Check document as the local check
- Remove `location` from CheckMetadata, MonitorStatusResponse, metadata schema, location index, buildCheck(), toEntity(), and toDocument()
- Change GlobalpingService return type to `GlobalpingCheckResult[]` (simplified shape without monitorId/teamId/type)
- Rewrite SuperSimpleQueueHelper: after local check is buffered, fire Globalping checks and backfill results into the same document (in-buffer mutation if still buffered, DB updateOne if already flushed)
- Rewrite `findLocationChecksByMonitorId` to `$unwind` locationResults instead of querying by `metadata.location`
- Remove `metadata.location` filter from `findUptimeDateRangeChecks`

## Test plan
- [ ] `npm run build` in `/server` passes with zero errors
- [ ] Monitor without Globalping: checks work exactly as before, no `locationResults` field
- [ ] Monitor with Globalping enabled: local check writes immediately, then `locationResults` array gets backfilled after 5-15s
- [ ] Details page location tabs show per-location charts from `locationResults` via `$unwind`
- [ ] Backfill works both when check is still in buffer (mutated in-place) and when already flushed (DB `updateOne`)
- [ ] Only local check data drives monitor status — `locationResults` has no effect on status pipeline

### DIFF
--- a/server/src/repositories/checks/MongoChecksRepistory.ts
+++ b/server/src/repositories/checks/MongoChecksRepistory.ts
@@ -170,7 +170,6 @@ class MongoChecksRepository implements IChecksRepository {
 			monitorId: toStringId(metadata.monitorId),
 			teamId: toStringId(metadata.teamId),
 			type: metadata.type,
-			...(metadata.location ? { location: metadata.location } : {}),
 		});
 
 		return {
@@ -196,6 +195,7 @@ class MongoChecksRepository implements IChecksRepository {
 			seo: doc.seo,
 			performance: doc.performance,
 			audits: mapAudits(doc.audits),
+			locationResults: (doc as any).locationResults,
 			__v: doc.__v ?? 0,
 			createdAt: toDateString(doc.createdAt),
 			updatedAt: toDateString(doc.updatedAt),
@@ -221,7 +221,6 @@ class MongoChecksRepository implements IChecksRepository {
 				monitorId: new mongoose.Types.ObjectId(metadata.monitorId),
 				teamId: new mongoose.Types.ObjectId(metadata.teamId),
 				type: metadata.type,
-				...(metadata.location ? { location: metadata.location } : {}),
 			},
 			...rest,
 		} as unknown as CheckDocument;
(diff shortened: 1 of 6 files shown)


server/src/repositories/checks/MongoChecksRepistory.ts
```diff
@@ -429,24 +428,25 @@ class MongoChecksRepository implements IChecksRepository {
...
-		};
 
 		const results = await CheckModel.aggregate([
-			{ $match: matchStage },
-			{ $sort: { createdAt: 1 } },
+			{
+				$match: {
+					"metadata.monitorId": monitorObjectId,
+					createdAt: { $gte: startDate, $lte: endDate },
+					locationResults: { $exists: true, $not: { $size: 0 } },
```
Br0wnHammer: `Nit`: The $exists: true + $not: { $size: 0 } combo can behave unexpectedly if locationResults is null (it exists but isn't an array, so $size won't work as intended). Safer to use:
`locationResults: { $type: "array", $ne: [] }`
This guarantees we're working with a non-empty array in all cases.

server/src/repositories/checks/MongoChecksRepistory.ts
```diff
@@ -429,24 +428,25 @@ class MongoChecksRepository implements IChecksRepository {
...
+				},
+			},
+			{ $unwind: "$locationResults" },
```
Br0wnHammer, line 440: `Perf`: Since we're now doing `$unwind` on `locationResults`, it's important that the `$match` stage narrows the dataset as much as possible before that expansion happens.

server/src/service/infrastructure/SuperSimpleQueue/SuperSimpleQueueHelper.ts
```diff
@@ -127,21 +127,32 @@ class SuperSimpleQueueHelper {
...
+				if (monitor.globalpingEnabled && (monitor.type === "http" || monitor.type === "ping") && check.id) {
+					const checkId = check.id;
 					this.globalpingService
 						.runChecks(monitor)
-						.then((locationStatuses) => {
-							for (const locStatus of locationStatuses) {
-								const locCheck = this.checkService.buildCheck(locStatus);
-								if (locCheck) {
-									this.buffer.addToBuffer(locCheck);
+						.then((locationResults) => {
```
Br0wnHammer: The dual write path here (buffer mutation vs DB update) depends on whether the buffer has flushed by the time runChecks resolves. Can we add a comment explaining this is intentionally fire-and-forget? Also could a flush mid-callback cause a lost write?

Br0wnHammer: This `.then()` chain doesn't have a `.catch()`. If `runChecks` or `updateLocationResults` rejects, the error gets silently swallowed. Since this is fire-and-forget, we won't notice failures unless we explicitly log them.

Br0wnHammer (changes requested):
Please see my comments. @gorkem-bwl

PR #3343, opened 2026-02-27
feat: adding infrastructure display support to status page

### LINKED ISSUE #3305: [Feature] Show infrastructure metrics on public status pages
### Summary
Allow users to display infrastructure/hardware information on public status pages, giving end-users visibility into the systems behind their services.

### Problem
Currently, infrastructure monitors (servers) cannot be added to the public status page. Users want to show their customers basic hardware metrics without exposing the full private dashboard.

### Proposed Solution
Add option to include infrastructure monitors on the public status page
Display basic metrics such as:
Storage capacity/usage over time
CPU usage
RAM usage
Allow users to toggle which infrastructure components are publicly visible
Optionally link monitored services to their underlying server hardware

### Additional Context
The current terminology "server" in the app can be confusing - users expect it to refer to their hardware/infrastructure, not uptime monitors.

akashmannil:
These changes focus on adding Infrastructure section to currently existing status Page. 

Fixes #3305 

https://github.com/user-attachments/assets/6b67339b-6f55-40e8-929e-67e993e45b7c

Latest UI after adding a little bit more padding.
image

image

- [ x ] (Do not skip this or your PR will be closed) I deployed the application locally.
- [ x ] (Do not skip this or your PR will be closed) I have performed a self-review and testing of my code.
- [ x ] I have included the issue # in the PR.
- [ x ] I have added i18n support to visible strings (instead of `<div>Add</div>`, use): 
```Javascript
<div>{t('add')}</div>
```
- [ x ] I have **not** included any files that are not related to my pull request, including package-lock and package-json if dependencies have not changed
- [ x ] I didn't use any hardcoded values (otherwise it will not scale, and will make it difficult to maintain consistency across the application).
- [ x ] I made sure font sizes, color choices etc are all referenced from the theme. I don't have any hardcoded dimensions.
- [ x ] My PR is granular and targeted to one specific feature.
- [ x ] I ran `npm run format` in server and client directories, which automatically formats your code.
- [ x ] I took a screenshot or a video and attached to this PR if there is a UI change.


client/src/locales/en.json
```diff
@@ -1154,7 +1144,30 @@
...
+					"responseTime": "Response time"
+				},
+				"badge": {
+					"http": "HTTP",
+					"ping": "Ping",
+					"port": "Port",
+					"docker": "Docker",
+					"pagespeed": "PageSpeed",
+					"hardware": "Infrastructure",
+					"game": "Game"
```
ajhollid: I have a feeling these values are all declared somewhere else already, can you please thoroughly check the localization file?  Duplicated strings is a maintenance nightmare

client/src/Pages/StatusPage/Status/Components/InfrastructureMetrics.tsx
```diff
@@ -0,0 +1,152 @@
...
+						>
+							<Typography
+								variant="body2"
+								color="text.secondary"
+							>
+								{upperLabel}
+							</Typography>
+							<Typography variant="body2">{upperValue}</Typography>
+						</Stack>
+					)}
```
ajhollid: I believe we can better handle this nested conditional rendering.

The rendering condition for the inner stack is equivalent to

```
if (A or B){
  if (A) {
    // do A
  }
  if (B) {
    // do B
  }
}
```

At this level of complexity, it's probably time to define a component that internalizes this logic in a clear easy to understand way

client/src/Pages/StatusPage/Status/Components/InfrastructureMetrics.tsx
```diff
@@ -0,0 +1,152 @@
...
+				<Typography variant="body2">{label}</Typography>
+			</Stack>
+			{(upperLabel || lowerLabel) && (
+				<Stack
+					spacing={theme.spacing(1)}
+					flex={1}
+					paddingBottom={theme.spacing(10)}
+				>
+					{upperLabel && (
+						<Stack
```
ajhollid: This is quite deeply nested. We can probably flatten this layout considerably using grids

client/src/Pages/StatusPage/Status/Components/InfrastructureMetrics.tsx
```diff
@@ -0,0 +1,152 @@
...
+						</Stack>
+					)}
+					{lowerLabel && (
+						<Stack
+							direction="row"
+							justifyContent="space-between"
+						>
+							<Typography
+								variant="body2"
+								color="text.secondary"
```
ajhollid: We generally don't use this contention in the project, let's stick with referencing the palette for theme colors for consistency.

client/src/Pages/StatusPage/Status/Components/InfrastructureMetrics.tsx
```diff
@@ -0,0 +1,152 @@
...
+	}
+
+	const cpuUsage = (latestCheck?.cpu?.usage_percent || 0) * 100;
+	const memoryUsage = (latestCheck?.memory?.usage_percent || 0) * 100;
+	const memoryUsed = latestCheck?.memory?.used_bytes || 0;
+	const memoryTotal = latestCheck?.memory?.total_bytes || 0;
+
+	const disks = latestCheck?.disk ?? [];
+	const totalDiskUsage = disks.reduce((acc, disk) => acc + (disk?.usage_percent || 0), 0);
+	const diskCount = disks.length || 1;
```
ajhollid: This seems problematic, if disks is empty, ie no disks, this will evaluate to 1 🤔 

If there are no discs, there should be no disk metrics and we should return.

client/src/Pages/StatusPage/Status/Components/MonitorsList.tsx
```diff
@@ -85,22 +113,49 @@ export const MonitorsList = ({ statusPage, monitors }: MonitorsListProps) => {
...
+								padding={theme.spacing(4)}
+							>
+								<Stack
+									direction="row"
+									alignItems="center"
+									justifyContent="space-between"
+									gap={theme.spacing(8)}
+								></Stack>
+								<InfrastructureMetrics monitor={monitor} />
+							</BaseBox>
```
ajhollid: This rendering logic is also becoming deeply nested, probably time to extract here as well

client/src/Pages/StatusPage/Status/Components/MonitorsList.tsx
```diff
@@ -53,30 +55,56 @@ export const MonitorsList = ({ statusPage, monitors }: MonitorsListProps) => {
...
+										variant="caption"
+										sx={{
+											backgroundColor:
+												monitor?.type === "hardware"
+													? theme.palette.info.light
+													: theme.palette.success.light,
+											color:
+												monitor?.type === "hardware"
+													? theme.palette.info.dark
+													: theme.palette.success.dark,
```
ajhollid: Let's create helper functions for these, there may be more monitor types in the future and having this all inline ternaries will become messy fast.

Additionally, they are somewhat difficult to read 🤔  I don't think there's enough contrast between the bg and the text.  Probably a black/transparent bg with colored text would be more readable.

Image

client/src/Pages/StatusPage/Status/Components/MonitorsList.tsx
```diff
@@ -53,30 +55,56 @@ export const MonitorsList = ({ statusPage, monitors }: MonitorsListProps) => {
...
+											padding: `${theme.spacing(0.5)} ${theme.spacing(1.5)}`,
```
ajhollid: Why fractional spacings?

ajhollid (changes requested):
In general, it seems to work well.

My main concern is the abundance of complexity and conditional rendering. 

I think we need to spend some time simplifying this, extracting complex rendering logic, and cleaning up as much as possible.

Aside from that, we need to be extremely careful with the localization file.  If we have "http" declared for a type of monitor somewhere else ("optionHttp": "HTTP(S)") , we don't want to do it again for the badges.

If we do, we now have to make sure these always stay in sync, which is frustrating.  If the same string needs to be used in multiple places it should be extracted to the "common" key in en.json.

client/src/Pages/StatusPage/Status/Components/MonitorsList.tsx
```diff
@@ -85,22 +113,49 @@ export const MonitorsList = ({ statusPage, monitors }: MonitorsListProps) => {
...
+								key={monitor.id}
+								padding={theme.spacing(4)}
+							>
+								<Stack
+									direction="row"
+									alignItems="center"
+									justifyContent="space-between"
+									gap={theme.spacing(8)}
+								></Stack>
+								<InfrastructureMetrics monitor={monitor} />
```
ajhollid: What's going on here actually?  Why do we have a BaseBox inside another BaseBox, holding an empty stack?

A BaseBox should be a base component 🤔 If showing infra stats is set to false, lets just display the regular histogram then rather than an empty row.

Br0wnHammer:
@akashmannil there are some conflicts in the PR, can you please take care of it?

akashmannil:
image
image
image
updated views

PR #3350, opened 2026-03-02
fix: server validation

ajhollid:
Convert Joi validation to Zod validation


server/src/validation/shared.ts
```diff
@@ -0,0 +1,33 @@
...
+	.max(50, "Name must be less than 50 characters")
+	.regex(
+		/^(?=.*[\p{L}\p{Sc}])[\p{L}\p{Sc}\s'\-().]+$/u,
+		"Names must contain at least 1 letter and may only include letters, currency symbols, spaces, apostrophes, hyphens (-), periods (.), and parentheses ()."
+	);
+
+export const lowercaseEmailValidation = z
+	.email()
+	.transform((val) => val.toLowerCase())
+	.refine((val) => val === val.toLowerCase(), {
```
Br0wnHammer: The .transform() already lowercases the value, so the subsequent .refine() checking val === val.toLowerCase() will always pass. This is dead code.

server/src/validation/monitorValidation.ts
```diff
@@ -0,0 +1,123 @@
...
+		tempAlertThreshold: z.number().optional(),
+		gameId: z.union([z.string(), z.literal("")]).optional(),
+		grpcServiceName: z.union([z.string(), z.literal("")]).optional(),
+		selectedDisks: z.array(z.string()).optional(),
+		group: z.union([z.string().max(50).trim(), z.null(), z.literal("")]).optional(),
+		geoCheckEnabled: z.boolean().optional(),
+		geoCheckLocations: z.array(z.enum(GeoContinents)).optional(),
+		geoCheckInterval: z.number().min(300000).optional(),
+	})
+	.passthrough();
```
Br0wnHammer: The original Joi validation used .options({ stripUnknown: true }), which removed unknown fields from the validated output. The Zod .passthrough() does the opposite -- it keeps unknown fields.

Br0wnHammer (changes requested):
The global error handler in `server/src/middleware/handleErrors.ts` has no awareness of ZodError. When Zod's .parse() throws, it produces a ZodError with error.issues but no .status property. The handler will default to HTTP 500 instead of the expected 400 (bad request). Previously, Joi's ValidationError had error.isJoi = true and error.details, and the middleware apparently relied on the status defaulting from the catch blocks. But with Zod, a ZodError has error.name === "ZodError" and error.issues array.

server/src/validation/monitorValidation.ts
ajhollid: Good catch, it's deprecated too

PR #3369, opened 2026-03-04
fix: dynamically set Swagger UI server URL from request

### LINKED ISSUE #3339: Swagger UI point to localhost
**Actual situation**

Docker with FE+BE V3.4.0

When accessing to swagger ui (/api-docs) its configuration point to localhost avoiding usage and testing from remote host

**Expected Situation**

Swagger UI point to an user defined IP/FQDN.

**Workaround**

Edit and mount a customized openapi.json file on the docker container

```
"servers": [
                {
                        "url":"http://MYIPADDRESS:{PORT}/{API_PATH}",
                }
]
```

**Step to reproduce**
=> Get the docker compose file (with be+fe)
=> docker compose up after settings ENV Variable 
- UPTIME_APP_API_BASE_URL
- UPTIME_APP_CLIENT_HOST
- CLIENT_HOST
=> Access to http://yourip:52345/api-docs

Image

In this situation you can't try API endpoints through

gorkem-bwl:
## Summary
- Swagger UI now prepends the current server URL (derived from request protocol and host) to the servers list
- This makes Swagger work correctly in Docker and remote environments instead of always pointing to localhost

## Closes
Fixes #3339

## Test plan
- [ ] Access `/api-docs` locally and verify "Current Server" points to `http://localhost:52345/api/v1`
- [ ] Access `/api-docs` from a remote host and verify it shows the correct remote URL
- [ ] Verify original localhost and demo servers still appear in the dropdown

### DIFF
--- a/server/src/app.ts
+++ b/server/src/app.ts
@@ -75,7 +75,7 @@ export const createApp = ({
 		})
 	);
 	// Swagger UI — dynamically set server URL from request
-	app.use("/api-docs", swaggerUi.serve, (req, res, next) => {
+	app.use("/api-docs", swaggerUi.serve, (req: express.Request, res: express.Response, next: express.NextFunction) => {
 		const protocol = req.protocol;
 		const host = req.get("host");
 		const dynamicSpec = {
--- a/server/src/app.ts
+++ b/server/src/app.ts
@@ -74,8 +74,22 @@ export const createApp = ({
 			},
 		})
 	);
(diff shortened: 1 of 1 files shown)


ajhollid (approved):

ajhollid (changes requested):
Code is fine, formatting error needs to be resolved and this is good to go

ajhollid (approved):

PR #3370, opened 2026-03-04
fix: prevent advanced matching checkbox from resetting on form changes

### LINKED ISSUE #3093: Advanced Matching checkbox gets unchecked when typing in Expected Value field
https://github.com/user-attachments/assets/309844fa-681d-473c-95db-976406aa66a4

## Description
The "Advanced matching" checkbox in the Uptime Monitor editing page gets automatically unchecked when users type in the "Expected value" field. Additionally, after saving data and reloading the page, the checkbox remains unchecked even when `expectedValue` or `jsonPath` fields contain data.

## Steps to Reproduce
1. Navigate to Uptime Monitor creation/editing page (`/uptime/create` or `/uptime/configure/:id`)
2. Select monitor type as "http"
3. Check the "Advanced matching" checkbox
4. Start typing in the "Expected value" field
5. Observe that the checkbox gets unchecked immediately

**Alternative scenario:**
1. Create/edit a monitor with advanced matching enabled
2. Fill in `expectedValue` or `jsonPath` fields
3. Save the monitor
4. Reload the page
5. Observe that the "Advanced matching" checkbox is unchecked despite having data in the fields

## Expected Behavior
- The "Advanced matching" checkbox should remain checked when typing in the "Expected value" or "JSON Path" fields
- After saving and reloading, if `matchMethod`, `expectedValue`, or `jsonPath` contains data, the checkbox should be automatically checked

## Actual Behavior
- The checkbox gets unchecked immediately when typing in "Expected value" field
- After reloading, the checkbox remains unchecked even when advanced matching fields contain data

## Root Cause
The `useEffect` hook that manages the checkbox state was watching the entire `monitor` object, causing it to re-run on every keystroke. Additionally, it only checked for `monitor.matchMethod` existence, ignoring `expectedValue` and `jsonPath` fields.

## Proposed Solution
1. Update the `useEffect` dependency array to only watch `monitor.matchMethod`, `monitor.expectedValue`, and `monitor.jsonPath` instead of the entire `monitor` object
2. Modify the condition to check for any of the three fields (`matchMethod`, `expectedValue`, or `jsonPath`) instead of just `matchMethod`
3. Add logic in the `onChange` handler to automatically check the checkbox when users type in `expectedValue` or `jsonPath` fields

## Environment
- Checkmate version: 3.2.0
- Browser: [Any]
- OS: [Any]

## Additional Context
This issue affects the user experience when configuring HTTP monitors with advanced matching rules, as users need to manually re-check the checkbox every time they type in the expected value field.

gorkem-bwl:
## Summary
- Added `useRef` guard to prevent `form.reset()` from overwriting the advanced matching checkbox after initial load
- The checkbox no longer unchecks when typing in the expected value field

## Closes
Fixes #3093

## Test plan
- [ ] Create a new HTTP monitor, enable "Advanced matching", type in the expected value field — checkbox should stay checked
- [ ] Edit an existing monitor with advanced matching enabled — checkbox should remain checked on page load
- [ ] Save and reload — verify checkbox state persists correctly

### DIFF
--- a/client/src/Pages/CreateMonitor/index.tsx
+++ b/client/src/Pages/CreateMonitor/index.tsx
@@ -1,4 +1,4 @@
-import { useMemo, useState } from "react";
+import { useMemo, useRef, useState } from "react";
 import { useEffect } from "react";
 import { logger } from "@/Utils/logger";
 import { useParams, useLocation, useNavigate } from "react-router";
@@ -188,9 +188,13 @@ const CreateMonitorPage = () => {
 	});
 	const { control, watch, handleSubmit, clearErrors } = form;
 
+	const hasInitialized = useRef(false);
 	useEffect(() => {
-		form.reset(defaults);
-	}, [defaults, form]);
+		if (!hasInitialized.current && existingMonitor) {
+			form.reset(defaults);
+			hasInitialized.current = true;
+		}
+	}, [defaults, form, existingMonitor]);
 
 	const watchedType = watch("type") as MonitorType;
 


ajhollid:
Is this actually an issue?  I have not noticed this beinga problem anywhere.  The form should never be reset after initial load?

gorkem-bwl:
> Is this actually an issue? I have not noticed this beinga problem anywhere. The form should never be reset after initial load?

It was an issue in previous releases I swear :-D

PR #3371, opened 2026-03-04
Fix infrastructure threshold defaults via migration (#3327)

### LINKED ISSUE #3327: Upgrading to v3.4.0 sets infrastructure thresholds to 5%
**Describe the bug**
After upgrading to v3.4.0 thresholds are set to 5%, incidents are automatically created and server status is 'Breached'

**To Reproduce**
1. Have infrastructure monitoring available
2. Upgrade to 3.4.0
3. Check infrastructure configuration to see thresholds changed

**Expected behavior**
Thresholds should not be changed

**Screenshots**

Image

 - Browser: Any
 - Version: Any

**Additional context**

gorkem-bwl:
## Summary
- Adds migration `0004_fixInfrastructureThresholds` that corrects hardware monitors with threshold values of 5 (old default from counter fields) to 100 (disabled)
- Fixes #3327 where infrastructure monitors had incorrect alert thresholds due to schema defaults being applied from counter fields

## Changes
- New migration file: `server/src/db/migration/0004_fixInfrastructureThresholds.ts`
- Registered migration in `server/src/db/migration/index.ts`

## Test plan
- [ ] Verify migration runs on startup without errors
- [ ] Confirm hardware monitors with threshold=5 get updated to 100
- [ ] Verify monitors with correct thresholds are not modified

### DIFF
--- a/server/src/db/migration/0004_fixInfrastructureThresholds.ts
+++ b/server/src/db/migration/0004_fixInfrastructureThresholds.ts
@@ -1,0 +1,65 @@
+import { MonitorModel } from "../models/Monitor.js";
+import { logger } from "@/utils/logger.js";
+
+/**
+ * Fix infrastructure monitors that had thresholds incorrectly set to 5
+ * due to the old default value. Updates them to 100 (disabled).
+ */
+export async function fixInfrastructureThresholds(): Promise<void> {
+	const SERVICE_NAME = "Migration:FixInfrastructureThresholds";
+
+	try {
+		logger.info({ service: SERVICE_NAME, message: "Starting infrastructure threshold fix" });
+
+		const thresholdFields = [
+			"cpuAlertThreshold",
+			"memoryAlertThreshold",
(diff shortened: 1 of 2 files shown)


server/src/db/migration/0004_fixInfrastructureThresholds.ts
```diff
@@ -0,0 +1,65 @@
...
+		let updatedCount = 0;
+
+		for (const monitor of monitors) {
+			const update: Record<string, number> = {};
+
+			for (const field of thresholdFields) {
+				if ((monitor as any)[field] === 5) {
+					update[field] = 100;
+				}
+			}
```
ajhollid, line 43: If we're doing the migration, we should also migrate any "0" values, since those also caused issues

ajhollid (changes requested):
We should handle the 0 case as well.

There are formatting issues to resolve as well

ajhollid (approved):

PR #3373, opened 2026-03-04
Add HTTP basic authentication for monitors (#3033)

### LINKED ISSUE #3033: [FEATURE]: Advanced Settings Addition: HTTP Basic Auth
Hey there,
I decided to switch to Checkmate recently and am now creating a bunch of Feature request of stuff I'm missing/liked to have.

Is your feature request related to a problem? Please describe.
I'd love to have a HTTP Basic Auth implementation on the Checks. It's hopefully fairly simple to implement.
Extremely useful when services are behind an extra auth layer that is not dependent in activity from the actual service.

Describe the solution you'd like
Having an extendable option in the Advanced Settings where you can Specify a HTTP Basic Auth Username and Password that the check uses when checking.

Additional context
I was trying to implement it myself but I'm not familiar with REACT and your system so help would be appreciated.

Thanks for your help and feel free to take a look at my other Issues!
#3032, #3034, #3035

gorkem-bwl:
## Summary
- Adds HTTP basic authentication support for HTTP monitors
- Users can now configure username/password credentials that are sent as `Authorization: Basic` headers
- Fixes #3033

## Changes
- Server: Added `basicAuthUser` and `basicAuthPassword` fields to Monitor model, types, and validation
- Server: Updated `networkService.ts` to send Basic auth header when credentials are configured
- Client: Added fields to Monitor type, Zod validation, and form defaults
- Client: Added basic auth form section with username/password fields in the create/edit monitor page
- Client: Added i18n strings for the new section

## Test plan
- [ ] Create an HTTP monitor with basic auth credentials and verify it sends the correct Authorization header
- [ ] Verify monitors without basic auth still work normally
- [ ] Verify bearer token (secret) takes priority over basic auth when both are set
- [ ] Edit an existing monitor and add/remove basic auth credentials


server/src/types/monitor.ts
```diff
@@ -27,6 +27,8 @@ export interface Monitor {
 	jsonPath?: string;
 	expectedValue?: string;
 	matchMethod?: MonitorMatchMethod;
+	basicAuthUser?: string;
+	basicAuthPassword?: string;
```
ajhollid, line 31: The repository has not been updated to return this data, so it will never show as being set on the FE, nor will it be editable.

Additionally, we have to probably do something similar to PageSpeedAPI key, otherwise we will be sending sensitive data back and forth.  The password should never be sent from the backend to the frontend, which should always be assumed to be compromised

server/src/validation/monitorValidation.ts
```diff
@@ -70,6 +70,8 @@ export const createMonitorBodyValidation = z.object({
 	jsonPath: z.union([z.string(), z.literal("")]).optional(),
 	expectedValue: z.union([z.string(), z.literal("")]).optional(),
 	matchMethod: z.union([z.string(), z.null(), z.literal("")]).optional(),
+	basicAuthUser: z.union([z.string(), z.literal("")]).optional(),
+	basicAuthPassword: z.union([z.string(), z.literal("")]).optional(),
```
ajhollid, line 74: This is being written in plain text to the database, this should be encrypted if it's going to be stored

server/src/service/infrastructure/networkService.ts
```diff
@@ -310,8 +310,15 @@ class NetworkService implements INetworkService {
 			if (!url) {
 				throw new Error("Monitor URL is required");
 			}
+			const headers: Record<string, string> = {};
+			if (secret) {
+				headers["Authorization"] = `Bearer ${secret}`;
+			} else if (basicAuthUser) {
+				const credentials = Buffer.from(`${basicAuthUser}:${basicAuthPassword || ""}`).toString("base64");
+				headers["Authorization"] = `Basic ${credentials}`;
+			}
```
ajhollid, line 319: Are these intentionally mutually exclusive?

ajhollid (changes requested):

PR #3398, opened 2026-03-09
feat: retention

ajhollid:
This PR adds a job-based retention cleanup feature
- Add a retention cleanup job based on the old CheckTTL value (default 30 days)
- Add UI for setting CheckTTL
- Update validation

image

image

### DIFF
--- a/client/src/Pages/Settings/index.tsx
+++ b/client/src/Pages/Settings/index.tsx
@@ -462,6 +462,43 @@ export const SettingsPage = () => {
 					/>
 				)}
 
+				{/* Check Retention */}
+				{isAdmin && (
+					<ConfigBox
+						title={t("pages.settings.form.retention.title")}
+						subtitle={t("pages.settings.form.retention.description")}
+						rightContent={
+							<Controller
+								name="checkTTL"
+								control={form.control}
+								defaultValue={defaults.checkTTL}
+								render={({ field }) => (
+									<SliderWithLabel
+										{...field}
+										fieldLabel={t("pages.settings.form.retention.option.days.label")}
+										min={1}
+										max={366}
+										sliderMaxWidth={{ xs: "100%", md: "50%" }}
+										value={field.value || 30}
+										onChange={(_, value) => field.onChange(value)}
+										valueLabelDisplay="auto"
+										valueLabelFormat={(value: number) =>
+											value >= 366
+												? t("pages.settings.form.retention.option.days.unlimited")
+												: `${value}`
+										}
+										formatDisplayValue={(value: number) =>
+											value >= 366
+												? t("pages.settings.form.retention.option.days.unlimited")
+												: `${value}`
+										}
+									/>
+								)}
+							/>
+						}
+					/>
+				)}
+
 				{/* Global Thresholds */}
 				{isAdmin && (
 					<ConfigBox
--- a/client/src/Pages/Settings/index.tsx
+++ b/client/src/Pages/Settings/index.tsx
@@ -37,6 +37,7 @@ import {
 } from "@/Features/UI/uiSlice.js";
 import timezones from "@/Utils/timezones.json";
 import type { RootState } from "@/Types/state";
+import { CHECK_TTL_SENTINEL } from "@/Types/Check";
 
 interface Timezone {
 	id: string;
@@ -477,18 +478,18 @@ export const SettingsPage = () => {
 										{...field}
 										fieldLabel={t("pages.settings.form.retention.option.days.label")}
 										min={1}
-										max={366}
+										max={CHECK_TTL_SENTINEL}
 										sliderMaxWidth={{ xs: "100%", md: "50%" }}
 										value={field.value || 30}
 										onChange={(_, value) => field.onChange(value)}
 										valueLabelDisplay="auto"
 										valueLabelFormat={(value: number) =>
-											value >= 366
+											value >= CHECK_TTL_SENTINEL
 												? t("pages.settings.form.retention.option.days.unlimited")
 												: `${value}`
 										}
 										formatDisplayValue={(value: number) =>
-											value >= 366
+											value >= CHECK_TTL_SENTINEL
 												? t("pages.settings.form.retention.option.days.unlimited")
 												: `${value}`
 										}
--- a/server/src/types/check.ts
+++ b/server/src/types/check.ts
@@ -1,5 +1,8 @@
 import type { MonitorType } from "@/types/index.js";
 import type { Response } from "got";
+
+export const CHECK_TTL_SENTINEL = 366;
+
 export type GotTimings = Response["timings"];
 
 export interface CheckMetadata {
--- a/server/src/types/check.ts
+++ b/server/src/types/check.ts
@@ -98,8 +98,6 @@ export interface Check {
 	timings?: GotTimings;
 	statusCode: number;
 	message: string;
-	ack: boolean;
-	ackAt?: string | null;
 	expiry: string;
 	cpu?: CheckCpuInfo;
 	memory?: CheckMemoryInfo;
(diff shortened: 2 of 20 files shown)


server/src/types/check.ts
```diff
@@ -98,8 +98,6 @@ export interface Check {
 	timings?: GotTimings;
 	statusCode: number;
 	message: string;
-	ack: boolean;
```
Br0wnHammer, line 101: The server removed `ack`, `ackAt`, and `__v` from the Check type and Mongoose schema, but the client-side type still has them. The client `CheckSnapshot` type also still references the removed fields.

server/src/validation/settingsValidation.ts
```diff
@@ -6,7 +6,7 @@ import { z } from "zod";
 
 export const updateAppSettingsBodyValidation = z
 	.object({
-		checkTTL: z.union([z.number(), z.literal("")]).optional(),
+		checkTTL: z.union([z.number().int().min(1).max(366), z.literal("")]).optional(),
```
Br0wnHammer: The sentinel value 366 meaning "unlimited" is scattered across 5+ locations with no named constant.
If anyone changes this in one place but not the others, the "unlimited" feature silently breaks. Consider extracting a shared constant like `CHECK_TTL_UNLIMITED = 366`

client/src/Pages/Settings/index.tsx
```diff
@@ -462,6 +462,43 @@ export const SettingsPage = () => {
...
+									<SliderWithLabel
+										{...field}
+										fieldLabel={t("pages.settings.form.retention.option.days.label")}
+										min={1}
+										max={366}
+										sliderMaxWidth={{ xs: "100%", md: "50%" }}
+										value={field.value || 30}
+										onChange={(_, value) => field.onChange(value)}
+										valueLabelDisplay="auto"
+										valueLabelFormat={(value: number) =>
```
Br0wnHammer, line 486: `valueLabelFormat` and `formatDisplayValue` callbacks are identical. Consider extracting to a single function variable to reduce duplication.

Br0wnHammer (changes requested):
Clean PR as always, minor changes are needed. Please see my comments.

gorkem-bwl (changes requested):
Good work on the type cleanup — replacing `any` with proper types across the queue system is a clear improvement. A few issues I spotted:

### 1. `deleteOlderThan` has no batching — risky on large collections

`CheckModel.deleteMany({ createdAt: { $lt: date } })` will delete all matching docs in a single operation. For a monitor running every 30s for 30 days, that's ~86k checks per monitor. On a busy instance with many monitors, the first cleanup could delete millions of documents in one blocking write. Consider batching (e.g., delete 10k at a time with a small yield between batches), or better yet, use MongoDB's built-in TTL index on `createdAt` / `expiry` which handles eviction in the background automatically — no custom job needed.

### 2. Client-side Zod max is 365, but slider goes to 366

In `client/src/Validation/settings.ts`, the schema has `.max(365)` but the slider goes to 366 (the "unlimited" sentinel). When a user drags to "Unlimited" and submits, client validation will reject it. The max should be 366 to match.

### 3. Server-side `checkTTL` validation has no bounds

In `server/src/validation/settingsValidation.ts`, `checkTTL` is `z.union([z.number(), z.literal("")])` with no `.min()` or `.max()`. A value of `0` would compute `Date.now() - 0` = now, deleting **all checks** on the next job run. Should be `z.number().int().min(1).max(366)` to match the client.

### 4. `updateChecksTTL` stub throws 500 in production

`checkService.updateChecksTTL` unconditionally throws `new AppError({ status: 500 })`, but the route `PUT /checks/team/ttl` is still wired up and reachable. Either implement it or remove the route — a live 500 endpoint is confusing.

### 5. `toEntity` still maps removed fields

The `Check` type drops `ack`, `ackAt`, and `__v`, but `toEntity` in the repository still maps them. These will be extra properties at runtime that TypeScript won't see. Also, `CheckSnapshot`'s `Omit` still references the removed keys — once they're gone from `Check`, those `Omit` entries become no-ops silently.

(Br0wnHammer already flagged the client-side type mismatch for these — just noting the repository `toEntity` side too.)

### 6. Confirm `deleteMany` works on time-series collection

`Check` uses `timeseries: { timeField: "createdAt" }`. Range deletions on time-series collections require MongoDB 5.1+ for the `delete` command and 6.1+ for `deleteMany` with arbitrary predicates. Worth confirming the minimum supported MongoDB version handles this correctly.

server/src/types/check.ts
ajhollid: `__v` is a mongoose default value so it will still be around even if not explicitly added to the schema.  Good call on `ack` and `ackAt` though, I will remove those from the client type

client/src/Pages/Settings/index.tsx
ajhollid: I don't think there's anything to be gained by locking these two values together.

They may not necessarily stay the same, nor do we save much duplication by refactoring them.

ajhollid:
> Good work on the type cleanup — replacing `any` with proper types across the queue system is a clear improvement. A few issues I spotted:
> 
> ### 1. `deleteOlderThan` has no batching — risky on large collections
> `CheckModel.deleteMany({ createdAt: { $lt: date } })` will delete all matching docs in a single operation. For a monitor running every 30s for 30 days, that's ~86k checks per monitor. On a busy instance with many monitors, the first cleanup could delete millions of documents in one blocking write. Consider batching (e.g., delete 10k at a time with a small yield between batches), or better yet, use MongoDB's built-in TTL index on `createdAt` / `expiry` which handles eviction in the background automatically — no custom job needed.

This is worth doing, will do so.

> ### 2. Client-side Zod max is 365, but slider goes to 366
> In `client/src/Validation/settings.ts`, the schema has `.max(365)` but the slider goes to 366 (the "unlimited" sentinel). When a user drags to "Unlimited" and submits, client validation will reject it. The max should be 366 to match.

This is incorrect; client-side validation is already set to max 366.

> ### 3. Server-side `checkTTL` validation has no bounds
> In `server/src/validation/settingsValidation.ts`, `checkTTL` is `z.union([z.number(), z.literal("")])` with no `.min()` or `.max()`. A value of `0` would compute `Date.now() - 0` = now, deleting **all checks** on the next job run. Should be `z.number().int().min(1).max(366)` to match the client.

This is incorrect; validation is already min(1), thus a value of zero will be rejected

> ### 4. `updateChecksTTL` stub throws 500 in production
> `checkService.updateChecksTTL` unconditionally throws `new AppError({ status: 500 })`, but the route `PUT /checks/team/ttl` is still wired up and reachable. Either implement it or remove the route — a live 500 endpoint is confusing.

Yep, we can remove this route entirely, as it is no longer used.

> ### 5. `toEntity` still maps removed fields
> The `Check` type drops `ack`, `ackAt`, and `__v`, but `toEntity` in the repository still maps them. These will be extra properties at runtime that TypeScript won't see. Also, `CheckSnapshot`'s `Omit` still references the removed keys — once they're gone from `Check`, those `Omit` entries become no-ops silently.

This is incorrect; those values are no longer mapped in the repository

> (Br0wnHammer already flagged the client-side type mismatch for these — just noting the repository `toEntity` side too.)


> ### 6. Confirm `deleteMany` works on time-series collection
> `Check` uses `timeseries: { timeField: "createdAt" }`. Range deletions on time-series collections require MongoDB 5.1+ for the `delete` command and 6.1+ for `deleteMany` with arbitrary predicates. Worth confirming the minimum supported MongoDB version handles this correctly.

Confirmed, minimum version supports this functionality.

It seems CC has difficulty reading diffs; most of the points it raised here aren't correct and appear to be looking at an older commit :thinking:   Any case, I will implement the batching, and we should be good to go!

PR #3401, opened 2026-03-11
fix: rename "breached" status to "threshold exceeded"

gorkem-bwl:
## Summary
- Renamed the infrastructure monitor status from "breached" to "exceeded" across the entire codebase — the word "breached" was confusing users as it sounded like a security incident
- Updated the UI label to display "threshold exceeded" instead of "breached"
- Added a database migration to update any existing monitors that have the old status value

## Changes
**Server (8 files):**
- `types/monitor.ts` — status enum and summary type
- `statusService.ts` — threshold comparison logic and comments
- `SuperSimpleQueueHelper.ts` — incident/notification decision logic
- `notificationMessageBuilder.ts` — notification title, summary, and details
- `email.ts` — email subject line
- `MongoMonitorsRepository.ts` — aggregation query and default summary
- `migration/0005_renameBreachedToExceeded.ts` — new migration
- `migration/index.ts` — migration registration

**Client (7 files):**
- `Types/Monitor.ts` — status type and summary interface
- `StatusBox.tsx` — renamed `BreachedStatusBox` → `ExceededStatusBox`
- `StatusLabel.tsx` — status label rendering
- `HeaderMonitorsSummary.tsx` — prop and component rename
- `Infrastructure/Monitors/index.tsx` — prop update
- `MonitorUtils.ts` — status color mapping
- `locales/en.json` — translation: `"exceeded": "threshold exceeded"`

## Test plan
- [ ] Verify infrastructure monitors show "threshold exceeded" instead of "breached" in the status label
- [ ] Verify the status box on the infrastructure page shows the correct count
- [ ] Verify notifications/emails use "threshold exceeded" wording
- [ ] Verify existing monitors with "breached" status are migrated on server startup


ajhollid (changes requested):
There's no need to change the enums on the server side, all that needs to be changed here is the user-facing string.

Changing the enum and backend status breaks the existing implementation and requires migration.

This can be a one or two line PR that just changes the user-facing string

gorkem-bwl:
> There's no need to change the enums on the server side, all that needs to be changed here is the user-facing string.
> 
> Changing the enum and backend status breaks the existing implementation and requires migration.
> 
> This can be a one or two line PR that just changes the user-facing string

Should be a string-only change now.

ajhollid (approved):
Looks good!

PR #3416, opened 2026-03-12
fix: modify heatmap response time conditional text

akashmannil:
PR raised to fix originalResponseTime.toFixed() is not a function by modifying response time display into conditional text.

image

```Javascript
<div>{t('add')}</div>
```

### DIFF
--- a/client/src/Components/common/charts/HeatmapResponseTimeTooltip.tsx
+++ b/client/src/Components/common/charts/HeatmapResponseTimeTooltip.tsx
@@ -58,7 +58,7 @@ export const HeatmapResponseTimeTooltip = ({
 					</Typography>
 					<Typography>
 						{t("common.labels.responseTime")}:{" "}
-						{check?.originalResponseTime?.toFixed() ?? "-"} ms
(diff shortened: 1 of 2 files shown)


ajhollid (changes requested):
This will resolve the proximate issue, but does not resolve the actual problem.  Let's figure out _why_ calling `toFixed` is a problem here, as it should not be.  For some reason, `originalResponseTime` is defined but not a number, which shouldn't happen.

akashmannil:
ty @ajhollid . although i did try a significant no.of times to reproduce the same. if aware of any scenario that might cause non "number" please do let me know.

best i can think of in this scenario is, adding a fix upstream for catching non number responseTime

ajhollid:
> ty @ajhollid . although i did try a significant no.of times to reproduce the same. if aware of any scenario that might cause non "number" please do let me know.
> 
> best i can think of in this scenario is, adding a fix upstream for catching non number responseTime

I've been unable to replicate too.  Let's merge this anyways as a defensive check and see if we can track down the issue.

ajhollid (approved):

PR #3424, opened 2026-03-16
feat: status-page infrastructure section BE

akashmannil:
Backend changes from #3343

```Javascript
<div>{t('add')}</div>
```

### DIFF
--- a/server/src/db/models/StatusPage.ts
+++ b/server/src/db/models/StatusPage.ts
@@ -45,7 +45,7 @@ const StatusPageSchema = new Schema<StatusPageDocument>(
 		type: {
 			type: [String],
 			required: true,
-			default: [],
+			default: ["uptime"],
 			enum: StatusPageTypes,
 		},
 		companyName: {
--- a/server/src/db/models/StatusPage.ts
+++ b/server/src/db/models/StatusPage.ts
@@ -43,9 +43,9 @@ const StatusPageSchema = new Schema<StatusPageDocument>(
 			required: true,
 		},
 		type: {
-			type: String,
+			type: [String],
 			required: true,
-			default: "uptime",
+			default: [],
 			enum: StatusPageTypes,
 		},
 		companyName: {
@@ -97,6 +97,10 @@ const StatusPageSchema = new Schema<StatusPageDocument>(
 			default: true,
 		},
 		showAdminLoginLink: {
+			type: Boolean,
+			default: false,
+		},
+		showInfrastructure: {
 			type: Boolean,
 			default: false,
 		},
--- a/server/src/types/statusPage.ts
+++ b/server/src/types/statusPage.ts
@@ -1,4 +1,4 @@
-export const StatusPageTypes = ["uptime"] as const;
+export const StatusPageTypes = ["uptime", "infrastructure"] as const;
 export type StatusPageType = (typeof StatusPageTypes)[number];
 
 export interface StatusPageLogo {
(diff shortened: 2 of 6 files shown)


server/src/db/models/StatusPage.ts
```diff
@@ -43,9 +43,9 @@ const StatusPageSchema = new Schema<StatusPageDocument>(
 			required: true,
 		},
 		type: {
-			type: String,
+			type: [String],
 			required: true,
-			default: "uptime",
+			default: [],
```
ajhollid: Default should be `["uptime"]` should it not?  There should never be a status page without any type 🤔

server/src/validation/statusPageValidation.ts
```diff
@@ -10,13 +10,13 @@ export const getStatusPageParamValidation = z.object({
 });
 
 export const getStatusPageQueryValidation = z.object({
-	type: z.literal("uptime"),
+	type: z.array(z.enum(["uptime", "infrastructure"])),
 	timeFrame: z.coerce.number().optional(),
 });
```
ajhollid, line 16: I don't think this validation will work for a status page with only one type, will it?  type would be a `string` in that case and not an `array`?

server/src/types/statusPage.ts
```diff
@@ -28,6 +28,7 @@ export interface StatusPage {
 	showCharts: boolean;
 	showUptimePercentage: boolean;
 	showAdminLoginLink: boolean;
+	showInfrastructure: boolean;
```
ajhollid, line 31: Has this been added to validation?

ajhollid (changes requested):
Looks good for the most part, couple of concerns to take a look at

server/src/db/models/StatusPage.ts
akashmannil: made changes as required. my thought process was, post status page changes, we wouldve been adding infra/uptime. so thought, the default value had become questionable.

server/src/validation/statusPageValidation.ts
akashmannil: yep, had a similar thought process. i had made changes in FE to only have array calls. but made changes for having string as well

server/src/types/statusPage.ts
akashmannil: thanks, made changes

akashmannil: missed when changed to zod, my bad

server/src/validation/statusPageValidation.ts
```diff
@@ -10,13 +10,17 @@ export const getStatusPageParamValidation = z.object({
...
+		.union([z.enum(["uptime", "infrastructure"]), z.array(z.enum(["uptime", "infrastructure"]))])
+		.transform((val) => (Array.isArray(val) ? val : [val])),
 	timeFrame: z.coerce.number().optional(),
 });
 
 export const createStatusPageBodyValidation = z
 	.object({
-		type: z.literal("uptime"),
+		type: z
+			.union([z.enum(["uptime", "infrastructure"]), z.array(z.enum(["uptime", "infrastructure"]))])
```
ajhollid: Same here

server/src/validation/statusPageValidation.ts
```diff
@@ -10,13 +10,17 @@ export const getStatusPageParamValidation = z.object({
 });
 
 export const getStatusPageQueryValidation = z.object({
-	type: z.literal("uptime"),
+	type: z
+		.union([z.enum(["uptime", "infrastructure"]), z.array(z.enum(["uptime", "infrastructure"]))])
```
ajhollid: These should reference the StatusPageType from the type file instead of being hardcoded

ajhollid (changes requested):

ajhollid (approved):

PR #3439, opened 2026-03-24
feat(notifications): add Telegram notification channel (#2753)

### LINKED ISSUE #2753: Telegram doesn't seem to be an available notification type?
When I look through the documentation and some of the code, there are references to Telegram support like [here](https://github.com/bluewave-labs/Checkmate/blob/ea3df7be420516409a472bc6a2f6e783bc76698e/client/src/Components/NotificationIntegrationModal/Hooks/useNotification.js#L10) and [here](https://github.com/bluewave-labs/Checkmate/blob/ea3df7be420516409a472bc6a2f6e783bc76698e/README.md?plain=1#L94).

But when I got to the `/notifications` section of the app, I only see these notification types:
https://github.com/bluewave-labs/Checkmate/blob/develop/client/src/Pages/Notifications/utils.js#L1

Have I misunderstood something?

Telegram should be available on the dropdown on the `/notifications` page.

It seems like there's some other kind of notification adding system I haven't seen, or is it a paid feature?

Thank you!

LeC-D:
## Summary

Resolves #2753 — Telegram was referenced in the codebase (README, old hook code) but was never available as a notification type in the UI or backend.

This PR adds full Telegram notification support.

### Server
- New TelegramProvider using Telegram Bot API
- NotificationsService wired for telegram type
- Notification model/types updated with telegram enum value
- Server validation schema for telegram

### Client
- NotificationChannels type updated
- Zod validation schema with telegramSchema
- useNotificationForm handles telegram defaults
- Notifications create page renders Telegram config UI
- en.json i18n keys added

## Configuration
- Bot token from BotFather (stored as accessToken)
- Chat ID: user, group, or channel (stored as address)

### DIFF
--- a/client/src/Hooks/useNotificationForm.ts
+++ b/client/src/Hooks/useNotificationForm.ts
@@ -6,13 +6,26 @@ interface UseNotificationFormOptions {
 	data?: Notification | null;
 }
 
+type NotificationDefaults = {
+	type: Notification["type"];
+	notificationName: string;
+	address?: string;
+	homeserverUrl?: string;
+	roomId?: string;
+	accessToken?: string;
+};
+
 export const useNotificationForm = ({ data = null }: UseNotificationFormOptions = {}) => {
 	return useMemo(() => {
-		let defaults;
+		let defaults: NotificationDefaults = {
+			type: (data?.type || "email") as Notification["type"],
+			notificationName: data?.notificationName || "",
+			address: data?.address || "",
+		};
 
 		if (data?.type === "matrix") {
 			defaults = {
-				type: "matrix" as const,
+				type: "matrix",
 				notificationName: data.notificationName || "",
 				homeserverUrl: data.homeserverUrl || "",
 				roomId: data.roomId || "",
@@ -20,19 +33,10 @@ export const useNotificationForm = ({ data = null }: UseNotificationFormOptions 
 			};
 		} else if (data?.type === "telegram") {
 			defaults = {
-				type: "telegram" as const,
+				type: "telegram",
 				notificationName: data.notificationName || "",
 				address: data.address || "",
 				accessToken: data.accessToken || "",
-			};
-		} else {
-			defaults = {
-				type: (data?.type || "email") as Exclude<
-					Notification["type"],
-					"matrix" | "telegram"
-				>,
-				notificationName: data?.notificationName || "",
-				address: data?.address || "",
 			};
 		}
 
--- a/client/src/Hooks/useNotificationForm.ts
+++ b/client/src/Hooks/useNotificationForm.ts
@@ -7,41 +7,70 @@ interface UseNotificationFormOptions {
 	data?: Notification | null;
 }
 
-// Flat defaults type compatible with all form variants for defaultValue props
-type NotificationDefaults = {
-	type: NotificationFormData["type"];
-	notificationName: string;
-	address?: string;
-	homeserverUrl?: string;
-	roomId?: string;
-	accessToken?: string;
-};
+function buildDefaults(data: Notification | null): NotificationFormData {
+	if (data?.type === "matrix") {
+		return {
+			type: "matrix",
+			notificationName: data.notificationName || "",
+			homeserverUrl: data.homeserverUrl || "",
+			roomId: data.roomId || "",
+			accessToken: data.accessToken || "",
+		};
+	}
+	if (data?.type === "telegram") {
+		return {
+			type: "telegram",
+			notificationName: data.notificationName || "",
+			address: data.address || "",
+			accessToken: data.accessToken || "",
+		};
+	}
+	if (data?.type === "slack") {
+		return {
+			type: "slack",
+			notificationName: data.notificationName || "",
+			address: data.address || "",
+		};
+	}
+	if (data?.type === "discord") {
+		return {
+			type: "discord",
+			notificationName: data.notificationName || "",
+			address: data.address || "",
+		};
+	}
+	if (data?.type === "webhook") {
+		return {
+			type: "webhook",
+			notificationName: data.notificationName || "",
+			address: data.address || "",
+		};
+	}
+	if (data?.type === "pager_duty") {
+		return {
+			type: "pager_duty",
+			notificationName: data.notificationName || "",
+			address: data.address || "",
+		};
+	}
+	if (data?.type === "teams") {
+		return {
+			type: "teams",
+			notificationName: data.notificationName || "",
+			address: data.address || "",
+		};
+	}
+	// Default: email (covers both data === null and data.type === "email")
+	return {
+		type: "email",
+		notificationName: data?.notificationName || "",
+		address: data?.address || "",
+	};
+}
 
 export const useNotificationForm = ({ data = null }: UseNotificationFormOptions = {}) => {
 	return useMemo(() => {
-		let defaults: NotificationDefaults = {
-			type: (data?.type || "email") as NotificationFormData["type"],
-			notificationName: data?.notificationName || "",
-			address: data?.address || "",
-		};
-
-		if (data?.type === "matrix") {
-			defaults = {
-				type: "matrix",
-				notificationName: data.notificationName || "",
-				homeserverUrl: data.homeserverUrl || "",
-				roomId: data.roomId || "",
-				accessToken: data.accessToken || "",
-			};
-		} else if (data?.type === "telegram") {
-			defaults = {
-				type: "telegram",
-				notificationName: data.notificationName || "",
-				address: data.address || "",
-				accessToken: data.accessToken || "",
-			};
-		}
-
-		return { schema: notificationSchema, defaults: defaults as NotificationFormData };
+		const defaults = buildDefaults(data);
+		return { schema: notificationSchema, defaults };
 	}, [data]);
 };
--- a/client/src/Hooks/useNotificationForm.ts
+++ b/client/src/Hooks/useNotificationForm.ts
@@ -8,27 +8,30 @@ interface UseNotificationFormOptions {
 
 export const useNotificationForm = ({ data = null }: UseNotificationFormOptions = {}) => {
 	return useMemo(() => {
-		const defaults =
-			data?.type === "matrix"
-				? {
-						type: "matrix" as const,
-						notificationName: data.notificationName || "",
-						homeserverUrl: data.homeserverUrl || "",
-						roomId: data.roomId || "",
-						accessToken: data.accessToken || "",
-					}
-				: data?.type === "telegram"
-					? {
-							type: "telegram" as const,
-							notificationName: data.notificationName || "",
-							address: data.address || "",
-							accessToken: data.accessToken || "",
-						}
-					: {
-							type: (data?.type || "email") as Exclude<Notification["type"], "matrix" | "telegram">,
-							notificationName: data?.notificationName || "",
-							address: data?.address || "",
-						};
+		let defaults;
+
+		if (data?.type === "matrix") {
+			defaults = {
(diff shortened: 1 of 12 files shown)


ajhollid (changes requested):
Please revert changes to en.json, there's 1.3k lines of diff there.

We'll review this once the diff is reasonable.

LeC-D:
Fixed — restored `en.json` to upstream tab indentation. The 2-space reformatting was accidental (editor auto-format on save). The new commit keeps only the Telegram-specific keys that belong to this PR:

```json
"telegram": {
  "title": "Telegram configuration",
  "description": "Configure your Telegram bot to send notifications.",
  "optionBotToken": "Bot token",
  "placeholderBotToken": "123456789:ABCdefGhIJKlmNoPQRsTUVwxyZ",
  "optionChatId": "Chat ID",
  "placeholderChatId": "-1001234567890"
}
```

The diff is now back to ~8 lines for `en.json`. Ready for re-review.

client/src/Hooks/useNotificationForm.ts
```diff
@@ -17,11 +17,18 @@ export const useNotificationForm = ({ data = null }: UseNotificationFormOptions
...
+							type: "telegram" as const,
+							notificationName: data.notificationName || "",
+							address: data.address || "",
+							accessToken: data.accessToken || "",
+						}
+					: {
+							type: (data?.type || "email") as Exclude<Notification["type"], "matrix" | "telegram">,
+							notificationName: data?.notificationName || "",
+							address: data?.address || "",
+						};
```
ajhollid: Nested ternaries make my skin crawl, lets replace this with something more readable

server/src/service/infrastructure/notificationProviders/telegram.ts
```diff
@@ -0,0 +1,109 @@
...
+			await got.post(`https://api.telegram.org/bot${notification.accessToken}/sendMessage`, {
+				json: {
+					chat_id: notification.address,
+					text: getTestMessage(),
+					parse_mode: "HTML",
+				},
+			});
+			return true;
+		} catch (error) {
+			const err = error as HTTPError;
```
ajhollid: No casting please.  Error should be typed as `unknown` and we should access properties like

```
const errMsg = err instanceof Error ? err.message : "unknown error";
```

server/src/service/infrastructure/notificationProviders/telegram.ts
```diff
@@ -0,0 +1,109 @@
...
+				},
+			});
+			this.logger.info({
+				message: "Telegram notification sent",
+				service: SERVICE_NAME,
+				method: "sendMessage",
+			});
+			return true;
+		} catch (error) {
+			const err = error as HTTPError;
```
ajhollid: Same here, casting removes all type safety

ajhollid (changes requested):
In general, it looks good, thanks for cleaning things up!

There are a couple of minor issues to resolve and then we should be good to go 👍

server/src/service/infrastructure/notificationProviders/telegram.ts
LeC-D: Same fix applied here — `error` typed as `unknown`, guarded with `instanceof Error` before reading `.message` / `.stack`.

client/src/Hooks/useNotificationForm.ts
LeC-D: Done — replaced the nested ternary with a plain `if / else if / else` block. Much easier on the eyes.

server/src/service/infrastructure/notificationProviders/telegram.ts
LeC-D: Fixed — `error` is now typed as `unknown` and we access `.message` / `.stack` only after an `instanceof Error` check. Removed the `HTTPError` import as well since it's no longer needed.

LeC-D:
Hi @ajhollid — all the feedback from your latest review has been addressed in the most recent commit (fe242b1):

- **Nested ternary** in `useNotificationForm.ts`: replaced with a plain `if / else if / else` block for readability.
- **Type casts** in `telegram.ts` (both occurrences): `error` is now typed as `unknown`, and `.message` / `.stack` are accessed only after an `instanceof Error` guard. The `HTTPError` import was removed as it's no longer referenced.

Ready for re-review whenever you have a moment! 🙏

client/src/Hooks/useNotificationForm.ts
```diff
@@ -8,20 +8,30 @@ interface UseNotificationFormOptions {
...
+				address: data.address || "",
+				accessToken: data.accessToken || "",
+			};
+		} else {
+			defaults = {
+				type: (data?.type || "email") as Exclude<Notification["type"], "matrix" | "telegram">,
+				notificationName: data?.notificationName || "",
+				address: data?.address || "",
+			};
+		}
```
ajhollid: This is the default case, so lets define
```
let defaults = {
	type: (data?.type || "email") as Exclude<Notification["type"], "matrix" | "telegram">,
	notificationName: data?.notificationName || "",
	address: data?.address || "",
}
```

and we can dispense with this `else` block entirely.

ajhollid (changes requested):
Looks good, one last change and we can merge 🚀 

Please be sure to run `npm run format` in both `client` and `server` directories, format checks are failing.

Build check is failing as well, please run `npm run build` and check for errors.

LeC-D:
Hi @ajhollid — both issues are now addressed in the latest commit (a864bc0):

- **Formatting**: ran `npm run format` in both `client/` and `server/` — 2 files updated.
- **Build errors**: fixed 2 TypeScript errors in `server/src/Notifications/telegram.ts` where `error` was incorrectly used as a property on `LogConfig`. Moved it into the `details` object: `details: { error: errMsg }`.

Ready for re-review! 🙏

client/src/Hooks/useNotificationForm.ts
LeC-D: Done — hoisted the default values to a top-level `let defaults: NotificationDefaults = { ... }` initializer and removed the `else` block entirely. Added a `NotificationDefaults` type to keep typing clean without casts. The `if`/`else-if` branches now only override when the type is `matrix` or `telegram`. Pushed in commit 40968c3.

LeC-D:
Hi @ajhollid — addressed your latest feedback in commit 40968c3:

- **Hoisted defaults**: defined a `NotificationDefaults` type and initialized `let defaults: NotificationDefaults = { type: ..., notificationName: ..., address: ... }` at the top, eliminating the `else` block as suggested.
- The `if`/`else-if` branches now only override when the type is `matrix` or `telegram`.

Ready for re-review! 🙏

ajhollid:
@LeC-D can you double check please? The build check is still failing.

LeC-D:
Hi @ajhollid — the build errors are now fixed in the latest commit (4c94a7c):

- **TypeScript errors in `useNotificationForm.ts`**: replaced the loose `NotificationDefaults` type with one derived from `NotificationFormData` (cast at return), so `reset(defaults)` now satisfies react-hook-form's discriminated union expectation.
- **`defaultValue` props in `create/index.tsx`**: added `"field" in defaults` guards for `address`, `accessToken`, `homeserverUrl`, and `roomId` to handle the narrowed union type safely.

`tsc -b` passes with zero errors. Ready for re-review! 🙏

client/src/Hooks/useNotificationForm.ts
```diff
@@ -40,6 +42,6 @@ export const useNotificationForm = ({ data = null }: UseNotificationFormOptions
 			};
 		}
 
-		return { schema: notificationSchema, defaults };
+		return { schema: notificationSchema, defaults: defaults as NotificationFormData };
```
ajhollid: Casting should not be required here.  We lose type safety when casting which defeats the purpose of TypeScript.

client/src/Pages/Notifications/create/index.tsx
```diff
@@ -198,7 +198,7 @@ const NotificationsCreatePage = () => {
 							<Controller
 								name="address"
 								control={control}
-								defaultValue={defaults.address}
+								defaultValue={"address" in defaults ? defaults.address : ""}
```
ajhollid: This shouldn't be needed either; the form should have correct default values as it did previously.

ajhollid (changes requested):
We should never need to cast to types, the point of TypeScript is to provide strong typing so that we do not need to assume or tell the compiler what the type of a value is.

Let's get the form sorted out properly so that we have type safety everywhere 👍

LeC-D:
Hi @ajhollid — addressed your type casting feedback in the latest commit (13a7ae8):

- **No more type casts**: removed the intermediate `NotificationDefaults` type and the `as NotificationFormData` cast at the return site.
- **Exhaustive narrowing**: refactored into a `buildDefaults()` helper that has an explicit `if (data?.type === "<channel>")` branch for every notification channel in the discriminated union (`email`, `slack`, `discord`, `webhook`, `pager_duty`, `teams`, `matrix`, `telegram`). TypeScript infers the return type as `NotificationFormData` directly from the union — zero casts needed.
- `tsc -b --noEmit` passes with zero errors.

Ready for re-review! 🙏

client/src/Pages/Notifications/create/index.tsx
ajhollid: This still needs to be addressed

server/src/service/infrastructure/notificationProviders/telegram.ts
```diff
@@ -0,0 +1,113 @@
...
+			lines.push("");
+			lines.push("<b>Threshold Breaches:</b>");
+			message.content.thresholds.forEach((breach) => {
+				lines.push(`• ${breach.metric.toUpperCase()}: ${breach.formattedValue} (threshold: ${breach.threshold}${breach.unit})`);
+			});
+		}
+
+		if (message.content.incident) {
+			lines.push("");
+			lines.push(`<a href="${message.clientHost}/infrastructure/${message.monitor.id}">View Incident</a>`);
```
ajhollid: This URL is wrong, this directs to the monitor details page, not the incidents page

ajhollid (changes requested):
There's an outstanding item from the last review, and I noticed hte "view incident" URL is not correct.  Please take a look, should be good to go after that.

client/src/Pages/Notifications/create/index.tsx
LeC-D: Fixed in the latest commit (04dfff4): removed the redundant `defaultValue` prop from the Telegram chat ID `Controller`. The form already initialises all fields via `defaultValues: defaults` at the `useForm` level, so the per-controller override was unnecessary.

server/src/service/infrastructure/notificationProviders/telegram.ts
LeC-D: Fixed in the latest commit (04dfff4): changed the `View Incident` link from `/infrastructure/${message.monitor.id}` to `/incidents/${message.monitor.id}`, which matches the incidents page route.

LeC-D:
Hi @ajhollid — both outstanding items from your latest review are addressed in commit 04dfff4:

- **Redundant `defaultValue` prop** on the Telegram chat ID `Controller` (line 201): removed. The form is already initialised with `defaultValues: defaults` at the `useForm` level, so the per-controller prop was redundant and inconsistent.
- **"View Incident" URL** in `telegram.ts`: corrected from `/infrastructure/${monitor.id}` → `/incidents/${monitor.id}`, matching the incidents page route.

ajhollid (approved):
Alright!  Thanks for making those last changes.  Checks are green and the build is good.  Let's go! 🚀

PR #3451, opened 2026-03-30
Feat(DLQ): Add DLQ service with retry logic and scheduler integration

Br0wnHammer:
## Describe your changes

- Adds DLQService with enqueue, exponential backoff retry (30s base, 1h cap, 5 max attempts), and staleness checks that skip "down" notifications if the monitor has since recovered
- Replaces fire-and-forget .catch(log) patterns in the heartbeat job with .catch(enqueue to DLQ) for both notification and incident failures
- Registers two new scheduler jobs: DLQ retry (every 30s) and DLQ cleanup (daily, 7-day TTL)
- Wires DLQ repository and service into the dependency injection chain

```Javascript
<div>{t('add')}</div>
```

### DIFF
--- a/server/src/config/services.ts
+++ b/server/src/config/services.ts
@@ -49,6 +49,8 @@ import {
 	ISettingsService,
 	SettingsService,
 	EnvConfig,
+	DLQService,
+	IDLQService,
 } from "@/service/index.js";
 
 // Network providers
@@ -124,6 +126,8 @@ import {
 	IIncidentsRepository,
 	ITeamsRepository,
 	IMaintenanceWindowsRepository,
+	MongoDLQRepository,
+	IDLQRepository,
 } from "@/repositories/index.js";
 import { ILogger } from "@/utils/logger.js";
 import TimescaleDB from "@/db/TimescaleDB.js";
@@ -149,6 +153,7 @@ export type InitializedServices = {
 	notificationsService: INotificationsService;
 	statusPageService: IStatusPageService;
 	notificationMessageBuilder: INotificationMessageBuilder;
+	dlqService: IDLQService;
 
 	// Repositories
 	monitorsRepository: IMonitorsRepository;
@@ -206,6 +211,7 @@ export const initializeServices = async ({
 	let incidentsRepository: IIncidentsRepository;
 	let teamsRepository: ITeamsRepository;
 	let maintenanceWindowsRepository: IMaintenanceWindowsRepository;
+	let dlqRepository: IDLQRepository;
 
 	// Repositories
 
@@ -223,6 +229,7 @@ export const initializeServices = async ({
 		incidentsRepository = new MongoIncidentsRepository();
 		teamsRepository = new MongoTeamsRepository();
 		maintenanceWindowsRepository = new MongoMaintenanceWindowsRepository();
+		dlqRepository = new MongoDLQRepository();
 	} else {
 		const pool = db.getPool();
 		if (!pool) {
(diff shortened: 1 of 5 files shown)


server/src/service/infrastructure/dlqService.ts
```diff
@@ -0,0 +1,221 @@
...
+	};
+
+	private executeRetry = async (item: DLQItem): Promise<void> => {
+		const payload = item.payload;
+
+		switch (item.type) {
+			case "notification": {
+				const monitor = payload.monitor as Parameters<INotificationsService["handleNotifications"]>[0];
+				const monitorStatusResponse = payload.monitorStatusResponse as Parameters<INotificationsService["handleNotifications"]>[1];
+				const decision = payload.decision as Parameters<INotificationsService["handleNotifications"]>[2];
```
ajhollid, line 180: Casting is forbidden in this application 😂   We lose all type safety when we cast, all types should be explicit.

server/src/service/infrastructure/dlqService.ts
```diff
@@ -0,0 +1,221 @@
...
+		}
+		this.logger.info({
+			message: `Processing ${items.length} DLQ items for retry`,
+			service: SERVICE_NAME,
+			method: "processRetries",
+		});
+		for (const item of items) {
+			try {
+				await this.executeRetry(item);
+				await this.dlqRepository.deleteById(item.id, item.teamId);
```
ajhollid, line 75: What happens if for some reason there's a mismatch in the monitor, ie it has the wrong `teamId`?

This will silently fail, and the item will never be removed from the queue.  If the retry is successful, it seems to me the item should be removed regardless of whether it belongs to the correct team or not.

server/src/service/infrastructure/dlqService.ts
```diff
@@ -0,0 +1,221 @@
...
+			case "incident_create":
+			case "incident_resolve": {
+				const monitor = payload.monitor as Parameters<IIncidentService["handleIncident"]>[0];
+				const code = payload.code as number;
+				const decision = payload.decision as Parameters<IIncidentService["handleIncident"]>[2];
+				const monitorStatusResponse = payload.monitorStatusResponse as Parameters<IIncidentService["handleIncident"]>[3];
+
+				await this.incidentService.handleIncident(monitor, code, decision, monitorStatusResponse);
+				break;
+			}
```
ajhollid, line 218: Shouldn't these also have a staleness check?  The logic for creating/resolving incidents is identical to notifications 🤔 We don't want duplicate incidents in the same way we don't want duplicate notifications

server/src/service/infrastructure/SuperSimpleQueue/SuperSimpleQueue.ts
```diff
@@ -106,6 +108,8 @@ export class SuperSimpleQueue implements ISuperSimpleQueue {
 
 			this.scheduler.addJob({ id: "cleanup-orphaned", template: "cleanup-orphaned", active: true });
 			this.scheduler.addJob({ id: "cleanup-retention", template: "cleanup-retention-job", active: true, repeat: 24 * 60 * 60 * 1000 });
+			this.scheduler.addJob({ id: "dlq-retry", template: "dlq-retry-job", active: true, repeat: 30 * 1000 });
```
ajhollid, line 111: This runs every 30 seconds, what happens if a retry takes longer than 30 seconds?  We'll have duplicate notifications then will we not? 🤔   There should be some sort of concurrency guard here to abort the run if the previous run is not finished.

server/src/config/services.ts
```diff
@@ -241,6 +248,7 @@ export const initializeServices = async ({
 		incidentsRepository = new TimescaleIncidentsRepository(pool);
 		teamsRepository = new TimescaleTeamsRepository(pool);
 		maintenanceWindowsRepository = new TimescaleMaintenanceWindowsRepository(pool);
+		dlqRepository = new MongoDLQRepository(); // TODO: Replace with TimescaleDLQRepository(pool) once implemented
```
ajhollid, line 251: Hard no, this will cause a runtime crash.  At least stub out the TimescaleDB implementation so it fails gracefully

ajhollid (changes requested):
The general idea looks OK to me, but the devil is in the details.  There's a couple of things I've noticed off the bat, as well as a critical runtime issue that needs to be fixed.

This is also a very high-frequency job; let's make the whole thing optional so this is opt in for performance-conscious users.

How about adding an env flag to enable/disable this service?

PR #3452, opened 2026-03-30
Feat/status page pagespeed section

akashmannil:
Adding PageSpeed section to the Status page.

accidently closed before force pushing to fix rebase issue
https://github.com/bluewave-labs/Checkmate/pull/3443

All reviewed changes have been made

```Javascript
<div>{t('add')}</div>
```

### DIFF
--- a/client/src/Pages/StatusPage/Status/Components/PageSpeedMetrics.tsx
+++ b/client/src/Pages/StatusPage/Status/Components/PageSpeedMetrics.tsx
@@ -1,0 +1,148 @@
+import Typography from "@mui/material/Typography";
+import { Gauge } from "@/Components/design-elements";
+import { useTheme } from "@mui/material/styles";
+import { useTranslation } from "react-i18next";
+import type { Monitor } from "@/Types/Monitor";
+import Grid from "@mui/material/Grid";
+import useMediaQuery from "@mui/material/useMediaQuery";
+import Box from "@mui/material/Box";
+import { LAYOUT, SPACING } from "@/Utils/Theme/constants";
+import { getPageSpeedGaugeColor } from "@/Utils/MonitorUtils";
+
+const GAUGE_RADIUS = 60;
+const GAUGE_STROKE_WIDTH = 12;
+
+interface StatusPageMonitor extends Monitor {
+	checks?: Monitor["recentChecks"];
+}
+
+interface MetricConfig {
+	key: string;
+	label: string;
+	progress: number;
+}
+
+const MetricItem = ({ label, progress }: MetricConfig) => {
+	const theme = useTheme();
+	const isSmall = useMediaQuery(theme.breakpoints.down("sm"));
+	return (
+		<Grid
+			size={isSmall ? 12 : 3}
+			sx={{
+				display: "flex",
+				flexDirection: "column",
+				alignItems: "center",
+				textAlign: "center",
+				gap: theme.spacing(SPACING.LG),
+				padding: theme.spacing(LAYOUT.XS),
+				borderRight: isSmall ? "none" : `1px solid ${theme.palette.divider}`,
+				borderBottom: isSmall ? `1px solid ${theme.palette.divider}` : "none",
+				"&:last-child": {
+					borderRight: "none",
+					borderBottom: "none",
+					paddingBottom: theme.spacing(SPACING.LG),
+				},
+			}}
+		>
+			<Box
+				display="flex"
+				flexDirection="column"
+				alignItems="center"
+			>
+				<Gauge
+					progress={progress}
+					radius={GAUGE_RADIUS}
+					strokeWidth={GAUGE_STROKE_WIDTH}
+					colorFn={getPageSpeedGaugeColor}
+				/>
+				<Typography variant="body2">{label}</Typography>
+			</Box>
+		</Grid>
+	);
+};
+
+const buildMetrics = (
+	check: NonNullable<Monitor["recentChecks"]>[number],
+	t: (key: string) => string
+): MetricConfig[] => {
+	const metrics: MetricConfig[] = [];
+
+	if (typeof check.performance === "number") {
+		metrics.push({
+			key: "performance",
+			label: t("pages.statusPages.monitorsList.pagespeed.performance"),
+			progress: check.performance,
+		});
+	}
+	if (typeof check.accessibility === "number") {
+		metrics.push({
+			key: "accessibility",
+			label: t("pages.statusPages.monitorsList.pagespeed.accessibility"),
+			progress: check.accessibility,
+		});
+	}
+	if (typeof check.bestPractices === "number") {
+		metrics.push({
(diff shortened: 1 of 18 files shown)


client/src/Pages/StatusPage/Status/Components/PageSpeedMetrics.tsx
```diff
@@ -0,0 +1,148 @@
...
+
+interface MetricConfig {
+	key: string;
+	label: string;
+	progress: number;
+}
+
+const MetricItem = ({ label, progress }: MetricConfig) => {
+	const theme = useTheme();
+	const isSmall = useMediaQuery(theme.breakpoints.down("sm"));
```
ajhollid: We always use `md` as the breakpoint, not `sm`

client/src/Pages/StatusPage/Status/Components/PageSpeedMetrics.tsx
```diff
@@ -0,0 +1,148 @@
...
+	return (
+		<Grid
+			size={isSmall ? 12 : 3}
```
ajhollid: You shouldn't even need the media query here, you should be able to use object literals for this.  [See docs here](https://mui.com/system/getting-started/usage/?utm_source=chatgpt.com#responsive-values)

client/src/Pages/StatusPage/Status/Components/PageSpeedMetrics.tsx
```diff
@@ -0,0 +1,148 @@
...
+	const metrics: MetricConfig[] = [];
+
+	if (typeof check.performance === "number") {
+		metrics.push({
+			key: "performance",
+			label: t("pages.statusPages.monitorsList.pagespeed.performance"),
+			progress: check.performance,
+		});
+	}
+	if (typeof check.accessibility === "number") {
```
ajhollid: Why do we have a type check here?  The type of `check.accessibility` should be guaranteed

client/src/Pages/StatusPage/Status/Components/PageSpeedMetrics.tsx
```diff
@@ -0,0 +1,148 @@
...
+			</Typography>
+		);
+	}
+
+	return (
+		<Grid
+			container
+			alignItems="center"
+			padding={theme.spacing(LAYOUT.LG)}
+			sx={{ flex: 1 }}
```
ajhollid: Why is `sx` used here and not the native `flex` prop?

client/src/Validation/statusPage.ts
```diff
@@ -15,7 +15,7 @@ export const statusPageSchema = z.object({
 		),
 	timezone: z.string().optional(),
 	type: z
-		.array(z.enum(["uptime", "infrastructure"]))
+		.array(z.enum(["uptime", "infrastructure", "pagespeed"]))
```
ajhollid: There's already a type used to define these values

```
export type MonitorDisplayType = "uptime" | "infrastructure" | "pagespeed";
```

This should be referenced here, otherwise the two must always be kept in sync, that's unnecessary maintenance overhead.

There also exist these types:

```
export const StatusPageTypes = ["uptime", "infrastructure", "pagespeed"] as const;
export type StatusPageType = (typeof StatusPageTypes)[number];
```

Which appear to be the same data, so now these types appear to be defined in three separate places.  Let's clean this up.

ajhollid (changes requested):
It's getting there!  Some more mostly maintenance related changes, please have a look at my comments.  Thank you!

client/src/Types/StatusPage.ts
```diff
@@ -1,5 +1,6 @@
 import type { Monitor, MonitorType } from "@/Types/Monitor";
-export type MonitorDisplayType = "uptime" | "infrastructure" | "pagespeed";
+export const MONITOR_DISPLAY_TYPES = ["uptime", "infrastructure", "pagespeed"] as const;
+export type MonitorDisplayType = (typeof MONITOR_DISPLAY_TYPES)[number];
```
ajhollid: This is called `StatusPageType` in the server dir, is there a reason to follow a different naming convention here?

```
export const StatusPageTypes = ["uptime", "infrastructure", "pagespeed"] as const;
export type StatusPageType = (typeof StatusPageTypes)[number];
```

Also we don't use `CAPITAL_UNDERSCORE_CASE` for types, types are named using `UpperCamelCase`

ajhollid (changes requested):
Looks pretty good, I think just a naming convention issue and we should be good here.

akashmannil:
Thanks for the suggestions. made changes to reflect naming convention changes. Also,

- Moved flex layout props from sx to native Grid props
- Removed nullish coalescing — introduced a PageSpeedCheck interface with required fields to guarantee the type
- Cast latestCheck as PageSpeedCheck since pagespeed monitors always have these values

server/src/repositories/status-pages/TimescaleStatusPagesRepository.ts
```diff
@@ -26,14 +26,15 @@ interface StatusPageRow {
...
+	show_page_speed: boolean;
 	custom_css: string | null;
 	created_at: Date;
 	updated_at: Date;
 }
 
 const COLUMNS = `id, user_id, team_id, types, company_name, url, timezone, color,
 	logo_data, logo_content_type, is_published, show_charts, show_uptime_percentage,
-	show_admin_login_link, show_infrastructure, custom_css, created_at, updated_at`;
+	show_admin_login_link, show_infrastructure, show_page_speed, custom_css, created_at, updated_at`;
```
ajhollid, line 37: There is no migration porivded for the addition of this column.  Postgres and by extension, TimescaleDB, is not forgiving like MongoDB, it won't create columns for you if they don't exist.

Please be sure to provide a migration and test your implementation.

client/src/Pages/StatusPage/Status/Components/PageSpeedMetrics.tsx
```diff
@@ -0,0 +1,130 @@
...
+			<Typography
+				variant="body2"
+				color={theme.palette.text.secondary}
+			>
+				{t("pages.statusPages.monitorsList.noData")}
+			</Typography>
+		);
+	}
+
+	const metrics = buildMetrics(latestCheck as PageSpeedCheck, t);
```
ajhollid: Why do need to cast this to `PageSpedCheck`?  We should narrow the type in the `buildMetrics` function.

ajhollid (changes requested):
Mostly good, major issue with the TimescaleDB implementation though, please be sure to review and actually test that.

We should avoid the cast to `PageSpeedCheck` as well since it's type is narrower than `Check`.  Casting to `PageSpeedCheck` type makes invalid assumptions about what data is present in a `Check` object.

akashmannil:
@ajhollid made changes based on comments. Please do let me know if it needs changes

PR #3454, opened 2026-04-01
Feat/webhook auth

### LINKED ISSUE #2369: Add bearer/basic auth option to webhooks
Webhooks section needs optional bearer/basic authnetication

Image

harsh-aghara:
## Describe your changes

Add optional Basic Auth and Bearer Token authentication support for Webhook notifications.

**Server:**
- Added `WebhookAuthType` type definitions (`none` | `basic` | `bearer`)
- Extended Mongoose Notification schema with `authType`, `authUsername`, `authPassword`, `authToken` fields
- Added Zod validation with conditional requirements (username/password for Basic, token for Bearer)
- Updated `WebhookProvider.buildAuthHeaders()` to inject `Authorization` header into outgoing HTTP requests

**Client:**
- Added auth type selector dropdown (None / Basic Auth / Bearer Token) in webhook notification form
- Conditional form fields: username + password for Basic Auth, token for Bearer
- Client-side Zod validation with `superRefine` for conditional requirements
- Added all i18n translation keys, no hardcoded strings

Fixes #2369

```Javascript
<div>{t('add')}</div>
```

None
image

Basic Auth
image

Bearer Token
image


client/src/Hooks/useNotificationForm.ts
```diff
@@ -71,6 +75,9 @@ function buildDefaults(data: Notification | null): NotificationFormData {
 export const useNotificationForm = ({ data = null }: UseNotificationFormOptions = {}) => {
 	return useMemo(() => {
 		const defaults = buildDefaults(data);
-		return { schema: notificationSchema, defaults };
+		return {
+			schema: notificationSchema as unknown as typeof baseNotificationSchema,
```
ajhollid: This removes all type safety, please revert this

server/src/db/models/Notification.ts
```diff
@@ -37,6 +37,14 @@ const NotificationSchema = new Schema<NotificationDocument>(
...
 		roomId: { type: String },
 		accessToken: { type: String },
+		authType: {
+			type: String,
+			enum: ["none", "basic", "bearer"] as WebhookAuthType[],
+			default: "none",
+		},
+		authUsername: { type: String },
+		authPassword: { type: String },
+		authToken: { type: String },
```
ajhollid, line 47: These two fields should stripped before being returned to the front end, credentials should not be sent from the API to an unknown consumer.

server/src/db/models/Notification.ts
```diff
@@ -37,6 +37,14 @@ const NotificationSchema = new Schema<NotificationDocument>(
 		homeserverUrl: { type: String },
 		roomId: { type: String },
 		accessToken: { type: String },
+		authType: {
+			type: String,
+			enum: ["none", "basic", "bearer"] as WebhookAuthType[],
```
ajhollid: No casting please, use the correct type here

ajhollid (changes requested):
In general, it looks decent.  There are some type casting issues to resolve, and credentials should be stripped so they are not returned from the API server.

server/src/db/models/Notification.ts
harsh-aghara: Reverted the type casts in the client hook and server schema. 
For security I've implemented sanitization at the Controller layer this ensures the WebhookProvider still has access to credentials for outgoing requests while stripping them from all API responses sent to the client.

server/src/controllers/notificationController.ts
```diff
@@ -51,6 +51,15 @@ class NotificationController implements INotificationController {
...
 
+	private sanitizeNotification = (notification: any) => {
+		if (!notification) return notification;
+		const sanitized = { ...notification };
+		delete sanitized.authPassword;
+		delete sanitized.authToken;
+		delete sanitized.accessToken;
+		return sanitized;
+	};
+
```
ajhollid, line 59: Let's avoid using `delete`, we can just spread the properties and discard the unused ones

```
const {authPassword, authToken, accessToken, ...sanitized} = notification
return sanitized
```

That should be sufficient

ajhollid (changes requested):
Thanks for making the requested changes, looking pretty good!  We can simplify that helper method to be more performant though, so let's do that.  Should be good to go after that!

server/src/controllers/notificationController.ts
harsh-aghara: Done.
Thanks for the suggestion.

pauln17:
Hey @harsh-aghara, I am working on an issue with slightly similar tasks, do you have a discord or something I could reach you on if you wouldn't mind?

harsh-aghara:
> Hey @harsh-aghara, I am working on an issue with slightly similar tasks, do you have a discord or something I could reach you on if you wouldn't mind?

Sure.
You can mail me!

harsh-aghara:
> Thanks for making the requested changes, looking pretty good!  We can simplify that helper method to be more performant though, so let's do that.  Should be good to go after that!

 @ajhollid   I have simplified the helper method.
Please review it.

harsh-aghara:
@ajhollid Just pushed an update to clear out the branch rot, sync up with develop. 

What i changed:
* **Rebase & #3530:** Got everything rebased and made sure the new Pushover channel plays nicely with the shared validation schemas and UI.
* **Build fix:** Caught a TS type mismatch in the client that broke the build after recent merges. Fixed it and verified with a local npm run build.
* **Review feedback:** Swapped in the lighter helper method you suggested and added credential sanitization to the controller. 
* All unit and integration tests are passing locally again.

Could you approve the CI workflow when you have a sec and give it another look?

harsh-aghara:
Hey @ajhollid, I just completed another rebase to resolve the fresh conflicts from the recently merged Twilio(#3534) and Pushover(#3530) PRs.

To make your review easier, here is what I've ensured in this update:

**Conflict Resolution:** Integrated the new Twilio/Pushover fields while keeping my Webhook Auth logic strictly isolated to avoid regressions.

**Maintainer Feedback:** Applied the performance-optimized sanitization (object destructuring) we discussed and ensured all credentials (including Webhook tokens) are stripped from API responses.

**Validation:** Verified both Client and Server builds locally, and all 995 tests are passing.

Since we were good to go after the last review, I'd love to get this merged before any more notification channels cause new conflicts. Let me know if there's anything else you need!

server/src/controllers/notificationController.ts
```diff
@@ -37,16 +37,26 @@ class NotificationController implements INotificationController {
...
+				success: true,
+				msg: "Notification sent successfully",
 				details: { service: SERVICE_NAME },
 			});
 		} catch (error) {
 			next(error);
 		}
 	};
 
+	private sanitizeNotification = (notification: any) => {
```
Br0wnHammer: Nit - type this as `Notification` instead of any.

server/src/service/infrastructure/notificationProviders/webhook.ts
```diff
@@ -6,6 +6,19 @@ import { getTestMessage } from "@/service/infrastructure/notificationProviders/u
 import got from "got";
 
 export class WebhookProvider extends NotificationProvider {
+	private buildAuthHeaders(notification: Partial<Notification>): Record<string, string> {
```
Br0wnHammer, line 9: There's already a `webhookProvider.test.ts` but nothing covers this. buildAuthHeaders is pure and security-sensitive, please add cases.

server/src/controllers/notificationController.ts
```diff
@@ -37,16 +37,26 @@ class NotificationController implements INotificationController {
...
+		if (!notification) return notification;
+		const { authPassword, authToken, accessToken, ...sanitized } = notification;
```
Br0wnHammer: This strips `accessToken` from every notification, but that's the live credential for Telegram/Twilio/Pushover/Matrix too, won't this leave their token fields empty and block re-saving on edit?

Br0wnHammer (changes requested):
Overall, a nice and clean implementation. The PR was easy to follow through. However, I have some questions and nitpicks. Please look at them.

server/src/controllers/notificationController.ts
harsh-aghara: Done!
Typed it to Notification.

server/src/service/infrastructure/notificationProviders/webhook.ts
harsh-aghara: Good call. I have added  test cases covering all the auth types and edge cases for missing credentials and empty strings.

server/src/controllers/notificationController.ts
harsh-aghara: Ah I didn't realize that.
I've removed accessToken from the destructuring list so it's no longer stripped

server/src/controllers/notificationController.ts
```diff
@@ -47,6 +48,12 @@ class NotificationController implements INotificationController {
 		}
 	};
 
+	private sanitizeNotification = (notification: Notification | null | undefined): Notification | null | undefined => {
+		if (!notification) return notification;
+		const { authPassword, authToken, ...sanitized } = notification;
```
Br0wnHammer, line 54: Since `sanitizeNotification` strips `authPassword/authToken` from responses, the edit form loads them empty, but `superRefine` still requires them, so editing a Basic/Bearer webhook seems to force re-entering credentials every time (and an empty submit would $set over the stored secret).

Br0wnHammer (changes requested):

server/src/controllers/notificationController.ts
harsh-aghara: I was focused on the sanitization that's why I didn't think through the "Edit" UX. It’s definitely a pain if a user has to re-enter their token just to rename a webhook. 
I can fix this by updating the Zod schema to allow empty strings on update and then tweak the service logic so it only $sets those fields if the user actually types something new. That way, leaving them blank would just keep the existing secret in the DB. Does that sound like the right path to you?

Br0wnHammer: @ajhollid what are your thoughts on this?

Br0wnHammer: @harsh-aghara, I was going through your proposed solution and I agree, leave blank to keep the secret is the right call. Just heads up that create and edit share the same validation, so loosening it would also let someone create a Basic auth webhook with no password. Probably want a separate lenient edit schema, and when authType changes, clear the old secret fields instead of skipping them so they don't linger in the DB.

harsh-aghara: @Br0wnHammer Thanks for the heads up.

I've just pushed a fix for this. I split out a lenient edit schema on both the client and server so users can safely leave secrets blank to keep their existing ones.

I also updated the service layer to explicitly $unset stale credentials if the authType (or the main channel type) changes, so we never leave lingering data in the DB.

Br0wnHammer: @harsh-aghara have your pushed the changes? I am not seeing any latest commits in the PR.

harsh-aghara: @Br0wnHammer Apologies for the confusion. I had to force-push to clean up an accidental commit that included unrelated files.

Br0wnHammer (approved):
LGMT! Great work on this @harsh-aghara 🚀

Br0wnHammer:
Hi @harsh-aghara,
We discussed your PR with the team, and we've decided that we can't merge it in its current form due to a number of related changes that still need to be addressed.

Given the scope of the code changes, it would be best to open a new PR and incorporate the following feedback there:

* The edit schema is largely a duplicate of the create schema (roughly 90% overlap), so we'd like to avoid maintaining two nearly identical implementations.
* There are a few type casts that undermine type safety and should be refactored.
* It looks like notification testing may no longer work as expected, since some fields are being stripped before processing.

Once these concerns are addressed in a new PR, we can take another look.
Thankyou again for all your hardwork! 🚀

PR #3460, opened 2026-04-04
fix: align server test suite with current source implementations

### LINKED ISSUE #3457: Test suite failures in server
## Description

Running the server test suite currently fails with multiple errors across notification, queue helper, and monitor service tests. These failures prevent reliable CI validation and block confidence in recent backend changes.

## Environment

- OS: Linux `6.8.0-85-generic`
- Runtime: Node.js (project default)
- Test runner: Jest (via `npm test`)
- Package: `server`

## Command Used

```bash
cd server
npm test
```

## Current Result

`3` test suites fail, with `8` failing tests and one module import/runtime error.

#### Observed failures

- `sendTestAlert` expected `true`, received `false`
- `sendMessage` expected `true`, received `false`
- `sendMessage` with no incident expected successful post, but mocked call payload is missing
- threshold/color tests fail because outbound payload assertions cannot read `gotPost.mock.calls[0]`

#### Error patterns

- `Expected: true / Received: false`
- `TypeError: undefined is not iterable (cannot read property Symbol(Symbol.iterator))`

#### Impact

Notification provider behavior cannot be validated, especially payload construction and Teams webhook delivery flow.

---

#### Observed failure

- Test suite fails before running tests.

#### Error pattern

- `SyntaxError: ... SuperSimpleQueueHelper.ts does not provide an export named 'default'`

#### Impact

Queue helper unit tests are fully blocked by import/export mismatch.

---

#### Observed failures

- `getMonitorsWithChecksByTeamId` expects `checks`, but response object contains a different field
- summary test assertion mismatch for monitor shape
- `getUptimeDetailsById` crashes due to missing mocked repository method

#### Error patterns

- `Expected path: "checks" / Received path: []`
- `Expected "checks" but received "recentChecks"`
- `TypeError: this.checksRepository.findByDateRangeAndMonitorId is not a function`

## Notes for Investigation

- Validate Teams provider test mocking strategy (`got` mocking and module import timing).
- Verify `SuperSimpleQueueHelper` export style used by implementation vs test import.
- Align monitor service tests with current response shape and repository interface methods.

gorkem-bwl:
## Summary
- Fixed all 3 failing test suites (teamsProvider, superSimpleQueueHelper, monitorService) that had fallen behind source code changes
- Extracted shared `createMockLogger` helper to reduce duplication across test files
- Fixed Jest config `@/utils` path mapping that prevented module resolution

## What changed

**teamsProvider.test.ts**: Fixed `got` ESM mock wiring — mock is now declared before dynamic import so it intercepts correctly. Fixed `getTestMessage` assertion and incident URL path.

**superSimpleQueueHelper.test.ts**: Changed default → named import, updated constructor from object-based to 16 positional params, renamed `getMonitorJob` → `getHeartbeatJob`, added all 8 missing dependencies.

**monitorService.test.ts**: Removed 4 stale constructor params (`db`, `stringService`, `papaparse`, `errorService`), added 2 missing ones (`geoChecksRepository`, `incidentsRepository`), fixed renamed repository methods (`findLatestByMonitorIds`, `findByDateRangeAndMonitorId`), updated assertions from `checks` → `recentChecks`.

**jest.config.ts**: Consolidated `@/utils` path mapping — the `AppError`-specific override was redundant and the generic rule mapped to `.js` instead of `.ts`.

## Test plan
- [x] `npm test` passes: 3 suites, 19 tests, 0 failures

Closes #3457

### DIFF
--- a/server/test/helpers/createMockLogger.ts
+++ b/server/test/helpers/createMockLogger.ts
@@ -1,8 +1,12 @@
 import { jest } from "@jest/globals";
 
 export const createMockLogger = () => ({
+	serviceName: "test",
 	info: jest.fn(),
 	warn: jest.fn(),
 	error: jest.fn(),
 	debug: jest.fn(),
+	cacheLog: jest.fn(),
+	getLogs: jest.fn().mockReturnValue([]),
+	buildLogEntry: jest.fn().mockReturnValue({ level: "info", timestamp: new Date().toISOString() }),
 });
--- a/server/test/helpers/createMockLogger.ts
+++ b/server/test/helpers/createMockLogger.ts
@@ -1,0 +1,8 @@
+import { jest } from "@jest/globals";
+
+export const createMockLogger = () => ({
+	info: jest.fn(),
+	warn: jest.fn(),
+	error: jest.fn(),
+	debug: jest.fn(),
+});
--- a/server/test/monitorService.test.ts
+++ b/server/test/monitorService.test.ts
@@ -109,6 +109,14 @@ describe("MonitorService", () => {
 			expect(result).toMatchObject({ count: 2 });
 			expect(result.monitors).toHaveLength(2);
 			expect(result.monitors[0]).toHaveProperty("recentChecks");
+			expect(result.monitors[0].recentChecks.length).toBeGreaterThan(0);
+			expect(result.monitors[0].recentChecks[0]).toEqual(
+				expect.objectContaining({
+					responseTime: expect.any(Number),
+					status: expect.any(Boolean),
+					message: expect.any(String),
+				})
+			);
 		});
 	});
 
--- a/server/test/monitorService.test.ts
+++ b/server/test/monitorService.test.ts
@@ -1,6 +1,13 @@
 import { jest } from "@jest/globals";
 import { MonitorService } from "../src/service/business/monitorService.ts";
-import type { IChecksRepository, IMonitorStatsRepository, IMonitorsRepository, IStatusPagesRepository } from "../src/repositories/index.ts";
+import type {
+	IChecksRepository,
+	IGeoChecksRepository,
+	IIncidentsRepository,
+	IMonitorStatsRepository,
+	IMonitorsRepository,
+	IStatusPagesRepository,
+} from "../src/repositories/index.ts";
 
 const createMonitorsRepositoryMock = () =>
 	({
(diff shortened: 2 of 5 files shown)


server/test/monitorService.test.ts
```diff
@@ -31,82 +33,82 @@ const createStatusPagesRepositoryMock = () =>
...
-		logger: { info: jest.fn(), error: jest.fn(), warn: jest.fn() },
-		errorService: {
-			createAuthorizationError: jest.fn(() => new Error("unauthorized")),
-			createServerError: jest.fn(() => new Error("server")),
-			createBadRequestError: jest.fn(() => new Error("bad request")),
-			createNotFoundError: jest.fn(() => new Error("not found")),
-		},
+		} as any,
+		emailService: { buildEmail: jest.fn(), sendEmail: jest.fn() } as any,
+		logger: { info: jest.fn(), error: jest.fn(), warn: jest.fn() } as any,
```
akashmannil, line 64: there is already a mockLogger, is it possible to use that here as well

server/test/helpers/createMockLogger.ts
```diff
@@ -0,0 +1,8 @@
+import { jest } from "@jest/globals";
+
+export const createMockLogger = () => ({
+	info: jest.fn(),
+	warn: jest.fn(),
+	error: jest.fn(),
+	debug: jest.fn(),
```
akashmannil, line 8: since we're creating a mock logger, isnt it ok to include other `ILogger `interface members
```
info: jest.fn(),
warn: jest.fn(),
error: jest.fn(),
debug: jest.fn(),
cacheLog: jest.fn(),
getLogs: jest.fn().mockReturnValue([]),
buildLogEntry: jest.fn().mockReturnValue({ level: "info", timestamp: "", message: "" }),
```

server/test/monitorService.test.ts
```diff
@@ -31,82 +33,82 @@ const createStatusPagesRepositoryMock = () =>
...
-			expect(result.monitors[0]).toHaveProperty("checks");
-			expect(result.monitors[0].checks.length).toBeGreaterThan(0);
-			expect(result.monitors[0].checks[0]).toEqual(
-				expect.objectContaining({
-					responseTime: expect.any(Number),
-					status: expect.any(Boolean),
-					message: expect.any(String),
-				})
-			);
+			expect(result.monitors[0]).toHaveProperty("recentChecks");
```
akashmannil, line 111: is it ok that we're not checking received structure in detail(or most important ones) like before. or we're ok with checking it like this?

server/test/monitorService.test.ts
```diff
@@ -155,31 +154,19 @@ describe("MonitorService", () => {
...
 				url: "https://example.com",
 				isActive: true,
-				alertThreshold: 5,
-				cpuAlertThreshold: 5,
-				memoryAlertThreshold: 5,
-				diskAlertThreshold: 5,
-				tempAlertThreshold: 5,
-				selectedDisks: [],
-				notifications: [],
-				group: null,
```
akashmannil, line 174: we are still using the removed properties right? should we remove this?

akashmannil (commented):
My opinion: If the goal is **only to fix the failing suites**, many of these changes (fixture trimming, assertion weakening, the as any casts) along with fewer fixes should do the deed.

If the goal **includes improving coverage**, the empty geoChecksRepository and incidentsRepository mocks need real methods (the service calls deleteByMonitorId on both in deleteMonitor), and deleteMonitor itself has no test at all. That should be a separate PR with a broader scope/title i guess.

Also. i had to install `cross-env` locally to get the test run successfully. do everyone have this issue?

ajhollid (approved):

Issue #3459, opened 2026-04-04
Seems like Uptime details page shows % instead of actual values


dmythro:
Uptime list shows actual response times in the chart. But details page shows values 0-100.

**Expected behavior**

If page response is 321ms I expect to see these actual values in details page.

**Screenshots**

this website in the list:

<img width="294" height="171" alt="Image" src="https://github.com/user-attachments/assets/c6c44662-fc7e-4cfd-acf5-08cac0017806" />

this website in details (same for Uptime):

<img width="1307" height="409" alt="Image" src="https://github.com/user-attachments/assets/df1c51f6-6086-4d3a-8bdf-4aeb14e900ee" />

**Additional context**

API has `responseTime` and  `originalResponseTime`, and there `originalResponseTime` is a correct value and `responseTime` seem like a percentage (0-100).

AlizaRK:
Hi, can I take this issue?

ajhollid:
Looks to me that those are the normalized values that are displaying in the tool tip. 

All response times are normalized between 0 and 100 for display purposes. 

This should be a simple fix as the original response time is also included in the api response.

dmythro:
Seems like infrastructure charts don't even have tooltips — would be cool to fix that too, so we could see the actual values at some point in time as well, like it works in uptime.

ajhollid:
> Seems like infrastructure charts don't even have tooltips — would be cool to fix that too, so we could see the actual values at some point in time as well, like it works in uptime.

Good idea, should be relatively straight forward

AlizaRK:
Hi, even though I haven’t been assigned this issue, I went ahead and created the fixes and changes based on this conversation to get familiar with the Checkmate codebase (I am new here). Please have a look and let me know if it works!

[Here's the PR.](https://github.com/bluewave-labs/Checkmate/pull/3467)

dmythro:
I hope it gets processed soon, I also planned to check it out and make a PR :)

gorkem-bwl:
Related: #3380 — same root cause. The Uptime details page sends `normalize=true`, and `server/src/utils/dataUtils.ts:74-88` rescales `avgResponseTime` into a fixed range (storing the real value as `originalAvgResponseTime`). `client/src/Components/monitors/charts/HistogramDetails.tsx:71` then renders that normalized value as `${responseTime} ms`, which looks like a percentage. The AVG tile uses raw `monitorStats.avgResponseTime` (correct), so the mismatch is purely in the histogram.

PR #3470, opened 2026-04-07
feat: adding mobile/desktop pagespeed strategy

### LINKED ISSUE #3453: PageSpeed: add mobile/desktop strategy option
The PageSpeed monitor currently calls the Google PageSpeed API without a `strategy` parameter, which defaults to `desktop`. There is no way to monitor mobile performance scores.

Mobile PageSpeed scores are typically lower and more representative of real-world user experience, since most web traffic is mobile. The ability to choose between `mobile` and `desktop` (or an option to monitor both) would make PageSpeed monitoring significantly more useful.

## Suggested implementation

The Google PageSpeed API accepts a `strategy` parameter with values `mobile` or `desktop`:

```
https://pagespeedonline.googleapis.com/pagespeedonline/v5/runPagespeed?url=...&strategy=mobile
```

Options:
1. Add a `strategy` field to the PageSpeed monitor configuration (default: `desktop` for backward compatibility)
2. Or allow creating separate monitors for mobile and desktop

## Reference

- [PageSpeed API docs — strategy parameter](https://developers.google.com/speed/docs/insights/v5/reference/pagespeedapi/runpagespeed)
- Current implementation: `server/src/service/infrastructure/network/PageSpeedProvider.ts`

akashmannil:
Adding Pagespeed desktop/mobile strategy.

Fixes #3453 

image

image

image

```Javascript
<div>{t('add')}</div>
```


dmythro:
Thanks for jumping on this so fast!

Couple things Claude spotted and I want to mention, might make sense:

  1. Strategy is immutable after creation — disabled={isEditMode} on the Select. If you want to switch a
  monitor from desktop to mobile, you must delete and recreate it. The server-side editMonitorBodyValidation
  does allow strategy though, so the API permits changes but the UI blocks it. This inconsistency could cause
  confusion.
  2. strategy column added to ALL monitors, not just pagespeed. Every http, ping, docker, etc. monitor gets
  strategy = 'desktop'.

I think editing should be allowed - so I can just switch existing monitors to mobile without re-adding new ones and without paused old monitors for historical data. And strategy should be empty by default, only set for pagespeed.

client/src/Components/design-elements/StrategyBadge.tsx
```diff
@@ -0,0 +1,35 @@
...
+		<Typography
+			variant={variant}
+			px={SPACING.XXL}
+			py={SPACING.SM}
+			textTransform="capitalize"
+			sx={{
+				borderRadius: theme.shape.borderRadius,
+				backgroundColor: theme.palette.secondary.light,
+				color: theme.palette.text.primary,
+			}}
```
ajhollid: Why are we using a combination of native props and `sx` here?  Why not use all native props?  `Typography` has native props for all of these values I believe.

server/src/repositories/monitors/TimescaleMonitorsRepository.ts
```diff
@@ -1033,6 +1037,7 @@ export class TimescaleMonitorsRepository implements IMonitorsRepository {
 		selectedDisks: row.selected_disks ?? [],
 		gameId: row.game_id ?? undefined,
 		grpcServiceName: row.grpc_service_name ?? undefined,
+		strategy: (row.strategy as Monitor["strategy"]) ?? undefined,
```
ajhollid: No casting

akashmannil:
Thanks for the review @dmythro. PFB thought process for each of the mentioned pointers.

1. Once we change the strategy, our history gets invalidated (our previous data points to Desktop, but later changed to mobile gives different stats - faulty assurances). Thats why we're making it immutable.
2. Since http pagespeed monitor is the only option with reliable pagespeed accessibility metrics (which is expected to be varying based on device pref) with proper strategy dependencies, we're adding only an option for that.

server/src/db/migration/timescaledb/0022_add_strategy_to_monitors.ts
```diff
@@ -0,0 +1,15 @@
+import type { Pool } from "pg";
+
+export const addStrategyToMonitors = async (pool: Pool) => {
+	await pool.query(`
+		ALTER TABLE monitors
+		ADD COLUMN IF NOT EXISTS strategy TEXT DEFAULT 'desktop';
```
ajhollid: Default here should be `null` not `desktop`, it doesn't make sense for an `http` type monitor to have a strategy type.

server/src/db/models/Monitor.ts
```diff
@@ -330,6 +330,11 @@ const MonitorSchema = new Schema<MonitorDocument>(
 			type: String,
 			default: "",
 		},
+		strategy: {
+			type: String,
+			enum: PageSpeedStrategies,
+			default: DefaultPageSpeedStrategy,
```
ajhollid: there should be no default value here, only pagespeed type monitors should have a strategy

client/src/Components/design-elements/StrategyBadge.tsx
```diff
@@ -0,0 +1,35 @@
...
+	if (!strategy) {
+		return null;
+	}
+
+	return (
+		<Typography
+			variant={variant}
+			px={SPACING.XXL}
+			py={SPACING.SM}
+			textTransform="capitalize"
```
ajhollid: I don't think we want this text transform since the string is provided by the translation file.  This will likely cause headaches for non-english languages

ajhollid (changes requested):
Some minor issues to resolve, but it generally looks good.

Allowing editing of the strategy type is a design decision; we can defer to @gorkem-bwl  on that one.

My personal opinion is to agree with you that it should be immutable once set, as the historical data will be nonsensical if you switch strategies.

The precedent has been set for allowing these types of history-breaking changes, though, so I'll leave this one to @gorkem-bwl  to make the call.

gorkem-bwl:
> Some minor issues to resolve, but it generally looks good.
> 
> Allowing editing of the strategy type is a design decision; we can defer to @gorkem-bwl on that one.
> 
> My personal opinion is to agree with you that it should be immutable once set, as the historical data will be nonsensical if you switch strategies.
> 
> The precedent has been set for allowing these types of history-breaking changes, though, so I'll leave this one to @gorkem-bwl to make the call.

My take here is that user knows what he's doing, so we could agreeable for the user to edit that section.

dmythro:
As a user I'd prefer to edit this. At least I hope you won't block editing strategy via API.

I have many monitors and creating new for each and pause old ones is much more hassle than just switch to mobile and know that older data was desktop is fine with me.

server/src/validation/monitorValidation.ts
```diff
@@ -73,6 +73,7 @@ export const createMonitorBodyValidation = z.object({
 	matchMethod: z.union([z.enum(MonitorMatchMethods), z.literal("")]).optional(),
 	gameId: z.union([z.string(), z.literal("")]).optional(),
 	grpcServiceName: z.union([z.string(), z.literal("")]).default(""),
+	strategy: z.enum(PageSpeedStrategies).default(DefaultPageSpeedStrategy),
```
ajhollid: There shouldn't be a default value for this, this should be an optional field.  Otherwise a `strategy` wil be set for non-pagespeed type monitors, which is not correct.

ajhollid (changes requested):
Looks pretty good.  There's an error in the montor validation that will set `strategy` to non-pagespeed type monitors that needs to be fixed, I think we're done after that.

akashmannil:
@ajhollid Default values removed and set only for pagespeed monitors. Please do let me know if this looks ok. Thanks!

dmythro:
Hi! I've seen a release recently and was mainly waiting for this one to come up but sadly it is not. Maybe I could help somehow? This and #3459 are mainly two things I am really looking forward to :)

ajhollid (approved):
I resolved conflicts and pushed some minor changes in the interest of expediency.

Thank you for working on this issue, I think we're good to go!

Issue #3478, opened 2026-04-08
Matrix notifications are not going through


ad2ien:
**Describe the bug**

Matrix notification are not working.

**To Reproduce**

1. Create a new notification
2. Select Matrix
3. Setup homeserver roomId and token
4. Click on Test button
5. See error

**Expected behavior**

If the configuration is valid it should succeed

**Desktop (please complete the following information):**
 - Firefox 149.0
 - Checkmate 3.5.1

**Additional context**

The matrix spec are not fully implemented. I looked into it and will do PR with all the information.

Thanks!

PR #3479, opened 2026-04-08
Fix matrix notifications

ad2ien:
## Summary

I tried to set up a matrix notification but the test message wasn't going through : #3478  This PR fixes it.

## What changed

- use `put` instead of `post` : [Matrix spec : send API](https://spec.matrix.org/v1.18/client-server-api/#put_matrixclientv3roomsroomidsendeventtypetxnid)
- Add transaction id : [Matrix spec : transaction ID](https://spec.matrix.org/v1.18/client-server-api/#transaction-identifiers)
- use access_token in header like it's recommended here : [access token](https://spec.matrix.org/v1.18/client-server-api/#using-access-tokens)
- factorize api call
- add the response message in the warning log

## Tests

- ✅ cd server && `npm run test` (thanks to #3460 ❤️)
- ✅ `npm run lint` `npm run format` 
- ✅ tested with a local [Synapse](https://github.com/element-hq/synapse) / [element](https://github.com/element-hq/element-web)
- ✅ also tested with a distant [Conduit](https://github.com/timokoesters/conduit) server

```Javascript
<div>{t('add')}</div>
```

### DIFF
--- a/server/src/service/infrastructure/notificationProviders/matrix.ts
+++ b/server/src/service/infrastructure/notificationProviders/matrix.ts
@@ -5,6 +5,7 @@ import type { AlertMatrixPayload, Notification } from "@/types/index.js";
 import type { NotificationMessage } from "@/types/notificationMessage.js";
 import { getTestMessage } from "@/service/infrastructure/notificationProviders/utils.js";
 import { ILogger } from "@/utils/logger.js";
+import { randomUUID } from "crypto";
 
 export class MatrixProvider implements INotificationProvider {
 	private logger: ILogger;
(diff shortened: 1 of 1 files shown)


server/src/service/infrastructure/notificationProviders/matrix.ts
```diff
@@ -181,4 +167,9 @@ export class MatrixProvider implements INotificationProvider {
 	private escapeHtml(text: string): string {
 		return text.replace(/&/g, "&amp;").replace(/</g, "&lt;").replace(/>/g, "&gt;").replace(/"/g, "&quot;").replace(/'/g, "&#039;");
 	}
+
+	private isNotificationValid(notification: Partial<Notification>): boolean {
+		const { homeserverUrl, accessToken, roomId } = notification;
+		return !!homeserverUrl && !!accessToken && !!roomId;
+	}
```
ajhollid: I don't think a helper method is needed for this, the logic is straightforward and this will never be reused elsewhere.

Let's move it back inline for readability's sake.

server/src/service/infrastructure/notificationProviders/matrix.ts
```diff
@@ -14,64 +14,50 @@ export class MatrixProvider implements INotificationProvider {
...
 		const body = {
 			msgtype: "m.text",
 			body: plainText,
 			format: "org.matrix.custom.html",
 			formatted_body: htmlText,
 		};
+		return this.sendMessageWithBody(notification, body);
+	};
+
+	sendMessageWithBody = async (notification: Partial<Notification>, body: unknown): Promise<boolean> => {
```
ajhollid: This should be a private method

server/src/service/infrastructure/notificationProviders/matrix.ts
```diff
@@ -14,64 +14,50 @@ export class MatrixProvider implements INotificationProvider {
...
+		const { homeserverUrl, accessToken, roomId } = notification;
+		const currentTimestamp = new Date().getTime().toString();
```
ajhollid: It appears you are using a timestamp as an identifier; this is not idempotent.   It is not out of the realm of possiblity that two notifications could be sent at the same time, causing a collision here.

UUID would be more appropriate.

ajhollid (changes requested):
In general it looks good 👍   I don't know anything about Matrix, so I'm taking your word for the implementation validity.

There are a couple of minor issues here and there, once resolved we should be good to go.

server/src/service/infrastructure/notificationProviders/matrix.ts
ad2ien: yes 👍  I was worrying about adding a  new dependency but we can use randomUUID() from crypto.

ajhollid (approved):
Looks good to me, thank you for the revisions and for your contribution to the project!🚀

PR #3511, opened 2026-04-14
Fix Open Site action for internal monitors

### LINKED ISSUE #3434: "Open site" action is unusable for monitors with internal URLs
### Description

The "Open site" action in the monitor actions menu opens the monitor's `url` directly in the browser. This works for HTTP monitors targeting public URLs, but is broken for monitors that target internal/private endpoints — specifically `hardware` and `port` type monitors, and any `http` monitor pointed at an internal health endpoint.

### Steps to reproduce

1. Create a hardware monitor pointing to a Capture agent (e.g. `http://capture:59232/api/v1/metrics`)
2. Or create an HTTP monitor pointing to an internal health endpoint (e.g. `http://traefik:8080/ping`)
3. Click the gear/actions icon on the monitor row
4. Click "Open site"

### Expected behavior

Either:
- The "Open site" action is hidden for monitors where the URL is not externally accessible (e.g. `hardware`, `port` types, or URLs using internal hostnames)
- Or monitors support a separate optional `externalUrl` / `siteUrl` field that "Open site" uses when present, falling back to `url` when not set

### Actual behavior

The browser attempts to navigate to the internal URL (e.g. `http://capture:59232/api/v1/metrics`), which is unresolvable from the user's machine. This results in a browser error page.

### Affected monitor types

| Type | Why it's broken |
|------|----------------|
| `hardware` | Always targets an internal Capture agent |
| `port` | URL is just a hostname (e.g. `postgres`), no protocol |
| `http` (internal) | When pointed at internal health endpoints like `http://traefik:8080/ping` instead of a public URL |

### Suggested fix

Add an optional `siteUrl` field to the monitor schema. When populated, "Open site" navigates to `siteUrl` instead of `url`. When empty, either fall back to `url` (current behavior) or hide the action.

This cleanly separates the monitoring target from the user-facing link, which is a common need when monitoring infrastructure behind a reverse proxy or on a private network.

### Environment

- Checkmate version: v3.5.1 (commit `9628aa02`)
- Deployment: Docker (backend-mono image)

ZeinebMegaadi:
## Describe your changes

Fixed the "Open Site" action in the monitor actions menu to prevent opening internal/private endpoints that are not accessible from the user's browser.

This includes:
- Hardware monitors (Capture agent targets)
- Port monitors (non-URL services)
- HTTP monitors pointing to internal services (e.g. traefik, localhost, internal Docker DNS)

The action is now hidden when the monitor target is not externally accessible.

Fixes #3434 

```Javascript
<div>{t('add')}</div>
```
- [x] There are two database implementations (MongoDB and PostgreSQL TimescaleDB) and I have taken care of them appropriately.
Open Site button stopped showing for internal links: 
Screenshot 2026-04-14 192543

### DIFF
--- a/client/src/Pages/Uptime/Monitors/Components/UptimeMonitorsTable.tsx
+++ b/client/src/Pages/Uptime/Monitors/Components/UptimeMonitorsTable.tsx
@@ -69,20 +69,20 @@ export const MonitorTable = ({
 
 	// ONLY CHANGE: detect internal/private URLs (issue requirement)
 	const isInternalUrl = (url: string) => {
-  try {
-    const { hostname } = new URL(url);
-    return (
-      hostname === "localhost" ||
-      hostname === "127.0.0.1" ||
-      hostname === "::1" ||
-      hostname.endsWith(".local") ||
-      hostname.endsWith(".internal") ||
-      !hostname.includes(".")  
-    );
-  } catch {
-    return true; // unparseable = internal (port monitors with no protocol)
-  }
-};
+		try {
+			const { hostname } = new URL(url);
+			return (
+				hostname === "localhost" ||
+				hostname === "127.0.0.1" ||
+				hostname === "::1" ||
+				hostname.endsWith(".local") ||
+				hostname.endsWith(".internal") ||
+				!hostname.includes(".")
+			);
+		} catch {
+			return true; // unparseable = internal (port monitors with no protocol)
+		}
+	};
 
 	const getActions = (monitor: Monitor): ActionMenuItem[] => {
 		const actions: ActionMenuItem[] = [];
@@ -154,7 +154,11 @@ export const MonitorTable = ({
 
 	const getHeaders = (chartType: string) => {
 		const renderSortIcon = (isActive: boolean) => (
-			<Box width={16} display="inline-flex" justifyContent="center">
+			<Box
+				width={16}
+				display="inline-flex"
+				justifyContent="center"
+			>
 				{isActive ? (
 					sortOrder === "asc" ? (
 						<ArrowUp size={16} />
--- a/client/src/Pages/Uptime/Monitors/Components/UptimeMonitorsTable.tsx
+++ b/client/src/Pages/Uptime/Monitors/Components/UptimeMonitorsTable.tsx
@@ -21,6 +21,8 @@ import type { Monitor } from "@/Types/Monitor";
 import type { ActionMenuItem } from "@/Components/actions-menu";
 import type { RootState } from "@/Types/state";
 
+const NON_BROWSER_TYPES = ["hardware", "docker"];
+
 export const MonitorTable = ({
 	monitors,
 	refetch,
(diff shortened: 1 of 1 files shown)


ajhollid (changes requested):
On further thought, I don't think the premise of this change is really correct.

If my client and server are both running on the same machine, which is not an unreasonable setup, I may indeed want to open internal URLs in my browser.  

I think the only discriminant for opening URLs in the browser should be to exclude types that don't display anything useful in the browser, like docker monitors or infrastrucutre monitors.

ZeinebMegaadi:
That makes sense yeah, thanks for the clarification!
I was trying to prevent cases where the link wouldn’t work from the browser but I agree that internal URLs can still be valid depending on the setup. I’ll update the logic to only hide the action for monitor types that don’t produce a meaningful browser view (like: infrastructure/docker), instead of filtering based on the URL.

ajhollid:
> That makes sense yeah, thanks for the clarification! I was trying to prevent cases where the link wouldn’t work from the browser but I agree that internal URLs can still be valid depending on the setup. I’ll update the logic to only hide the action for monitor types that don’t produce a meaningful browser view (like: infrastructure/docker), instead of filtering based on the URL.

Yeah I understand the intent behind it, but let's keep it flexible since we don't want to assume too much about the user's setup :+1:   Sounds like you've got it in hand, we'll get this merged in when the changes are pushed.

ZeinebMegaadi:
Is this better?

ZeinebMegaadi:
@ajhollid

client/src/Pages/Uptime/Monitors/Components/UptimeMonitorsTable.tsx
```diff
@@ -21,6 +21,8 @@ import type { Monitor } from "@/Types/Monitor";
 import type { ActionMenuItem } from "@/Components/actions-menu";
 import type { RootState } from "@/Types/state";
 
+const NON_BROWSER_TYPES = ["hardware", "docker"];
```
ajhollid, line 24: What about other non-browser types like gRPC, Docker, etc?  I don't know exactly which ones are not browser-friendly, but at least some of them are not.

ajhollid, line 24: This should be types as an `Array<MonitorType>`.  We want type safety everywhere

ajhollid (changes requested):
Change looks good, but is incomplete.  Here is the full list of montior types:

```
export const MonitorTypes = ["http", "ping", "pagespeed", "hardware", "docker", "port", "game", "grpc", "websocket", "unknown"] as const;
```

Let's figure out which ones should be in the non browser type or not.

client/src/Pages/Uptime/Monitors/Components/UptimeMonitorsTable.tsx
ZeinebMegaadi: Good point. I initially only included hardware and docker, but you're right that other types like gRPC and port-based checks are also not browser-friendly.  Also I'm assuming that ping (ICMP) isn’t browser-compatible and game likely uses non-HTTP protocols, but I want to confirm before adding them.

ajhollid:
Closing as inactive

PR #3550, opened 2026-04-19
feat: group maintenance windows in table view

### LINKED ISSUE #3548: Group maintenance windows by name in the table when created for multiple monitors
Basically when creating a maintenance window for multiple monitors, the app creates one database record per monitor. This results in multiple rows with the same name, same schedule, and same settings appearing in the maintenance windows table , one per monitor. This is confusing and hard to manage, especially at a larger scale.

The maintenance windows table should display one row per logical window (grouped by name/schedule). Each row should be expandable to show the individual monitors it applies to, and allow managing them individually from there.

**Suggested Fix**

Either:

Group records client-side by name + start + end when displaying the table, or
Return pre-grouped data from the API

**Additional context**
Here is a screenshot of how many windows with same name and no grouping looks like:

Image

ayshaL:
t## Describe your changes

The maintenance windows table previously displayed one row per database record, resulting in multiple duplicate rows when a window was created for multiple monitors (one DB record is created per monitor). This was confusing and hard to manage at scale.

This PR implements client-side grouping of maintenance window records by `name + start + end` so that logically identical windows are collapsed into a single row. Each grouped row is expandable to show the individual monitors it applies to, with per-monitor actions (configure, pause, delete) available in the expanded view.

Fixes #3548

```Javascript
<div>{t('add')}</div>
```
- [ ] There are two database implementations (MongoDB and PostgreSQL TimescaleDB) and I have taken care of them appropriately.

grp-maintenance


gorkem-bwl:
Can you please shoot a video or a full page? I am not sure I get it from the screenshot you pasted.

ayshaL:
Sure! here's a short screen recording.

https://github.com/user-attachments/assets/781be2da-0352-4702-b10f-9a75eaf73f66

gorkem-bwl:
Is there a way to understand if a row is collapsible or not? Currently I see all rows having the same style.

ayshaL:
All rows are expandable even those with a single monitor since it shows the monitor name and its actions (configure, pause, delete). 
I can add a chevron icon to indicate that it's expandable.

client/src/Pages/Maintenance/MaintenanceWindowTable.tsx
```diff
@@ -177,16 +231,46 @@ export const MaintenanceWindowTable = ({
...
 		<Box>
 			<Table
 				headers={getHeaders()}
-				data={maintenanceWindows}
-				onRowClick={(row) => navigate(`/maintenance/create/${row.id}`)}
+				data={grouped}
+				expandableRows={true}
+				onExpandChange={(id) => setExpandedId(id as string | null)}
+				renderExpandedContent={(row) => (
+					<Box sx={{ pl: 2, pd: 2 }}>
```
Br0wnHammer, line 247: `pd` is not a valid MUI shorthand and is silently ignored. You probably meant `pb` (padding-bottom). Easy fix.

client/src/Pages/Maintenance/MaintenanceWindowTable.tsx
```diff
@@ -177,16 +231,46 @@ export const MaintenanceWindowTable = ({
 		setPage(0);
 	};
 
+	const grouped = groupWindows(maintenanceWindows);
```
Br0wnHammer: Grouping happens here on the client after the server has already paginated. That breaks pagination correctness in two ways:

1. Group splitting at page boundaries. A window that spans N monitors can have some of those records on page 1 and the rest on page 2. After grouping, the same group renders on both pages — the duplication this PR set out to remove reappears at every page edge.
2. Footer count is wrong. Below at L273–275, `count={maintenanceWindowCount}` is the server's count of raw records, but `itemsOnPage={grouped.length}` is groups. The `1–N of M` label is mixing two different units, so users will see e.g. `1–3 of 12` while only 3 group-rows are visible.

Br0wnHammer: Wrap `groupWindows(maintenanceWindows)` in `useMemo(... , [maintenanceWindows])`. Same for the monitors lookup at L262 — turn `monitorsList` into a `Map<id, Monitor>` once via `useMemo`, then `.get(monitor.monitorId)` in the row. Currently every row re-scans the entire monitor list on every render, and the grouping is rebuilt for unrelated state updates (expand/dialog/rowsPerPage)

Br0wnHammer (changes requested):
I like to approach taken. However, some bugs and nits are needed to be taken care of. Please see my comments.

ayshaL:
I addressed the `useMemo` and `MonitorMap` feedback, grouping is now using `Map.get()` instead of find().
For the pagination issue however (page boundary and wrong footer count), these can't be fully solved client-side because the root cause is that grouping happens after the server has already paginated.
So a proper fix would require a server-side grouping. I can handle that in another PR if you'd like.

ajhollid:
Hi @ayshaL , this issue has been resolved at an architectural level, so we can safely close this PR.  Thank you for taking the time to work on the project, we'd love to have your contributions in other areas!

PR #3557, opened 2026-04-20
fix: hide register tip on login and redirect to setup when no super admin

### LINKED ISSUE #3556: [Bug / Improvement] Disable register tip on login page
**Describe the bug**
After the initial Checkmate setup, users can still see the register tip below the login button on the login page. This confused me into thinking that anyone could register on the site which is actually not true.

**To Reproduce**
1. Open the demo page for example 

**Expected behavior**
The register tip should not be visible on the login page at all, and on initial setup (no super admin registered) Checkmate should redirect you to the register page until a super admin registers. Additionally, in the "description" of the register page, you could mention that this page is only available during initial setup and new users can be created / invited in the team settings.

gorkem-bwl:
Closes #3556

## Summary
- Removes the "Register here" link from the login page — it previously misled users into thinking public registration was open.
- Adds a redirect from `/login` to `/register` when no super admin exists, so fresh installs land on the setup page.
- Adds an info alert on the register page clarifying it's only available during initial setup; additional users are invited from team settings. Hidden for the invite-token flow.
- Extracts a shared `useSuperAdminRedirect` hook to dedupe the identical fetch/redirect logic that Login and Register were both doing.


ajhollid (approved):

SkyKingPX:
Fire, thank you very much

PR #3534, opened 2026-04-16
feat: add Twilio SMS notification channel

egeoztass:
## Summary
- Adds **Twilio SMS** as a new notification channel (#3095)
- Users can configure their Twilio **Account SID**, **Auth Token**, **From number** (Twilio), and **To number** (recipient) to receive SMS alerts
- Follows the exact same provider pattern as Pushover, Telegram, Discord, etc.

### Backend (7 files)
- **New:** `twilio.ts` provider — sends via Twilio REST API with HTTP Basic Auth
- Registered in notification types, MongoDB model enum, Zod validation
- Wired into `NotificationsService` (both `send` and `sendTestNotification` routing)
- Instantiated in `services.ts`, exported from service index

### Frontend (5 files)
- Added `"twilio"` to `NotificationChannels` type
- Added Zod validation schema (all 4 fields required)
- Added form defaults in `useNotificationForm` hook
- Added form UI with 4 fields: Account SID, Auth Token, From number, To number
- Added English i18n keys (other languages via PoEditor)

### Tests (2 files)
- **New:** Twilio provider test suite — **18 tests** covering all paths
- Updated `NotificationsService` test constructor with `twilioProvider`

### Field mapping
Uses existing Notification schema fields — no schema migration needed:
- `address` → Account SID
- `accessToken` → Auth Token
- `homeserverUrl` → From number (Twilio number)
- `phone` → To number (recipient)

## Test plan
- [ ] Run `npm test -- --testPathPatterns="twilioProvider"` — all 18 tests pass
- [ ] Run `npm test -- --testPathPatterns="notificationsService"` — all 24 tests pass
- [ ] Start dev server, navigate to Notifications → Create
- [ ] Select "twilio" from channel type dropdown
- [ ] Verify form shows Account SID, Auth Token, From number, and To number fields
- [ ] Save and verify notification appears in the list
- [ ] Edit existing Twilio notification and verify fields are pre-populated

Closes #3095

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3530, opened 2026-04-16
feat: add Pushover notification channel

egeoztass:
## Summary
- Adds **Pushover** as a new notification channel (#3445)
- Users can configure their Pushover **Application API token** and **User key** to receive push notifications on all their devices
- Follows the exact same provider pattern as Telegram, Discord, Slack, etc.

### Backend (7 files)
- **New:** `pushover.ts` provider — sends via `POST https://api.pushover.net/1/messages.json`
- Registered in notification types, MongoDB model enum, Zod validation
- Wired into `NotificationsService` (both `send` and `sendTestNotification` routing)
- Instantiated in `services.ts`, exported from service index

### Frontend (5 files)
- Added `"pushover"` to `NotificationChannels` type
- Added Zod validation schema (user key + app token required)
- Added form defaults in `useNotificationForm` hook
- Added form UI with two fields: Application API token and User key
- Added English i18n keys (other languages via PoEditor)

### Tests
- Full unit test suite: **15 tests, 100% coverage** on the provider
- Follows existing contract test pattern (`testNotificationProviderContract`)
- Covers: success paths, missing fields, error handling, thresholds, incidents, optional sections

## Test plan
- [ ] Run `npm test -- --testPathPatterns="pushoverProvider"` — all 15 tests pass
- [ ] Start dev server, navigate to Notifications → Create
- [ ] Select "pushover" from channel type dropdown
- [ ] Verify form shows "Application API token" and "User key" fields
- [ ] Enter valid Pushover credentials and click "Test" to receive a test notification
- [ ] Save and verify notification appears in the list
- [ ] Edit existing Pushover notification and verify fields are pre-populated

Closes #3445

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3559, opened 2026-04-20
fix: mask sensitive fields in notification API responses

egeoztass:
## Summary
Follow-up to #3534 as suggested by @ajhollid — masks `accessToken` and `accountSid` in all notification API responses to prevent plaintext exposure of credentials.

## Changes

Single file change: `server/src/controllers/notificationController.ts`

- **Mask in responses**: All GET and mutating endpoints now return masked sensitive fields (e.g., `ACxxxxxxxx` → `ACxx****`)
- **Strip on edit**: When updating a notification, masked values are detected and stripped so they don't overwrite real credentials in the database
- **No impact on notification sending**: Internal providers read directly from the database, so masking API responses doesn't affect functionality

### Fields masked
| Field | Used by | Example |
|-------|---------|---------|
| `accessToken` | Twilio, Telegram, Pushover, Matrix | `token-abc` → `toke****` |
| `accountSid` | Twilio | `ACxxxxxxxx` → `ACxx****` |

### Endpoints affected
- `POST /notifications` (create response)
- `GET /notifications/team` (list)
- `GET /notifications/:id` (detail)
- `PATCH /notifications/:id` (edit response)

## Test plan
- [ ] Create a Twilio notification, verify response has masked `accountSid` and `accessToken`
- [ ] List notifications, verify sensitive fields are masked
- [ ] Edit a notification without changing credentials — verify existing credentials are preserved
- [ ] Edit a notification with new credentials — verify new values are saved
- [ ] Test notification sending still works (masking only affects API responses, not internal provider calls)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/server/src/controllers/notificationController.ts
+++ b/server/src/controllers/notificationController.ts
@@ -16,23 +16,16 @@ import type { Notification } from "@/types/notification.js";
 
 const SENSITIVE_FIELDS: (keyof Notification)[] = ["accessToken", "accountSid"];
 
-const MASK_SUFFIX = "****";
+const sanitizeNotification = (notification: Notification) => {
+	const sanitized: Record<string, unknown> = { ...notification };
+	const sentinels: Record<string, boolean> = {};
 
-const maskValue = (value: string): string => {
-	if (value.length <= 4) return MASK_SUFFIX;
-	return value.slice(0, 4) + MASK_SUFFIX;
-};
-
-const isMasked = (value: string): boolean => value.endsWith(MASK_SUFFIX);
-
-const maskSensitiveFields = (notification: Notification): Notification => {
-	const masked = { ...notification };
 	for (const field of SENSITIVE_FIELDS) {
-		if (masked[field]) {
-			(masked as Record<string, unknown>)[field] = maskValue(masked[field] as string);
-		}
+		sentinels[`${field}Set`] = typeof sanitized[field] !== "undefined" && sanitized[field] !== null && sanitized[field] !== "";
+		delete sanitized[field];
 	}
-	return masked;
+
+	return { ...sanitized, ...sentinels };
 };
 
 const SERVICE_NAME = "NotificationController";
@@ -84,7 +77,7 @@ class NotificationController implements INotificationController {
 			return res.status(200).json({
 				success: true,
 				msg: "Notification created successfully",
-				data: maskSensitiveFields(notification),
+				data: sanitizeNotification(notification),
 			});
 		} catch (error) {
 			next(error);
@@ -99,7 +92,7 @@ class NotificationController implements INotificationController {
 			return res.status(200).json({
 				success: true,
 				msg: "Notifications fetched successfully",
-				data: notifications.map(maskSensitiveFields),
+				data: notifications.map(sanitizeNotification),
 			});
 		} catch (error) {
 			next(error);
@@ -131,7 +124,7 @@ class NotificationController implements INotificationController {
 			return res.status(200).json({
 				success: true,
 				msg: "Notification fetched successfully",
-				data: maskSensitiveFields(notification),
+				data: sanitizeNotification(notification),
 			});
 		} catch (error) {
 			next(error);
@@ -146,19 +139,11 @@ class NotificationController implements INotificationController {
 			const teamId = requireTeamId(req.user?.teamId);
 			const notificationId = validatedParams.id;
 
-			// Strip masked sensitive fields so they don't overwrite real values
-			const updateData = { ...validatedBody };
-			for (const field of SENSITIVE_FIELDS) {
-				if (updateData[field as keyof typeof updateData] && isMasked(updateData[field as keyof typeof updateData] as string)) {
-					delete updateData[field as keyof typeof updateData];
-				}
-			}
-
-			const editedNotification = await this.notificationsService.updateById(notificationId, teamId, updateData);
+			const editedNotification = await this.notificationsService.updateById(notificationId, teamId, validatedBody);
 			return res.status(200).json({
 				success: true,
 				msg: "Notification updated successfully",
-				data: maskSensitiveFields(editedNotification),
+				data: sanitizeNotification(editedNotification),
 			});
 		} catch (error) {
 			next(error);
--- a/server/src/controllers/notificationController.ts
+++ b/server/src/controllers/notificationController.ts
@@ -12,6 +12,28 @@ import { AppError } from "@/utils/AppError.js";
 import { INotificationsService } from "@/service/index.js";
 import { requireTeamId, requireUserId } from "./controllerUtils.js";
 import { IMonitorsRepository } from "@/repositories/index.js";
+import type { Notification } from "@/types/notification.js";
+
+const SENSITIVE_FIELDS: (keyof Notification)[] = ["accessToken", "accountSid"];
+
+const MASK_SUFFIX = "****";
+
+const maskValue = (value: string): string => {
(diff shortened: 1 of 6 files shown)


server/src/controllers/notificationController.ts
```diff
@@ -12,6 +12,28 @@ import { AppError } from "@/utils/AppError.js";
...
+import type { Notification } from "@/types/notification.js";
+
+const SENSITIVE_FIELDS: (keyof Notification)[] = ["accessToken", "accountSid"];
+
+const MASK_SUFFIX = "****";
+
+const maskValue = (value: string): string => {
+	if (value.length <= 4) return MASK_SUFFIX;
+	return value.slice(0, 4) + MASK_SUFFIX;
+};
```
ajhollid: This works, but it's a pretty fragile solution and an undocumented contract 🤔 

Why don't we return a sentinel value like how setting the Pagespeed API key works?  I think that's a more robust solution.  It requires a bit of fiddling around on the back and front end but makes for better UX

ajhollid (changes requested):
Thanks for the quick turnaround here!  I think there is a more optimal solution we can implement here for better UX, please see how handling of API keys works on the settings page.  I think that's a more robust solution here.  Take a look and see what you think!

egeoztass:
Refactored to use the sentinel value pattern from the settings page:

**Backend:**
- `accessToken` and `accountSid` are deleted from all API responses
- Boolean sentinels `accessTokenSet` and `accountSidSet` are added to indicate if credentials are stored

**Frontend:**
- Edit mode: shows "Token is set" + Reset button when sentinel is `true`
- After clicking Reset (or when creating): shows the input field
- Unchanged sensitive fields are omitted from edit submissions
- Zod schemas updated to make sensitive fields optional for edit mode

Affects all channels using `accessToken`: Twilio, Telegram, Pushover, and Matrix.

995/995 tests passing.

egeoztass:
## Screenshot

Edit mode showing sentinel UI for sensitive fields — Account SID and Auth Token are hidden with a "Reset" option:

image

ajhollid (changes requested):
Saving after setting token and sid doesn't work at the moment.  Those fields should be stripped if they are already set

Image

egeoztass:
Fixed — the backend Zod validation was still requiring `accessToken` and `accountSid` as mandatory fields, so editing without changing them caused a validation error. Made them optional on all affected schemas (Telegram, Pushover, Twilio, Matrix).

995/995 tests passing.

ajhollid (commented):
Image

Errors on creating without required fields 🤔  Can you take a look at the validation schema?

egeoztass:
Fixed — split validation into two schemas:

- **`createNotificationBodyValidation`** — requires `accessToken`/`accountSid` (new notifications must provide credentials)
- **`editNotificationBodyValidation`** — makes them optional (existing credentials preserved when not provided)

The edit endpoint now uses the edit schema. Create and test endpoints still use the strict schema.

995/995 tests passing.

server/src/controllers/notificationController.ts
```diff
@@ -12,6 +13,21 @@ import { AppError } from "@/utils/AppError.js";
...
+	const sanitized: Record<string, unknown> = { ...notification };
+	const sentinels: Record<string, boolean> = {};
+
+	for (const field of SENSITIVE_FIELDS) {
+		sentinels[`${field}Set`] = typeof sanitized[field] !== "undefined" && sanitized[field] !== null && sanitized[field] !== "";
+		delete sanitized[field];
+	}
+
+	return { ...sanitized, ...sentinels };
+};
```
ajhollid, line 30: This is untested as far as I can tell, lets add some tests for this to keep coverage up

server/src/controllers/notificationController.ts
```diff
@@ -114,7 +134,7 @@ class NotificationController implements INotificationController {
 
 	editNotification = async (req: Request, res: Response, next: NextFunction) => {
 		try {
-			const validatedBody = createNotificationBodyValidation.parse(req.body);
+			const validatedBody = editNotificationBodyValidation.parse(req.body);
```
ajhollid, line 137: Stripping the values out of the edit notiifcation seem to rely entirely on the client formatting the request correctly.

We have no idea what kind of client the user is going ot be interacting with, so we should not rely on that.  We should also strip out values server side.  Example, what happens if I craft a `cURL` request with a `null` for `accessToken`?  I'm not actually sure, but we should definitely handle this server side and not depend on the client

ajhollid (changes requested):
This is a decent approach, but we should derive the edit notification validation from the create notification validation so we don't duplicate the schema.

Why don't we declare the schemas explictly: 

```
  const pushoverSchema = z.object({                                                                                                                                                                                          
        notificationName: z.string().min(1, "Notification name is required"),
        type: z.literal("pushover"),
        address: z.string().min(1, "User key is required"),
        accessToken: z.string().min(1, "App token is required"),                                                                                                                                                             
  });
                                                                                                                                                                                                                             
  const twilioSchema = z.object({
        notificationName: z.string().min(1, "Notification name is required"),
        type: z.literal("twilio"),                                                                                                                                                                                           
        accountSid: z.string().min(1, "Account SID is required"),
        accessToken: z.string().min(1, "Auth token is required"),                                                                                                                                                            
        phone: z.string().min(1, "Recipient phone number is required"),                                                                                                                                                      
        twilioPhoneNumber: z.string().min(1, "Twilio phone number is required"),
  }); 
```

And then use unions to create the validations:

```
  export const createNotificationBodyValidation = z.discriminatedUnion("type", [
        ...
        pushoverSchema,
        twilioSchema,
  ]);

  export const editNotificationBodyValidation = z.discriminatedUnion("type", [
        ...                                                                                                                                                              
        pushoverSchema.partial({ accessToken: true }),
        twilioSchema.partial({ accessToken: true, accountSid: true }),                                                                                                                                                       
  ]);  
```

This saves us a lot of duplication.

Other than that, we should do something server side about stripping values when editing notifications, we don't want to rely on the client to submit a well formatted request.

Thanks again for changes!

egeoztass:
Both points addressed:

1. **DRY schemas** — Each notification schema is declared once. Edit validation derives from create using `.partial()`:
   ```typescript
   matrixSchema.partial({ accessToken: true }),
   telegramSchema.partial({ accessToken: true }),
   pushoverSchema.partial({ accessToken: true }),
   twilioSchema.partial({ accessToken: true, accountSid: true }),
   ```
   This cut ~80 lines of duplication.

2. **Server-side stripping** — The edit controller now strips undefined sensitive fields before passing to `updateById`, so we don't rely on the client to format the request correctly.

995/995 tests passing.

edobusy:
Hi @egeoztass! Are you still working on this? If you've moved on, I'd be happy to finish what you have started :)

For @ajhollid, I opened #3771 before I spotted this PR (should have searched harder, sorry). It's the same fix, plus one case this PR doesn't cover: the Test button sends the form's current values, so once the token is hidden, testing an existing channel goes out without it. Happy to work on this however you'd prefer, either pick up this branch, or open a fresh PR carrying over your review feedback here.

egeoztass:
Superseded by #3803, which covers this plus the Test-button case I missed — the form sends its current values, so testing a saved channel after the credential is hidden goes out without it. @edobusy's version has manual verification on the thread too.

Closing this in favour of that one. @edobusy thanks for picking it up, and sorry for the stall on my end. The two merged channel PRs (#3530, #3534) are unaffected.

PR #3584, opened 2026-04-27
fix: allow mono and arm containers to run as non-root user (UID 1000)

### LINKED ISSUE #3577: Feature: Allow docker image to run as non-root user

Certain environments don't allow containers to run as root. When attempting to run as another UID this error occurs and the container crashes:

```
sed: couldn't open temporary file ./public/assets/sednxgh3x: Permission denied
```

chown the files  when building the docker image to the `node` user, UID 1000. This should allow users to run the app with `--user 1000`, without impacting existing deployments.

**Additional context**

 The mongo container can be ran as UID 999 - without issues.

Iriome-Santana:
## Problem

When running the container with `--user 1000` in environments that
don't allow containers to run as root (e.g. Kubernetes with
`runAsNonRoot: true`, OpenShift, or any hardened deployment), the
container crashes immediately:
sed: couldn't open temporary file ./public/assets/sednxgh3x: Permission denied

This happens because `inject-vars.sh` uses `sed -i`, which creates a
temporary file in the same directory as the file being edited.
The `./public/assets/` directory is owned by root after the COPY
instruction, so a non-root process can't write there.

## Solution

In both `dist-mono/server.Dockerfile` and `dist-arm/server.Dockerfile`:

1. `chown -R node:node` on `public/`, `dist/`, and `scripts/` after
   the frontend build is copied
2. `USER node` before `EXPOSE` so the container runs as UID 1000 by default

The `node` user (UID 1000) is already present in the `node:20-slim`
base image — no new users are created.

## Impact

- Existing deployments running as root: **no change**
- Deployments with `--user 1000` or `runAsNonRoot: true`: **now work**

Closes #3577

### DIFF
--- a/docker/dist-arm/server.Dockerfile
+++ b/docker/dist-arm/server.Dockerfile
@@ -44,7 +44,6 @@ RUN chown -R node:node ./public \
     && chown -R node:node ./dist \
     && chown -R node:node ./scripts
 
-USER node
 
 EXPOSE 52345
 
--- a/docker/dist-arm/server.Dockerfile
+++ b/docker/dist-arm/server.Dockerfile
@@ -40,6 +40,12 @@ RUN npm run build
 
 COPY --from=frontend-build /app/client/dist ./public
 
+RUN chown -R node:node ./public \
+    && chown -R node:node ./dist \
+    && chown -R node:node ./scripts
+
+USER node
+
 EXPOSE 52345
 
 CMD ["sh", "-c", "./scripts/inject-vars.sh && node ./dist/index.js"]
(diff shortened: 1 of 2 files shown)


Iriome-Santana:
I can also switch this to an opt-in approach if changing the default user is a concern

ajhollid:
> I can also switch this to an opt-in approach if changing the default user is a concern

Hey @Iriome-Santana ,

As this behaviour has been the standard thus far and users expect this to run as root, let's make this opt-in to make the disruption as minimal as possible.  Thanks!

Iriome-Santana:
Done! Removed the USER node directive — the container still runs as root by default, but --user 1000 now works correctly since the files are owned by node. Let me know if you'd like any other changes

ajhollid (approved):
Thanks for making changes!

PR #3590, opened 2026-04-28
feat(create-monitor): explain server-side query setup for game monitors

gorkem-bwl:
## Summary

The game monitor type relies on each game's native query protocol — many require a config file change on the game server (e.g. Minecraft Java needs `enable-query=true` in `server.properties`). Without that, the monitor silently fails to reach the server, and there was no in-app hint pointing users at the cause.

This appends one sentence to the game-type general settings description that tells the user some games need server-side queries enabled, and links to the GameDig supported-games list for protocol details.

## Changes

- `CreateMonitor/index.tsx` — swap the `t(...)` call for the general-settings subtitle to a `<Trans>` so the link can render inline. Defines a `gamedigLink` slot pointing to the GameDig GAMES_LIST.md
- `SplitBox.tsx` — loosen `ConfigBox.subtitle` from `string` to `React.ReactNode` so it can host the `<Trans>` output
- `locales/*.json` — extend the `pages.createMonitor.form.general.description.game` string in all 16 locales:
  - **Translated** (existing description present): de, es, ru, th, zh-CN, zh-TW
  - **English fallback** (description was empty): ar, cs, fi, fr, ja, pt-BR, tr, uk, vi

The link itself is the same in every locale, only the surrounding sentence is translated.

## Test plan

- [ ] On the create-monitor page, pick **Game** as the type. The general-settings subtitle now ends with a sentence about enabling server queries plus a clickable "supported games list" link
- [ ] Click the link — opens GameDig's GAMES_LIST.md in a new tab
- [ ] Switch language to a translated locale (de / es / ru / zh-CN / zh-TW / th) and confirm the new sentence renders in that language
- [ ] Switch to an untranslated locale (e.g. fr) and confirm the English text shows
- [ ] Pick a non-game monitor type — the subtitle is unchanged, no link appears


gorkem-bwl:
UI change, tested, localized.

PR #3448, opened 2026-03-27
feat(ui): add bulk edit notifications modal and floating action bar to uptime monitors

ashu130698:
## Changes

This PR implements the frontend user interface for bulk-editing monitor notifications, building seamlessly on the backend work from my previous PR (#3376).

As requested by @gorkem-bwl in the issue thread, this introduces a clean, non-intrusive **Floating Action Bar** pattern that appears only when multiple monitors are selected in the table, rather than a bulky static "bulk edit" button.

**Core components introduced:**
- **[FloatingActionBar.tsx](cci:7://file:///c:/Projects/Checkmate/client/src/Components/common/FloatingActionBar.tsx:0:0-0:0)**: A reusable slide-up bar component.
- **[BulkEditNotificationsModal.tsx](cci:7://file:///c:/Projects/Checkmate/client/src/Components/monitors/BulkEditNotificationsModal.tsx:0:0-0:0)**: A dialog to select the action (`add`, `remove`, `set`) and query your available notification channels via [useGet](cci:1://file:///c:/Projects/Checkmate/client/src/Hooks/UseApi.ts:19:0-37:2).
- **[UptimeMonitorsTable.tsx](cci:7://file:///c:/Projects/Checkmate/client/src/Pages/Uptime/Monitors/Components/UptimeMonitorsTable.tsx:0:0-0:0) Integration**: Safely injected a checkbox column to the generic table headers to handle selection states.
- Lifted the `selectedMonitors` array into the [Uptime/Monitors/index.tsx](cci:7://file:///c:/Projects/Checkmate/client/src/Pages/Uptime/Monitors/index.tsx:0:0-0:0) page to sync the bar and table.

*(Note: Currently applied to the Uptime Monitors page as an MVP to gather feedback on the design. If approved, I will follow up and apply this exact same logic to the Infrastructure and PageSpeed tables!)*

Fixes #2320

- [x] I deployed the application locally.
- [x]  I have performed a self-review and testing of my code.
- [x] I have included the issue #2320 in the PR.
- [x] I have added i18n support to visible strings.
- [x] I have **not** included any files that are not related to my pull request.
- [x] I didn't use any hardcoded values.
- [x] I made sure font sizes, colour choices, etc., are all referenced from the theme. I don't have any hardcoded dimensions.
- [x] My PR is granular and targeted to one specific feature.
- [x] I ran `npm run format` in server and client directories, which automatically formats your code.
- [x] I took a screenshot and attached it to this PR if there is a UI change.

[...]

PR #3626, opened 2026-05-01
feat/bulk pause resume

### LINKED ISSUE #3602: Feature Request: Bulk Pause and Resume actions for Monitors
There is currently no way to pause or resume multiple monitors at once. While individual toggles exist on each alert/monitor, managing maintenance windows or large-scale updates is difficult because users must click "Pause" or "Resume" one by one.

Proposed Solution:

Add a bulk action menu to the main dashboard or monitor list.

* Checkbox selection: Allow users to select multiple monitors via checkboxes.
* Action Bar: When items are selected, show a "Bulk Actions" bar with "Pause Selected" and "Resume Selected" buttons.

Use Case:

When performing server maintenance that affects multiple services, I need to quickly pause all related monitors to prevent a "notification storm." Once maintenance is over, I need to resume them all just as quickly to ensure coverage.

harsh-aghara:
## Describe your changes

Implements bulk pause/resume functionality for monitors, allowing users to select multiple monitors via checkboxes and pause or resume them in a single action.

**Backend:**
- Added `POST /api/v1/monitors/bulk/pause` endpoint with Zod validation (`pause: boolean` flag controls pause/resume)
- Optimized `bulkTogglePause` repository method using `$in`/`$ne` queries to only update monitors that actually need a state change
- Orchestrates BullMQ job pausing/resuming for each affected monitor
- Added unit tests for the new service method

**Frontend:**
- Created reusable `useTableSelection` hook for generic table row selection with select-all/indeterminate support
- Created `useBulkMonitorActions` hook to handle bulk API calls and toast notifications
- Created `BulkActionsBar` component with smooth Collapse animation
- Integrated selection checkboxes and bulk actions bar into both Uptime and Infrastructure monitor tables
- Enhanced `Table.tsx` to support `mobileLabel` prop for clean mobile card rendering of non-standard columns (e.g., checkboxes)
- Added `stopPropagation` on checkbox clicks to prevent row navigation
- Added i18n keys across all 16 locales

Fixes #3602

```Javascript
<div>{t('add')}</div>
```

https://github.com/user-attachments/assets/8be8661d-f2fc-45b0-9f48-4834484138f1


ajhollid (commented):
Hey @harsh-aghara ,

Thanks for submitting this PR.

I think visually it looks great, however there is going to be a collision with [this PR](https://github.com/bluewave-labs/Checkmate/pull/3448).

I suggest you collaborate with the author @ashu130698 and @gorkem-bwl to see how to resolve this with regard to both UI/UX and code.

gorkem-bwl:
> Hey @harsh-aghara ,
> 
> Thanks for submitting this PR.
> 
> I think visually it looks great, however there is going to be a collision with [this PR](https://github.com/bluewave-labs/Checkmate/pull/3448).
> 
> I suggest you collaborate with the author @ashu130698 and @gorkem-bwl to see how to resolve this with regard to both UI/UX and code.
> 
> Thanks for working on the project!

We need to convert this to a pill at the bottom of the page. Also can you rework on the strings? We need to be picky about the message.

`0 monitor(s) resumed successfully`

This shouldn't read like this. If there are 0 monitors to be resumed, this line should read better. Also check singularity/plurality and refrain from using (s). 

@ajhollid when those fixes are there I believe this should go first as it's a bit more cleaner than the other PR, and 3448 should be rebased/worked on after letting this one ine.

harsh-aghara:
Thanks @gorkem-bwl 
I will convert the bar to a bottom pill and fix the toast strings. Will push and ping you soon.

harsh-aghara:
Hey @ajhollid @gorkem-bwl, I just pushed an update:

- Moved the actions to a floating bottom pill ( I didn't make it fully round because it didn't fit the UI well, but I centered it dynamically so it smoothly adjusts when the sidebar opens and closes ). I also made it modular so it won't collide with PR #3448.
- Fixed the table alignments and swapped in our custom Checkbox components.
- Cleaned up the translations to use proper pluralization across all locales instead of the generic (s).

Should be good to go now, let me know what you think!
image
image
image
image
image

ashu130698:
Hey @harsh-aghara, the new modular BulkActionsBar and generic selection hooks look awesome! Really clean implementation.

I'm tracking this PR and will rebase my bulk notifications UI (#3448) to hook directly into your new components as soon as this gets merged. Great work!

gorkem-bwl:
@ajhollid we should be good to merge this hopefully and move on to the other PR 👍

ajhollid:
This PR is very large, 1k line PRs are generally not accepted.

I'll take a look tomorrow, but this should be properly rescoped into frontend and backend functionality PRs as per project standard

I see some of the diff is due to translations, I'll take that into account during final review, but in general PRs should be much smaller and focused

harsh-aghara:
@ajhollid I understand this pr is too long. Should I split it into frontend and backend prs before your review, or would you prefer to review as it is?

ajhollid:
Let me take a look at it tomorrow, if I can reasonably review it I will 👌

In future though yes, backend functionality should be implemented first with tests. 

Then once functionality is confirmed a UI PR can be made and evaluated on its own merit.

Will check in with you tomorrow!

client/src/Pages/Infrastructure/Monitors/Components/MonitorsTable.tsx
```diff
@@ -101,21 +114,14 @@ export const InfraMonitorsTable = ({
...
-			// },
 			{
 				id: 6,
 				label:
 					monitor.isActive === false
 						? t("common.buttons.resume")
 						: t("common.buttons.pause"),
 				action: async () => {
-					await post(`/monitors/pause/${monitor.id}`, {});
+					await post(`/monitors/pause/${monitor.id}`, {} as Record<string, never>);
```
ajhollid: No casting please, lets use correct types.  Casting removes type safety and defeats the purpose of TypeScript really

client/src/Components/design-elements/Table.tsx
```diff
@@ -123,7 +143,9 @@ export function DataTable<
 												component="div"
 												color={theme.palette.text.primary}
 											>
-												{header.content}
+												{header.mobileLabel !== undefined
+													? header.mobileLabel
+													: header.content}
```
ajhollid: This is a triple condition that is _very_ easy to miss if you aren't very careful.  It is also not easy to grep if we are looking for this behaviour later.

This is equivalent to

```
if(null) {do A}
else if (!undefined) {do B}
else {do C}
```

An explicit `boolean` flag here makes more sense; we should be precise in our code rather than depending on side effects of something being defined or undefined.

client/src/Hooks/useTableSelection.ts
```diff
@@ -0,0 +1,74 @@
...
+
+interface UseTableSelectionReturn {
+	isAllSelected: boolean;
+	isSomeSelected: boolean;
+	handleSelectAll: (checked: boolean) => void;
+	handleSelectRow: (itemId: string, checked: boolean) => void;
+	isRowSelected: (itemId: string) => boolean;
+}
+
+export const useTableSelection = (
```
ajhollid: I don't think a hook is necessary for this functionality; the abstraction obfuscates more than it simplifies.  It moves state away from the page where it is used for no real gain in my opinion.

All we're doing with this hook really is moving state out of the page it is used just to pipe it back in again.

client/src/Hooks/useBulkMonitorActions.ts
```diff
@@ -0,0 +1,84 @@
...
+			} else {
+				const key = pause
+					? "pages.common.monitors.bulkPause.paused"
+					: "pages.common.monitors.bulkPause.resumed";
+				toastSuccess(t(key, { count: affectedCount }));
+			}
+
+			setSelectedRows([]);
+			refetch();
+		} catch (err: any) {
```
ajhollid: Errors should be typed as `unknown` and their type discriminated in the catch block.  The `any` type is disallowed in this project.

server/src/repositories/monitors/MongoMonitorsRepository.ts
```diff
@@ -234,6 +234,31 @@ class MongoMonitorsRepository implements IMonitorsRepository {
...
+	bulkTogglePause = async (monitorIds: string[], teamId: string, pause: boolean): Promise<Monitor[]> => {
+		const objectIds = monitorIds.map((id) => new mongoose.Types.ObjectId(id));
+		const filter = {
+			_id: { $in: objectIds },
+			teamId: new mongoose.Types.ObjectId(teamId),
+			isActive: pause, // Only pause if active (true), only resume if inactive (false)
+		};
+
+		const eligibleIds = (await MonitorModel.find(filter).select("_id").lean()).map((doc) => doc._id);
+		if (eligibleIds.length === 0) return [];
```
ajhollid: This DB op is not necessary; the eligibility can be integrated into the `updateMany` call.

You should be able to just call something like

```
  const filter = { _id: { $in: objectIds }, teamId, isActive: pause };                                                                                                                                                       
                                                                                                                                                                                                                             
  const { modifiedCount } = await MonitorModel.updateMany(filter, {
      $set: { isActive: !pause, status: newStatus },                                                                                                                                                                         
  });                                               
  if (modifiedCount === 0) return [];  
```

server/src/service/business/monitorService.ts
```diff
@@ -462,6 +463,22 @@ export class MonitorService implements IMonitorService {
...
+				if (monitor.isActive) {
+					await this.jobQueue.resumeJob(monitor);
+				} else {
+					await this.jobQueue.pauseJob(monitor);
+				}
+			})
+		);
+
+		return monitors;
+	};
```
ajhollid: This is somewhat problematic. What happens if one op rejects here? All the other ops keep running, some may succeed, some may fail, we won't know, their success/rejection is unhandled.

The DB and Queue have also now drifted out of sync, the DB shows they are all paused/resumed, but the queue is possibly not correct.

This is not a trivial problem to solve so I won't ask you to look at this right now, but at the very least we should use [promise.allSettled()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/allSettled) here to log individual failures and report that to the user.

ajhollid (changes requested):
It's a decent start, but there are some issues that need to be addressed before this is ready to merge.

Please have a look at my comments in the code review and revise as needed.
Thank you!

client/src/Pages/Infrastructure/Monitors/Components/MonitorsTable.tsx
harsh-aghara: Removed the type casting..

client/src/Components/design-elements/Table.tsx
harsh-aghara: Replaced the null and undefined checks in with a hideMobileLabel bool flag.

server/src/repositories/monitors/MongoMonitorsRepository.ts
harsh-aghara: I initially tried doing exactly this to save the DB round trips! The issue I ran into is that updateMany only returns the modifiedCount, not the actual updated data.

Because the service layer immediately needs the updated monitor objects to pass into the BullMQ job queue (and the frontend uses the returned array to count how many were actually changed) , I had to stick with the 3-query approach (find eligible -> update -> fetch updated).
I also tried a 2-query approach where I modified the old documents in-memory before returning them, but that caused stale updatedAt timestamps.

 Let me know if there's a cleaner way around that!

server/src/service/business/monitorService.ts
harsh-aghara: I've updated the backend to use Promise.allSettled(). It now processes the entire batch safely, catching and logging individual queue failures to the console without crashing.

I deliberately left two things out of this PR to keep the scope manageable:

- Connecting the frontend partial-success warning toasts.
- The actual DB/Queue rollback logic when a failure occurs.

I think we should merge this allSettled() backend safety net for now, and I can open two follow-up issues to tackle the frontend warnings and state drift properly later.
Does that work?

server/src/repositories/monitors/MongoMonitorsRepository.ts
ajhollid: > I initially tried doing exactly this to save the DB round trips! The issue I ran into is that updateMany only returns the modifiedCount, not the actual updated data.
> 
> Because the service layer immediately needs the updated monitor objects to pass into the BullMQ job queue (and the frontend uses the returned array to count how many were actually changed) , I had to stick with the 3-query approach (find eligible -> update -> fetch updated). I also tried a 2-query approach where I modified the old documents in-memory before returning them, but that caused stale updatedAt timestamps.
> 
> Let me know if there's a cleaner way around that!

Why not set the timestamps manually?

```
const now = new Date()
await MonitorModel.updateMany(
	{ _id: { $in: eligible.map((doc) => doc._id) } },
	{ $set: { isActive: !pause, status: nextStatus, updatedAt: now } },
	{ timestamps: false }
);
```

```
eligible.forEach((doc) => {
	doc.isActive = !pause;
	doc.status = nextStatus;
	doc.updatedAt = now;
});
return this.mapDocuments(eligible);
```

That should resolve the problem 🤔

ajhollid (changes requested):
Thanks for making those changes.  I believe the multiple DB calls can be handled without too much headache, please see my example in this latest review.

As for the DB rollback issue, don't worry about that, that's not unique to or caused by your PR, so we can resolve that at a later date.

Handling the messaging can also be done in a separate PR, you can tackle that one if you'd like!

Thanks for working on it 🚀

server/src/service/business/monitorService.ts
harsh-aghara: `{ timestamps: false }` was exactly what I needed! I didn't realize we could just bypass Mongoose auto-timestamp like that. 
That perfectly solves the stale timestamp issue I was hitting. 
Just pushed the update, thanks for the suggestion! 🙌

ajhollid (approved):
Thanks for making the changes!

harsh-aghara:
Oops, server build fails let me fix it quickly

harsh-aghara:
@ajhollid Build issue is fixed.
Thanks for the approval!

PR #3449, opened 2026-03-29
Dns addition in monitor

Owaiseimdad:
## Describe your changes

Implemented a new DNS Monitor type to allow users to track domain resolution success and query latency. Key changes include:
Core Logic: Created DnsProvider.ts using Node's dns/promises API. It supports configurable DNS servers and multiple record types (A, AAAA, CNAME, MX, TXT, NS).

Backend Persistence:
Updated the Mongoose Monitor schema to include dnsServer and dnsRecordType.
Updated both MongoMonitorsRepository and TimescaleMonitorsRepository to ensure DNS configuration persists across both database types. Updated MonitorService to support fetching uptime details for the new DNS type.

Validation: Refactored the backend validation to use the centralized MonitorTypes constant in monitorValidation.ts and checkValidation.ts, resolving Zod validation errors for the new type.

Frontend UI:
Added DNS-specific fields (DNS Server, Record Type) to the CreateMonitorPage.
Updated the monitor list and ControlsFilter to include DNS monitors by default.
Improved UI consistency by rounding both Average and Maximum latency values to the nearest millisecond in 
HistogramResponseTime.tsx.

i18n: Added comprehensive English translations for all new DNS-related labels and descriptions in en.json
.

## Write your issue number after "Fixes "
#3388 

## Please ensure all items are checked off before requesting a review. "Checked off" means you need to add an "x" character between brackets so they turn into checkmarks.

- [x] (Do not skip this or your PR will be closed) I deployed the application locally.
- [x] (Do not skip this or your PR will be closed) I have performed a self-review and testing of my code.
- [x] I have included the issue # in the PR.
- [x] I have added i18n support to visible strings (instead of `<div>Add</div>`, use): 
```Javascript
const { t } = useTranslation();
<div>{t('add')}</div>
```
- [x] I have **not** included any files that are not related to my pull request, including package-lock and package-json if dependencies have not changed
- [x] I didn't use any hardcoded values (otherwise it will not scale, and will make it difficult to maintain consistency across the application).
- [x] I made sure font sizes, color choices etc are all referenced from the theme. I don't have any hardcoded dimensions.
- [x] My PR is granular and targeted to one specific feature.
- [x] I ran `npm run format` in server and client directories, which automatically formats your code.
- [x] I took a screenshot or a video and attached to this PR if there is a UI change.
Screenshot 2026-03-29 at 10 56 05 PM
Screenshot 2026-03-29 at 10 56 19 PM

PR #3635, opened 2026-05-06
feat(monitor): add DNS monitor type (backend)

### LINKED ISSUE #3388: Add a new DNS monitor type that periodically queries DNS records and tracks two key metrics
Users want to monitor whether their domains are resolving correctly and how fast DNS responses are. This complements existing HTTP and ping monitors by catching DNS-level issues that may not surface through higher-level checks.

Add a new DNS monitor type that periodically queries DNS records and tracks two key metrics:

1. Query time: The time it takes for a DNS query to resolve (similar to ping/HTTP response time). This can be visualized as a graph on the monitor details page, consistent with how we display response times for other monitor types.
2. Query status: Whether the DNS record is resolving successfully or not. This enables uptime/downtime tracking for DNS  resolution, triggering incidents when a domain stops resolving.

- Use a DNS library (e.g., Node.js dns.resolve) to perform lookups
- Support configurable DNS server (e.g., 8.8.8.8, 1.1.1.1, or custom)
- Support common record types (A, AAAA, CNAME, MX, TXT, NS)
- Track query time in milliseconds and resolution success/failure
- Integrate with existing incident and notification systems

gorkem-bwl:
## Summary

Adds a new DNS monitor type that resolves a hostname against an optional custom DNS server and records query latency. Supports `A`, `AAAA`, `CNAME`, `MX`, `TXT`, `NS`. Pairs with the frontend in **#3636** — recommend merging this PR first.

This is a takeover of @Owaiseimdad's work in #3449 (closed). The original PR's review feedback from @Br0wnHammer is fully addressed; original commits are credited via `Co-authored-by`.

### What it does

- **DnsProvider** (new): uses Node's `dns/promises` `Resolver`, injected via the constructor. Each query is wrapped in a 10s timeout and timed via the shared `timeRequest()` helper. Unknown record types and missing hostnames raise `AppError` with `service`/`method`/`details` (matches `WebSocketProvider` / `GrpcProvider`).
- **Expected-value matching**: when the operator sets `expectedValue` on a DNS monitor, the provider compares the resolved records and marks the monitor down if none match. Comparison rules per record type: A/AAAA/CNAME/NS = string equality; MX = match against `exchange`; TXT = joined chunks. Case-insensitive, trims whitespace.
- **Last resolution snapshot**: a `lastDnsResolution` field on the `Monitor` document captures `hostname`, `dnsServer`, `recordType`, `results`, `matched`, and `resolvedAt` after each successful check. The frontend renders this on the monitor detail page (in the FE PR). Failed resolutions don't overwrite the previous snapshot.
- **Schema**: `Monitor` model gains optional `dnsServer`, `dnsRecordType` (enum over `DnsRecordTypes`), and `lastDnsResolution`. `DnsStatusPayload` added to `MonitorPayloadMap`. Validation routed through the centralized `MonitorTypes` const.
- **OpenAPI**: `monitorResponseSchema` exposes the new fields so SDK consumers and the generated `openapi.json` reflect them.
- **Tests**: 18 unit tests (`server/test/unit/providers/network/dnsProvider.test.ts`) — contract test, all 6 record types, custom resolver, failure path, AppError on missing hostname, AppError on unknown record type, timeout, expected-value matching (6 cases). Matches peer-provider conventions (inline mock + provider per test).

### Test plan

- [ ] `cd server && npm test` — all 49 suites / 1050 tests pass
- [ ] `cd server && npm run build` succeeds (uses `tsconfig.build.json`)
- [ ] `cd server && npm run format-check` clean
- [ ] Create a DNS monitor against `example.com` with no resolver, A record. Verify it goes up.
- [ ] Create a DNS monitor against `example.com` with `expectedValue` set to a value not in the record set. Verify it goes down with "Resolved records do not match expected value …" message.
- [ ] Create a DNS monitor with `dnsServer: 8.8.8.8`. Verify it queries Google DNS.
- [ ] After a successful check, GET `/monitors/{id}` and confirm `lastDnsResolution` is populated with the correct shape.

### Closes / supersedes

- Closes #3388 (original DNS monitor request).
- Closes #3449 (the original PR, which the author was unable to continue).


ajhollid:
Already implemented

PR #3641, opened 2026-05-08
Fix affected monitors not updating when editing maintenance windows

### LINKED ISSUE #3551: [Bug] Changing monitors when editing a maintenance window has no effect
**Describe the bug**

When editing a maintenance window, changes to **name, repeat, start date, …** are saved correctly. However, adding or removing monitors has no effect; the monitor list is not updated after clicking **Save**.

**To Reproduce**
1. Go to **Maintenance Windows**
2. Click **Configure** on any maintenance window
3. Add or remove a monitor from the monitors list
4. Click **Save**
5. Re-open the same window via **Configure**
6. The monitor list is unchanged (change didn't persist)

**Expected behavior**
Adding or removing monitors from a maintenance window should be saved when clicking **Save**, consistent with how other fields are already updated correctly.

**Screenshots**

 - Browser [e.g. Chrome, Safari] Chrome
 - Version [e.g. 22] 147.0.7727.57 (Official Build) (arm64) 

**Additional context**
Unlike some fields that are stored directly on the maintenance window document and updated correctly via PATCH, monitors are stored as separate DB records (one record per monitor).
The edit flow sends `monitors: string[]` in the PATCH body, but the server only updates one document without creating or deleting sibling records to reflect the change.

Autumn1101:
## Describe your changes

Fix maintenance window edits so changing the affected monitor is persisted.

- Map the edit form's `monitors` payload to the stored `monitorId` field.
- Allow `monitorId` to be updated on maintenance windows.
- Validate edited monitor selections and verify team ownership before updating.
- Add service and validation coverage for affected monitor edits.

Fixes #3551

- [x] I have added i18n support to visible strings.
- [ ] I ran `npm run format` in server and client directories, which automatically formats your code.

### DIFF
--- a/client/src/Hooks/useMaintenanceWindowForm.ts
+++ b/client/src/Hooks/useMaintenanceWindowForm.ts
@@ -32,7 +32,7 @@ export const useMaintenanceWindowForm = ({
 				startTime: startDate.format("HH:mm"),
 				duration: data.duration ?? 0,
 				durationUnit: data.durationUnit ?? "minutes",
-				monitors: [data.monitorId],
+				monitors: data.monitors ?? [data.monitorId],
 			};
 		} else {
 			const now = dayjs();
--- a/client/src/Types/MaintenanceWindow.ts
+++ b/client/src/Types/MaintenanceWindow.ts
@@ -2,7 +2,9 @@ export type DurationUnit = "seconds" | "minutes" | "hours" | "days";
 
 export interface MaintenanceWindow {
 	id: string;
+	groupId?: string;
 	monitorId: string;
+	monitors?: string[];
 	teamId: string;
 	active: boolean;
 	name: string;
--- a/server/src/db/models/MaintenanceWindow.ts
+++ b/server/src/db/models/MaintenanceWindow.ts
@@ -1,7 +1,11 @@
 import { Schema, model, type Types } from "mongoose";
 import { DurationUnits, type MaintenanceWindow } from "@/types/maintenanceWindow.js";
 
-type MaintenanceWindowDocumentBase = Omit<MaintenanceWindow, "id" | "monitorId" | "teamId" | "start" | "end" | "createdAt" | "updatedAt"> & {
+type MaintenanceWindowDocumentBase = Omit<
+	MaintenanceWindow,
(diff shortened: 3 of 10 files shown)


Autumn1101:
"Hi @Br0wnHammer, @ajhollid, @karenvicent, sorry to bother you. I've verified the fix locally, but it seems the CI workflows need a maintainer's approval to run. Could you please help trigger them? Thanks!"

Br0wnHammer:
Hi @Autumn1101, you've got a failing check. Can you please look into it?

Br0wnHammer (changes requested):
Thanks for the work here, but I don't think this fully resolves #3551. Because each maintenance window is stored as one document per monitor. Now that you are mapping monitors with monitor ID, it only handles swaping the monitor. When I am trying to add a new monitor in the edit form, it is being silently dropped, and removing one isn't handled at all, so the bug in the issue still reproduces. I'd suggest either narrowing the scope or addressing the data model. Moreover, creating a single maintenance window with two-three monitors produces 2-3 separate identically named rows in the list.

Autumn1101:
Thanks for the detailed feedback. I pushed an update that handles the per-monitor maintenance-window rows as a group.Multi-monitor maintenance windows now share a groupId and are returned as one grouped item in the list. Editing that grouped item now keeps the underlying rows in sync when monitors are added, removed, or retained. Field-only edits such as pause/resume are applied across the related group as well.
I also kept a fallback for existing maintenance-window rows that do not have a groupId yet, so older data can still be handled consistently.

ajhollid:
I think perhaps it would be better to reconsider how maintenance windows work entirely, the current way they function doesn't seem to align with how folks want them to work.

Rather than try to shoehorn functionality in, let's take some time to think about how the maintenance windows currently work vs how people would like them to work.

Once we have a solid understanding, then we can develop a plan, then we can write code.

Autumn1101:
That makes sense, thanks for the clarification. I understand this may need a broader rethink before more code changes are made.
I’ll pause further changes on this PR for now. Thanks again for taking the time to review and explain the direction.

ajhollid:
Hey @Autumn1101 ,

I appreciate your work on this, in the interest of expediency I've made the necessary architectural changes to allow for proper groupings of monitors within a single maintenance window, so I believe we can close this one down.

Thank you for working on the project, we'd love to have your contributions in other areas!

PR #3649, opened 2026-05-12
feat: maintenance window groups

ajhollid:
This PR adds grouping features to maintenance windows.

The current behaviour when creating a maintenance window is that one window is created per monitor.  This PR changes this behaviour and instead creates one maintenance window with multiple monitors.


Br0wnHammer (approved):
LGTM!

PR #3673, opened 2026-05-25
Feat: Uptime Progress Indicator

Br0wnHammer:
## Describe your changes

Splits the overwhelming single-page "Create Uptime Monitor" form into a focused 3-step wizard (Basics → Monitoring → Advanced) with a slim segmented progress bar in Checkmate's green palette, validating each step before advancing. Scoped to uptime create only — edit/configure, pagespeed, and hardware flows keep the existing flat form unchanged.

```Javascript
<div>{t('add')}</div>
```

Screenshot 2026-05-25 at 23 36 22
Screenshot 2026-05-25 at 23 36 51
Screenshot 2026-05-25 at 23 40 35

### DIFF
--- a/client/src/Components/design-elements/StepProgress.tsx
+++ b/client/src/Components/design-elements/StepProgress.tsx
@@ -14,29 +14,30 @@ export const StepProgress = ({ steps, current }: StepProgressProps) => {
 	const activeBg = `linear-gradient(135deg, ${theme.palette.primary.light} 0%, ${theme.palette.primary.main} 100%)`;
 	const completedBg = `linear-gradient(135deg, ${alpha(theme.palette.primary.light, 0.5)} 0%, ${alpha(theme.palette.primary.main, 0.5)} 100%)`;
 
+	const getSegmentBackground = (index: number) => {
+		if (index < current) return completedBg;
+		if (index === current) return activeBg;
+		return theme.palette.divider;
+	};
+
 	return (
 		<Stack
 			direction="row"
 			gap={theme.spacing(LAYOUT.XS)}
 			width="100%"
 		>
-			{Array.from({ length: steps }).map((_, index) => {
-				const background =
-					index < current
-						? completedBg
-						: index === current
-							? activeBg
-							: theme.palette.divider;
-				return (
-					<Box
-						key={index}
-						flex={1}
-						height={theme.spacing(LAYOUT.XXS)}
-						borderRadius={theme.shape.borderRadius}
-						sx={{ background, transition: "background 0.2s ease" }}
-					/>
-				);
-			})}
+			{Array.from({ length: steps }).map((_, index) => (
+				<Box
+					key={index}
+					flex={1}
+					height={theme.spacing(LAYOUT.XXS)}
+					borderRadius={theme.shape.borderRadius}
+					sx={{
+						background: getSegmentBackground(index),
+						transition: "background 0.2s ease",
+					}}
+				/>
+			))}
 		</Stack>
 	);
 };
--- a/client/src/Components/design-elements/StepProgress.tsx
+++ b/client/src/Components/design-elements/StepProgress.tsx
@@ -1,7 +1,6 @@
 import Stack from "@mui/material/Stack";
 import Box from "@mui/material/Box";
-import { useTheme } from "@mui/material/styles";
-import { alpha } from "@mui/material/styles";
+import { useTheme, alpha } from "@mui/material/styles";
 import { LAYOUT } from "@/Utils/Theme/constants";
 
 interface StepProgressProps {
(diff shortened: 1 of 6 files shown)


client/src/Components/design-elements/StepProgress.tsx
```diff
@@ -0,0 +1,42 @@
...
+			gap={theme.spacing(LAYOUT.XS)}
+			width="100%"
+		>
+			{Array.from({ length: steps }).map((_, index) => {
+				const background =
+					index < current
+						? completedBg
+						: index === current
+							? activeBg
+							: theme.palette.divider;
```
ajhollid: Can we please extract this logic?  Nested ternaries aren't my favourite

client/src/Pages/CreateMonitor/index.tsx
```diff
@@ -249,12 +282,25 @@ const CreateMonitorPage = () => {
...
 	});
-	const { control, watch, handleSubmit, clearErrors } = form;
+	const { control, watch, handleSubmit, clearErrors, trigger } = form;
 
 	useEffect(() => {
 		form.reset(defaults);
 	}, [defaults, form]);
 
+	// Multi-step wizard (uptime create only). Other flows keep the flat form.
+	const isWizard = showTypeSelector;
```
ajhollid: isWizard just shadows showTypeSelector 🤔   I assume we can just remove this one the other pages also implement the wizard flow?

ajhollid (changes requested):
Visually looks good.

There are a couple of bugs to work out.

1.  I can create invalid monitor types by manipulating the form steps.  The form should be reset on changing monitor type.

2.  A blank third screen is rendered for many monitor types, we should skip rendering the advanced third page if it is not used for that monitor type

I think other than that things look good!

client/src/Pages/CreateMonitor/index.tsx
```diff
@@ -249,12 +282,25 @@ const CreateMonitorPage = () => {
...
+	const [currentStep, setCurrentStep] = useState(0);
+	const showStep = (step: number) => !isWizard || currentStep === step;
+
+	const handleNext = async () => {
+		const isValid = await trigger(STEP_FIELDS[currentStep]);
+		if (isValid) {
+			setCurrentStep((step) => Math.min(step + 1, TOTAL_STEPS - 1));
+		}
+	};
+	const handleBack = () => setCurrentStep((step) => Math.max(step - 1, 0));
```
ajhollid: Are these clamps here to avoid issues with keyboard shortcuts?  You should never be able to advance past the page limit or go back past page 0 via the UI controls since there are render guards in place already.

client/src/Pages/CreateMonitor/index.tsx
```diff
@@ -207,6 +213,36 @@ const getGeneralSettingsConfig = (
...
+		"ignoreTlsErrors",
+		"useAdvancedMatching",
+		"matchMethod",
+		"expectedValue",
+		"jsonPath",
+		"geoCheckEnabled",
+		"geoCheckLocations",
+		"geoCheckInterval",
+	],
+];
```
ajhollid: I think there is probably a better approach here.  This is quite fragile as it relies on unstated assumptions about how the zod schema is set up.

For example, if we rename one of the fields in the validation schema/db model, someone needs to rememeber to also update these arrays.

This is an unstated dependency and will be difficult to track down later on.

A more robust approach is probably to derive these fields from the monitor type so there is an explicit dependency.  

In other words, the set of fields to validate should not be the superset of all fields available for `STEP[n]`, it should be the set of exactly the fields available for the selected monitor type.

It also relies on an unstated assumption that the order of the array is the same order as the steps.  Not a huge deal, but this is a code smell, so I'd avoid it and be more explicit.  A map or set is probably a better choice here :thinking:

ajhollid (changes requested):
I think most of the issues are resolved now.

My only real remaining concern is that there is an implicit and unstated dependency between the zod validation schema and the array of field names in the create monitor page.

I think this relationship should be made explicit so that there will not be divergence between the schema and the fields in the array in the future.

client/src/Pages/CreateMonitor/index.tsx
```diff
@@ -207,6 +213,11 @@ const getGeneralSettingsConfig = (
 	return configs[type] || configs.http;
 };
 
+// Whether the wizard's "Advanced" step (TLS / matching / geo) has any content
+// for this monitor type. Types without it skip the step entirely.
+const hasAdvancedStep = (type: MonitorType): boolean =>
+	type === "http" || type === "grpc" || type === "websocket" || supportsGeoCheck(type);
+
```
ajhollid: I don't think we need this anymore since validations now have the steps in the metadata

ajhollid (changes requested):

ajhollid (approved):

PR #3682, opened 2026-05-27
docs: add source installation guide

### LINKED ISSUE #3631: Not using docker?
Is this worth installing from source if one never uses docker and has no docker containers to monitor?
From the get go it looks like this is all docker docker docker.. Really strange, i.m.o., for a monitoring system, since it then becomes a point of failure by itself that it can't monitor.

jhcpeixoto:
Closes #3631

### DIFF
--- a/docs/source-installation.md
+++ b/docs/source-installation.md
@@ -1,0 +1,120 @@
+# Source Installation Guide
+
+This guide explains how to run Checkmate without Docker. It is useful when you
+want the monitoring server and web client to run directly on a host or VM.
+
+Checkmate is split into two applications:
+
+- `server/`: Node.js API and background monitor workers
+- `client/`: React static web application
+
+MongoDB is still required. Redis and the Capture agent are optional.
+
+## Prerequisites
+
+- Node.js 20 or newer
+- npm
+- MongoDB 6 or newer
+- A reverse proxy or static file server for the built client, such as Nginx or
+  Caddy
+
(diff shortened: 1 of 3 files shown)


docs/source-installation.md
```diff
@@ -0,0 +1,120 @@
...
+
+This guide explains how to run Checkmate without Docker. It is useful when you
+want the monitoring server and web client to run directly on a host or VM.
+
+Checkmate is split into two applications:
+
+- `server/`: Node.js API and background monitor workers
+- `client/`: React static web application
+
+MongoDB is still required. Redis and the Capture agent are optional.
```
ajhollid, line 11: Redis is no longer used by this project.

docs/source-installation.md
```diff
@@ -0,0 +1,120 @@
...
+For production, run this command under a process manager such as `systemd`,
+PM2, or your platform supervisor.
+
+## 3. Build and Serve the Client
+
+The client is a static Vite build. Configure the public API URL before building:
+
+```bash
+cd client
+cat > .env.production <<'EOF'
```
ajhollid, line 65: This is destructive and users that aren't familiar with bash commands may not know what this does.  Perhaps a note to users?

ajhollid (changes requested):
Couple of nitpicks, but overall pretty good document 👍

gorkem-bwl:
@jhcpeixoto let me know if you are good with the change requests?

Issue #3657, opened 2026-05-17
Accept status codes in monitors


gorkem-bwl:
Some monitors return 4xx but they are working services.

mayssagl:
i just finished working on it

aroy114:
@gorkem-bwl , can you please provide more context/details about this?

gorkem-bwl:
> [@gorkem-bwl](https://github.com/gorkem-bwl) , can you please provide more context/details about this?

Can you be specific?

harsh-aghara:
Hi @gorkem-bwl,
Just want to confirm my understanding before I start working on this:
Currently, Checkmate treats only `2xx` status codes as "up" and everything else as "down." But some services intentionally return non-2xx codes (like `401`, `403`, etc.) when they're perfectly healthy, for example, an API behind auth that returns `401` for unauthenticated requests. The service is still running fine, but Checkmate marks it as down.

So the ask here is to let users configure a list of **accepted status codes** per monitor, so they can say "hey, `403` is expected for this service, treat it as up."

Is that the correct understanding? And a question before i dive into it:

1. Should users be able to specify individual codes (e.g., `403, 404`), or also ranges (e.g., `400-499`), or both?

Happy to pick this up once confirmed. Thanks!

ajhollid:
Hi @harsh-aghara,

I believe your understanding is correct. 

I think ideally a user should probably be able to select which codes they'd like to accept.

Let's keep it simple and allow them to select from an auto complete list.

In order to keep these PRs manageable and reviewable, please limit your PRs to at most 300 lines or so, and create separate PRs for front and backend functionality. 

Backend PRs should come first as they are critical. 

Thank you for your interest!

Jerotronics:
Hi

I just ran into the same issue so glad to see that it's already being worked on.
(Not sure if it's helpful to know others use need this feature too.)

Thanks.

PR #3690, opened 2026-05-31
Add customUpCodes support for HTTP monitors (Backend)

harsh-aghara:
## Describe your changes

This PR adds backend support for a `customUpCodes` field to HTTP monitors. This feature allows users to specify which non-2xx HTTP status codes (e.g., `401 Unauthorized`, `301 Moved Permanently`) should be treated as a successful "UP" status instead of the default "DOWN". 

**Key Changes:**
- **Schema & Types**: Added `customUpCodes: number[]` to the Mongoose `Monitor` schema and TypeScript types, defaulting to `[]`.
- **Validation**: Updated Zod schemas in `monitorValidation.ts` to validate the new field on monitor creation and updates.
- **Core Logic**: Modified `HttpProvider` to check `customUpCodes` during both success paths and error paths (when `got` throws an `HTTPError`).
- **Geo Checks**: Updated `GlobalPingService` and `geoChecksService` to ensure geographically distributed HTTP checks also respect the custom status codes.
- **Testing**: Added an exhaustive unit test suite in `httpProvider.test.ts` to cover error scenarios, success overrides, empty arrays, and matcher precedence.

Refs #3657

- [ ] I have added i18n support to visible strings


ajhollid (changes requested):
I haven't had a chance to do a full review yet, but off the bat this should use the http codes defined in the http library, these should not be arbitrary or self defined numbers.

The http library already properly defines the codes and should be used for type discrimination. 

https://nodejs.org/api/http.html#httpstatus_codes

harsh-aghara:
> I haven't had a chance to do a full review yet, but off the bat this should use the http codes defined in the http library, these should not be arbitrary or self defined numbers.
> 
> The http library already properly defines the codes and should be used for type discrimination.
> 
> https://nodejs.org/api/http.html#httpstatus_codes

@ajhollid I just pushed a commit that updates the Zod schemas to validate customUpCodes directly against http.STATUS_CODES. It now uses the http library for discrimination and will reject any fake or arbitrary numbers (like 9999) from being saved. I also added full unit tests for this behavior.

Br0wnHammer (changes requested):
Overall, the PR looks solid and clean. I have one concern: 
`got` follows redirects by default, so a 301/302 never shows up as the final status; the redirect target's code does. That means adding 3xx codes to `customUpCodes' won't actually work here (the 301 test only passes because the response is mocked). Should 3xx be supported, or should we drop the 301 example?

server/src/types/monitor.ts
```diff
@@ -60,6 +60,7 @@ export interface Monitor {
 	uptimePercentage?: number;
 	notifications: string[];
 	tags: string[];
+	customUpCodes: number[];
```
ajhollid: Let's properly define an `HttpStatusCode` type and a set we can use to verify

```
export const HttpStatusCodes = Object.keys(http.STATUS_CODES).map(Number);
export const HttpStatusCodeSet = new Set(HttpStatusCodes);
export type HttpStatusCode = number;
```

and use it here to properly type this:

```
	customUpCodes: HttpStatusCode[];
```

server/src/validation/monitorValidation.ts
```diff
@@ -2,6 +2,9 @@ import { z } from "zod";
 import { booleanCoercion, dnsHostnameRegex, dnsServerValidation } from "./shared.js";
 import { GeoContinents } from "@/types/geoCheck.js";
 import { DnsRecordTypes, MonitorMatchMethods, MonitorStatuses, MonitorTypes, PageSpeedStrategies } from "@/types/monitor.js";
+import http from "node:http";
+
+const httpStatusCode = z.number().refine((code) => code.toString() in http.STATUS_CODES, { message: "Must be a valid HTTP status code" });
```
ajhollid: This can now be simplified 

```
const httpStatusCode = z.number().refine((code) => HttpStatusCodeSet.has(code), { message: "Must be a valid HTTP status code" });
```

server/src/service/infrastructure/network/HttpProvider.ts
```diff
@@ -31,11 +31,13 @@ export class HttpProvider implements IStatusProvider<HttpStatusPayload> {
 
 	private handleHttpError<T>(error: unknown, monitor: Monitor): MonitorStatusResponse<T> {
 		if (error instanceof HTTPError || error instanceof RequestError) {
+			const isCustomUp = error instanceof HTTPError && (monitor.customUpCodes ?? []).includes(error.response?.statusCode);
```
ajhollid: This is very hard to read and has a superfluous instance assertion.  Why not do something like the following:

```
const statusCode = error.response?.statusCode;
const statusUp = statusCode !== undefined && monitor.customUpCodes.includes(statusCode);
...
return {
				monitorId: monitor.id,
				teamId: monitor.teamId,
				type: monitor.type,
				status: statusUp,
  				code: statusCode ?? NETWORK_ERROR,
  				message: error.message,
  				responseTime: error.timings?.phases?.total ?? 0,
  				timings: error.timings,
  				payload: null as T,
			};
```

Something along those lines is clearer and much easier to read.

server/src/service/infrastructure/network/HttpProvider.ts
```diff
@@ -106,7 +108,7 @@ export class HttpProvider implements IStatusProvider<HttpStatusPayload> {
 				monitorId: monitor.id,
 				teamId: monitor.teamId,
 				type: monitor.type,
-				status: response.ok && matchResult.ok,
+				status: ((monitor.customUpCodes ?? []).includes(response.statusCode) || response.ok) && matchResult.ok,
```
ajhollid: This difficult to read line then simply becomes

```
status: statusUp && matchResult.ok,
```

It's also worth noting that this is acutally not needed since `got` throws on non 2xx responses.

This is harmless though and we can leave it in in case we decide to change the `got` default behaviour.

server/src/service/infrastructure/globalPingService.ts
```diff
@@ -119,7 +119,7 @@ export class GlobalPingService implements IGlobalPingService {
 		}
 	}
 
-	async pollForResults(measurementId: string, timeoutMs: number = MAX_POLL_TIMEOUT_MS): Promise<GeoCheckResult[]> {
+	async pollForResults(measurementId: string, timeoutMs: number = MAX_POLL_TIMEOUT_MS, customUpCodes: number[] = []): Promise<GeoCheckResult[]> {
```
ajhollid: Correct type can be used here

ajhollid (changes requested):
This needs a bit more work, please see my comments in the review.

Thanks!

harsh-aghara:
> Overall, the PR looks solid and clean. I have one concern:
> `got` follows redirects by default, so a 301/302 never shows up as the final status; the redirect target's code does. That means adding 3xx codes to `customUpCodes' won't actually work here (the 301 test only passes because the response is mocked). Should 3xx be supported, or should we drop the 301 example?

@Br0wnHammer Good catch, you are right. Since got follows redirects by default, it will return the final status code instead of the 3xx code in production.

To handle this, I see two options:
1. Restrict 3xx codes: We can update the Zod validation to reject 3xx codes so users don't get confused (and remove the 301 test example).
3. Set followRedirect: false: We can configure got to stop following redirects. However, this creates a new issue: if the final destination service is actually down, the monitor will still report as "Up" just because it received the initial 3xx response.

I think Option 1 is the safer choice.
What are your thoughts on this @ajhollid ?

server/src/types/monitor.ts
harsh-aghara: Done!

server/src/validation/monitorValidation.ts
harsh-aghara: simplified this, thanks for the review!

server/src/service/infrastructure/network/HttpProvider.ts
harsh-aghara: Done.
broke down code into more readable parts.

server/src/service/infrastructure/globalPingService.ts
harsh-aghara: Done.
Updated globalPingService.ts to use correct type.

server/src/service/infrastructure/network/HttpProvider.ts
harsh-aghara: Done.

ajhollid:
> > Overall, the PR looks solid and clean. I have one concern:
> > `got` follows redirects by default, so a 301/302 never shows up as the final status; the redirect target's code does. That means adding 3xx codes to `customUpCodes' won't actually work here (the 301 test only passes because the response is mocked). Should 3xx be supported, or should we drop the 301 example?
> 
> @Br0wnHammer Good catch, you are right. Since got follows redirects by default, it will return the final status code instead of the 3xx code in production.
> 
> To handle this, I see two options:
> 
> 1. Restrict 3xx codes: We can update the Zod validation to reject 3xx codes so users don't get confused (and remove the 301 test example).
> 2. Set followRedirect: false: We can configure got to stop following redirects. However, this creates a new issue: if the final destination service is actually down, the monitor will still report as "Up" just because it received the initial 3xx response.
> 
> I think Option 1 is the safer choice. What are your thoughts on this @ajhollid ?

Thanks for making changes here, this is looking good :rocket:

I think the issue around redirects isn't straightforward, especially since I don't know how the geo checks handle redirects.

Additionally, `got`'s `responseok` value is depends on the config, ie if `followRedirect` is turned on or off.

What about extracting a utility to deal with status that is independent of `got`'s `response.ok`, instead just returning a value based purely on status codes?

```
export const isStatusUp = (statusCode, customUpCodes) => {
   return statusCode >= 200 && statusCode < 300 || customUpCodes.includes(statusCode)
}
```

something like that, which we can use in both the `HttpProvider` and `globalPing` provider to ensure consistent values?

harsh-aghara:
@ajhollid I've implemented the isStatusUp utility as requested, so the UP status is now determined independently of got.

server/src/service/infrastructure/network/HttpProvider.ts
```diff
@@ -31,12 +31,15 @@ export class HttpProvider implements IStatusProvider<HttpStatusPayload> {
...
+			const statusUp = isStatusUp(statusCode, monitor.customUpCodes);
+
 			return {
 				monitorId: monitor.id,
 				teamId: monitor.teamId,
 				type: monitor.type,
-				status: false,
-				code: error.response?.statusCode ?? NETWORK_ERROR,
+				status: statusUp,
+				code: statusCode ?? NETWORK_ERROR,
```
ajhollid: There is an issue here.

If we accept that there are error codes that may actually be considered as "Up", we need to run the advanced matcher here as well.

If a user sets advanced matching and a custom code, their advanced matcher will be ignored.

ajhollid (changes requested):
The advanced matcher is ignored in the error path.  This is as it should have been before this change, but now that there are codes that could be considered as "up" in the error path, the advanced matcher should be run in those cases.

server/src/service/infrastructure/network/HttpProvider.ts
harsh-aghara: Fixed this in the latest commit.
It now properly runs the advanced matcher for custom up codes too.
Thank you for the review.

server/src/service/infrastructure/network/HttpProvider.ts
```diff
@@ -42,16 +42,67 @@ export class HttpProvider implements IStatusProvider<HttpStatusPayload> {
...
 			return {
 				monitorId: monitor.id,
 				teamId: monitor.teamId,
 				type: monitor.type,
-				status: false,
-				code: error.response?.statusCode ?? NETWORK_ERROR,
-				message: error.message,
-				responseTime: error.timings?.phases?.firstByte ?? error.timings?.phases?.total ?? 0,
+				status: statusUp && matchResult.ok,
+				code: statusCode ?? NETWORK_ERROR,
```
ajhollid: This code is nearly identical to the code in the success branch.  This should be extracted to a private helper function

ajhollid (changes requested):
There's too much code duplication in the error branch, it's nearly identical to the success branch.  Let's extract it to a helper.

server/src/service/infrastructure/network/HttpProvider.ts
harsh-aghara: Done.
Moved this logic into a private buildResponse helper method. Also fixed the MonitorStatusResponse interface so it properly accepts string fallbacks without needing as unknown as T casts in the helper.

ajhollid (approved):
Thanks for making your revisions here, I believe we're good to go now 👍

PR #3694, opened 2026-06-03
Feat: Progress indicator infra & pagespeed

Br0wnHammer:
## Describe your changes

Decoupled the multi-step wizard from the uptime-only type selector via a new `isWizard` flag, so infrastructure and pagespeed creation now show the segmented progress bar and Back/Next stepping (edit mode stays a flat form). Also moved hardware's four alert-threshold fields to wizard step 2 in the Zod schema so they validate on the step where they're actually shown.

```Javascript
<div>{t('add')}</div>
```

Screenshot 2026-06-03 201900
Screenshot 2026-06-03 201917

### DIFF
--- a/client/src/Pages/CreateMonitor/index.tsx
+++ b/client/src/Pages/CreateMonitor/index.tsx
@@ -234,6 +234,9 @@ const CreateMonitorPage = () => {
 	}, [location.pathname]);
 
 	const showTypeSelector = pageType === "uptime" && !isEditMode;
+	// The multi-step wizard applies to every create flow; only the monitor-type
+	// radio is uptime-specific. Edit mode keeps the flat form.
+	const isWizard = !isEditMode;
 	const defaultType: MonitorType =
 		pageType === "pagespeed"
 			? "pagespeed"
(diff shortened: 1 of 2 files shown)


client/src/Pages/CreateMonitor/index.tsx
```diff
@@ -234,6 +234,9 @@ const CreateMonitorPage = () => {
 	}, [location.pathname]);
 
 	const showTypeSelector = pageType === "uptime" && !isEditMode;
+	// The multi-step wizard applies to every create flow; only the monitor-type
+	// radio is uptime-specific. Edit mode keeps the flat form.
+	const isWizard = !isEditMode;
```
ajhollid: Do we need to shadow this variable?

I think it's sufficient to just use the original var, it's pretty clear 🤔

ajhollid (changes requested):
I don't think we need to shadow the edit flag, just one more thing someone has to remember

ajhollid (changes requested):
I don't think we need to shadow isEdit here, just one more thing someone has to remember

ajhollid (approved):

PR #3715, opened 2026-06-08
feat: custom domain support for published status pages

### LINKED ISSUE #3647: custom domain/subdomain for published status pages

It is a problem in fact. Imagine deploying Checkmate internally for a company under `checkmate.svc.internal.example.com`. and you created a status page at `tatus/public/111111` path. This is almost impossible to share with clients/people.

1. It discloses the instance URL to everyone
2. Serving over a sub-path looks ugly.

Like `Uptime-Kuma`, setting a custom domain for a status page (instead of setting slug) would be VERY beneficial.

I am not sure if this is already doable, though.

Thanks

Cincinnatus101010:
## Summary

- Adds optional `customDomain` field to status pages so published pages can be served at a dedicated hostname (e.g. `https://status.acme.com`) instead of only `/status/public/{slug}`
- Introduces `GET /api/v1/status-page/resolve` to look up published pages by custom domain
- Routes custom-domain hosts to the public status page at `/` on the frontend
- Allows CORS from registered published custom domains via dynamic origin validation
- Slug-based public URLs remain fully backward compatible

Happy to make any needed changes

Fixes #3647

## Test plan

- [x] Create status page with custom domain and publish with monitors
- [x] Verify slug URL still works: `/status/public/{slug}`
- [x] Verify custom domain loads at root via `/status-page/resolve`
- [x] Verify CORS allows API calls from custom domain origin
- [x] Verify duplicate custom domains are rejected
- [x] `npm run format-check` passes in `client/` and `server/`
- [x] `npm run build` passes in `client/` and `server/`
- [x] Server test suite passes (57 suites, 1230 tests)

## Deployment note

Production uses standard HTTPS on port 443 (no port in URL). Operators point DNS at their Checkmate instance and configure TLS/reverse proxy for the custom domain. The app handles hostname resolution and CORS for registered domains.

## Screenshots
Screenshot 2026-06-07 at 10 34 59 PM
Screenshot 2026-06-07 at 10 36 03 PM


alioguzhan:
Thanks for the work. 💯

Br0wnHammer:
Hi @Cincinnatus101010 there are a lot of conflicts in the PR. Can you please take care of them?

Cincinnatus101010:
of course fixing now

Cincinnatus101010:
conflicts fixed lmk if there is anything else that needs to be updated

alioguzhan:
Hi @Br0wnHammer @gorkem-bwl,

Any blocker to merge this? we can work together any further modification is needed.

Thanks

Br0wnHammer:
Hi @alioguzhan and @Cincinnatus101010 thankyou for fixing the conflicts, the team will review the PR shortly.

alioguzhan:
Thanks. Waiting to see this in next release. 

Btw, it was all @Cincinnatus101010's work. Kudos to him only.

ajhollid:
Yep I'll have a look at this shortly 👍

server/src/domain/status-pages/status-page.model.ts
```diff
@@ -65,6 +65,14 @@ const StatusPageSchema = new Schema<StatusPageDocument>(
...
 			default: "",
 		},
+		customDomain: {
+			type: String,
+			default: null,
+			lowercase: true,
+			trim: true,
+			sparse: true,
+			unique: true,
+		},
```
ajhollid, line 72: This is incorrect.  The `sparse` constraint here will _not_ skip a `null` value, hence if you create a status page without a custom domain, and therefore a `null` for `customDomain`, you will never be able to create another status page with `customDomain` set to `null`.

```
2026-06-11T16:33:53.746Z error: [unknownService](unknownMethod) E11000 duplicate key error collection: uptime_db.statuspages index: customDomain_1 dup key: { customDomain: null }
```

The tests included in this PR only ever add one status page, so this goes unnoticed if you aren't aware of how `sparse` works.  This PR needs to be tested more thoroughly before it can be accepted.  

The correct solution here is probably to use a partial index on `customDomain` and stop setting the field to `null` explicitly.

server/src/app.ts
```diff
@@ -33,7 +34,7 @@ export const createApp = ({
 
 	app.use(
 		cors({
-			origin: allowedOrigin,
+			origin: createStatusPageCorsOrigin(allowedOrigin, services.statusPagesRepository),
```
ajhollid: This is dangerous, allowing user configurable CORS configuration allows a user to expose the entire API to requests from any origin.  

At the very least this must be scoped only to public status page routes.

server/src/api/middleware/statusPageCorsOrigin.ts
```diff
@@ -0,0 +1,39 @@
...
+
+		statusPagesRepository
+			.findByCustomDomain(customDomain)
+			.then((statusPage) => {
+				callback(null, statusPage.isPublished ? origin : false);
+			})
+			.catch(() => {
+				callback(null, false);
+			});
+	};
```
ajhollid, line 67: This will run on _every_ request for any origin that is not `clientHost`, that seems excessive to me.  This should be cached.

client/src/Utils/statusPageUrl.ts
```diff
@@ -0,0 +1,72 @@
...
+	}
+
+	return `${PUBLIC_STATUS_PAGE_PREFIX}/${statusPage.url}`;
+};
+
+export const getStatusPagePublicUrl = (
+	statusPage: Pick<StatusPage, "url" | "customDomain">
+): string => {
+	if (statusPage.customDomain) {
+		return `https://${statusPage.customDomain}`;
```
ajhollid: What about users not using `https`?

ajhollid (changes requested):
There are some serious issues with this PR with respect to both functionality and security.

The basic concept is good, but the devil is in the details as always.  I believe some more thought needs to go into this before it's ready for final review and merging.

Thanks for working on it so far!

Cincinnatus101010:
Thank you for your review! Making changes now

ajhollid:
> Thank you for your review! Making changes now 
> 

No problem, thank you for working on it!

server/src/domain/status-pages/status-page.model.ts
Cincinnatus101010: Switched to a partial unique index on string customDomain values and stopped writing explicit null when the field is unset. Added a test that creates two status pages without a custom domain. syncIndexes() should handle the index on deploy.

server/src/app.ts
Cincinnatus101010: Fixed, Dynamic CORS only applies on public status page GET/OPTIONS routes like /resolve and slug paths. Everything else still uses the static clientHost.

server/src/api/middleware/statusPageCorsOrigin.ts
Cincinnatus101010: Added a 5 minute cache per domain. This also only runs on public status page routes now, not on every request.

client/src/Utils/statusPageUrl.ts
Cincinnatus101010: Updated to use the protocol from VITE_APP_CLIENT_HOST or window.location instead of hardcoding https.

server/src/api/middleware/statusPageCorsOrigin.ts
```diff
@@ -0,0 +1,59 @@
...
+				const allowed = statusPage.isPublished;
+				cache.set(customDomain, {
+					allowed,
+					expiresAt: Date.now() + CUSTOM_DOMAIN_CORS_CACHE_TTL_MS,
+				});
+				callback(null, allowed ? origin : false);
+			})
+			.catch(() => {
+				callback(null, false);
+			});
```
ajhollid, line 66: This is better, but what about caching negative results?  If I wanted to cause headaches for the owner of the app I could spam unaccepted origins and trigger a DB op for every single origin that doesn't belong to a status page.

How about a cache with a shorter TTL for negative results?

ajhollid (changes requested):
Thanks for making revisions, this is much stronger now.

I think we should also cache negative results for the origin check to prevent excess DB operations.

Other than that looks pretty good to me.

server/src/api/middleware/statusPageCorsOrigin.ts
Cincinnatus101010: setting up now good callout

Cincinnatus101010: Negative lookups (unknown domain or unpublished) are now cached with a 1 minute TTL. Published domains stay at 5 minutes. Added tests for both cases.

ajhollid (approved):
Thanks for your revisions and contribution!

PR #3718, opened 2026-06-09
Add customUpCodes support for HTTP monitors (Frontend)

### LINKED ISSUE #3657: Accept status codes in monitors

harsh-aghara:
## Describe your changes

Implemented the frontend functionality to allow users to define custom accepted HTTP status codes (`customUpCodes`) when creating or editing a monitor. 

Specifically, this PR includes:
- **UI Integration:** Added a new multi-select input field in the `CreateMonitor` form to select accepted HTTP status codes.
- **Constants:** Added a comprehensive list of HTTP status code constants in `Utils/statusCode.ts` for the dropdown options.
- **Type & Validation Updates:** Updated the `Monitor` type and `Validation/monitor.ts` schema to support and validate the new `customUpCodes` array field.
- **Form State:** Updated `useMonitorForm.ts` to handle the new form state correctly.
- **Localization:** Added i18n support in `locales/en.json` for all new visible strings.

Working proof on a test server
screenshot-20260609-210239

Creating a Monitor
screenshot-20260609-205743

Configuring a Monitor
screenshot-20260609-210611

Closes #3657 

```Javascript
<div>{t('add')}</div>
```


client/src/Utils/statusCode.ts
```diff
@@ -74,3 +74,83 @@ export const getStatusCodeValueType = (code: number): ValueType => {
...
+	{ id: 510, name: "510 - Not Extended" },
+	{ id: 511, name: "511 - Network Authentication Required" },
+	{ id: 520, name: "520 - Web Server Returned an Unknown Error (Cloudflare)" },
+	{ id: 521, name: "521 - Web Server Is Down (Cloudflare)" },
+	{ id: 522, name: "522 - Connection Timed Out (Cloudflare)" },
+	{ id: 523, name: "523 - Origin Is Unreachable (Cloudflare)" },
+	{ id: 524, name: "524 - A Timeout Occurred (Cloudflare)" },
+	{ id: 525, name: "525 - SSL Handshake Failed (Cloudflare)" },
+	{ id: 526, name: "526 - Invalid SSL Certificate (Cloudflare)" },
+	{ id: 527, name: "527 - Railgun Error (Cloudflare)" },
```
ajhollid, line 165: There are non-standard HTTP codes in this list that will be rejected.  Please make sure to thoroughly test code before submitting for review.  You should test each one of these codes.

client/src/Validation/monitor.ts
```diff
@@ -64,6 +64,10 @@ const httpSchema = baseSchema.extend({
 		.register(monitorStepRegistry, { step: 2 }),
 	expectedValue: z.string().optional().register(monitorStepRegistry, { step: 2 }),
 	jsonPath: z.string().optional().register(monitorStepRegistry, { step: 2 }),
+	customUpCodes: z
+		.array(z.number().min(100).max(599))
```
ajhollid: This is inconsistent with the backend validation.

Two options

1.  Loosen the backend validation if the goal is to match non-standard codes
2. Tighten the frontend validation to only accept standard HTTP codes.

This is a design decision, so I'll leave this with you.  I don't know what you want to achieve here so please explain and we can make a decision.

If the goal is to allow non-standard HTTP codes, then simply use the same validation on the backend.  If the goal is to only allow standard HTTP codes, then tighten the FE validation.

ajhollid (changes requested):
There is a mismatch between frontend validation and backend validation.  Frontend accepts non-standard HTTP codes, while the backend does not.

This requires a design decision. I have no horse in this race so I will leave the decision with you, but whatever you decide it must be consistent between the FE and the BE.

client/src/Utils/statusCode.ts
harsh-aghara: my bad 😅 I only spot checked the standard codes and naively assumed the rest would pass the backend.
I'll make sure to test all edge cases next time. 🫡 Left a note below on how we can actually fix this!

harsh-aghara:
@ajhollid Let's go with allowing the non standard codes. Since users may monitor servers behind cloudflare, it's useful to support the 52x errors.

I'll update the backend to allow the non standard codes we currently show in the frontend UI (499, 520-527) rather than relying on node's default list. I'll also update the frontend validation to strictly match that exact list so both sides are in sync.

ajhollid:
> @ajhollid Let's go with allowing the non standard codes. Since users may monitor servers behind cloudflare, it's useful to support the 52x errors.
> 
> I'll update the backend to allow the non standard codes we currently show in the frontend UI (499, 520-527) rather than relying on node's default list. I'll also update the frontend validation to strictly match that exact list so both sides are in sync.

Sure, sounds good to me :+1:

ajhollid:
@harsh-aghara just a heads up, there is a big reorg PR coming for the back end, so probably best to wait for that before making your back end changes.

ajhollid:
@harsh-aghara reorg PR is in, you can go ahead and work on the backend now whenever you are ready.  Things have moved around significantly, so you may have to search around for a bit as you get used to the new structure.  Thanks!

ajhollid (approved):
Looks good, thanks for making changes here!

PR #3723, opened 2026-06-10
Feat/customcss status page

### LINKED ISSUE #2564: Add Custom CSS Section to Status Page
Description:
Implement a custom CSS section in the Status Page creation interface to allow users to add custom styling to their status pages.

Requirements:
1. Add a new "Custom CSS" section to the Status Page Settings tab after the appearance section
2. Implement a simple code editor with CSS syntax highlighting only
3. Update the StatusPage model/schema to include a customCSS field (string type)
4. Add validation and sanitization for the CSS input to prevent XSS attacks
5. Update the API endpoints and rendering logic to handle and apply the custom CSS

Buco7854:
## Describe your changes

Finishes the custom CSS feature for status pages. The `customCSS` field already
existed on the model and types, but there was no way to actually use it. The
form never sent it, the server validator stripped it, and nothing rendered it.

This PR wires it end to end:

- **Server**: `customCSS` is now accepted by the create/update status page
  validation, and the OpenAPI spec is regenerated to match.
- **Client**: new "Custom CSS" textarea on the appearance step of the status
  page form, and the value is included in the submitted form data.
- **Rendering**: the saved CSS is injected as a `<style>` tag in
  `BaseStatusPage`, so it applies to all four themes. React sets style children
  via `textContent`, so the string is never HTML parsed and there is no
  `</style>` breakout / XSS vector.
- **Preview isolation**: the admin preview now renders the real public URL in
  an iframe inside the browser chrome instead of rendering the page inline in
  the admin document. Without this, the injected CSS leaked into the admin UI
  around the preview. As a bonus the preview is now pixel-identical to what
  visitors see. Auth for previewing unpublished pages keeps working since the
  same-origin iframe shares localStorage. The preview was already showing the
  live saved page before, so this changes how it renders, not what it shows.
- `TextInput` got a tiny fix to support `multiline`. It hardcoded a 34px input
  height, which squashed textareas, and this is the first multiline usage.
- Added validation tests for the new field (accepts, clears, rejects oversized
  and non-string values).

To test: create or edit a status page, paste something obvious like
`li { outline: 3px dashed red; }` into the Custom CSS box on the appearance
step, save, and open the public URL or the admin preview. Heads up that page
backgrounds need `!important` because the theme paints its background on a
wrapper div with an inline style.

## Fixes 

Fixes #2564

- [x] (Do not skip this or your PR will be closed) I deployed the application locally (also Docker).
- [x] I have added i18n support to visible strings (instead of `<div>Add</div>`)

## Screenshots:

Before:

adminformbefore
adminpreviewcssoff
publiccssoff

After:

adminformwith
adminpreviewwith
publicwith


Buco7854:
Testing it to local will force push once all is valid. Putting as a Draft for now.

Buco7854:
Tested it everything is fine.

Buco7854:
For a quick docker test you can also use : ghcr.io/buco7854/checkmate-backend-mono:customcss

gorkem-bwl:
@Buco7854  Do you think we should also document the available CSS selectors, classes and elements that can be customized through CSS overrides? We could surface this in an info box, help section, or it would just go into Checkmate documentation. Without that guidance, it may not be obvious for developers which selectors they should target when customizing the status page.

ajhollid:
@gorkem-bwl @Buco7854 I don't think this feature is used at all on the backend at the moment is it? I think the field is vestigial 🤔 

Accepting custom css is a dangerous proposition at it opens an attack vector, so we have to mitigate that before we do anything with the css

gorkem-bwl:
> @gorkem-bwl @Buco7854 I don't think this feature is used at all on the backend at the moment is it? I think the field is vestigial 🤔
> 
> Accepting custom css is a dangerous proposition at it opens an attack vector, so we have to mitigate that before we do anything with the css

I am not sure about the backend side really.

When it comes to attack vectors only admins can add custom CSS, the risk is much lower than allowing arbitrary users to do it. In fact anything admin is doing is his/her responsibility so if we add a warning there I think we should be safe (eg something like: 

`
Warning: Custom CSS can significantly alter the appearance and behavior of your status page. Use only trusted CSS and review any external resources it references. You are responsible for the content and effects of any custom CSS you apply.
`

Buco7854:
@ajhollid @gorkem-bwl 

> @Buco7854 Do you think we should also document the available CSS selectors, classes and elements that can be customized through CSS overrides? We could surface this in an info box, help section, or it would just go into Checkmate documentation. Without that guidance, it may not be obvious for developers which selectors they should target when customizing the status page.

Currently status page style isn't really using particular classes etc, someone adding custom CSS will eventually open their developer tool to search for the element they want to apply the style on etc so I'd say it's not required.

> @gorkem-bwl @Buco7854 I don't think this feature is used at all on the backend at the moment is it? I think the field is vestigial 🤔

There wasn't any UI to set it and it was not rendered, otherwise it was already mostly working.

> Accepting custom css is a dangerous proposition at it opens an attack vector, so we have to mitigate that before we do anything with the css

We could add stricter CSP so img-src, font-src etc aren't loaded from external sources, even stricter there could be some parsing done backend side to remove any dangerous things like `@import` etc, but they restrict what you could legitimately want to do. I'd say since it is admin only (just pushed a fix for that actually) it is relatively safe and can stay as if but at the that is your calls, can do it.

Buco7854:
@gorkem-bwl @ajhollid I was kind of asking for your opinion if you dont mind.

ajhollid:
> @gorkem-bwl @ajhollid I was kind of asking for your opinion if you dont mind.

I don't really have any opinions on the frontend implementation of this as that's not my field of expertise, but the CSS has to be very strictly and carefully sanitized on the backend so we don't allow any malicious CSS to be accepted.

The current validation

```
customCSS: z.string().max(100000, "Custom CSS must be at most 100000 characters").optional()
```

is not sufficient.

User submitted CSS must also be correctly scoped to the status page, we don't want them applying styles that are out of scope.

To be honest, this feature is a huge can of worms I'm not personally interested in opening, but if it is going to be worked on then it must be done very carefully.

Buco7854:
> > @gorkem-bwl @ajhollid I was kind of asking for your opinion if you dont mind.
> 
> I don't really have any opinions on the frontend implementation of this as that's not my field of expertise, but the CSS has to be very strictly and carefully sanitized on the backend so we don't allow any malicious CSS to be accepted.
> 
> The current validation
> 
> ```
> customCSS: z.string().max(100000, "Custom CSS must be at most 100000 characters").optional()
> ```
> 
> is not sufficient.
> 
> User submitted CSS must also be correctly scoped to the status page, we don't want them applying styles that are out of scope.
> 
> To be honest, this feature is a huge can of worms I'm not personally interested in opening, but if it is going to be worked on then it must be done very carefully.

Could you give me an example of malicious CSS you'd like sanitized? Just so we are on the same page about the potential attack surface: the CSS is injected as a text node in a single <style> tag on the status page, so this isn't script injection, and anything that isn't valid CSS is just ignored by the browser. Nothing other than CSS can be executed.

The only thing that I cans see as a "maybe" is `url()` and `@import`, which make the visitor's browser do a request and so leak their IP and User-Agent. Beyond that, url() with attribute selectors can read DOM attribute values, though here the page only renders our already public status page so it doesnt mater. If that is what is causing you issue the scope, I'm happy to update the PR to strip @import and off-origin url(). I'd just like to confirm the scope before working on it.

On scoping: we only add a <style> tag with the custom CSS to that specific status page (no other page can be affected, it is not global to checkmate or to mulitple page). Editing status page is also now admin/superadmin only on the backend and frontend.

ajhollid:
> > > @gorkem-bwl @ajhollid I was kind of asking for your opinion if you dont mind.
> > 
> > 
> > I don't really have any opinions on the frontend implementation of this as that's not my field of expertise, but the CSS has to be very strictly and carefully sanitized on the backend so we don't allow any malicious CSS to be accepted.
> > The current validation
> > ```
> > customCSS: z.string().max(100000, "Custom CSS must be at most 100000 characters").optional()
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > is not sufficient.
> > User submitted CSS must also be correctly scoped to the status page, we don't want them applying styles that are out of scope.
> > To be honest, this feature is a huge can of worms I'm not personally interested in opening, but if it is going to be worked on then it must be done very carefully.
> 
> Could you give me an example of malicious CSS you'd like sanitized? Just so we are on the same page about the potential attack surface: the CSS is injected as a text node in a single <style> tag on the status page, so this isn't script injection, and anything that isn't valid CSS is just ignored by the browser. Nothing other than CSS can be executed.
> 
> The only thing that I cans ee as a "maybe" is url() and @import, which make the visitor's browser do a request and so leak their IP and User-Agent. Beyond that, url() with attribute selectors can read DOM attribute values, though here the page only renders our already public status page so it doesnt mater. If that is what is causing you issue the scope, I'm happy to update the PR to strip @import and off-origin url(). I'd just like to confirm the scope before working on it.
> 
> On scoping: we only add a <style> tag with the custom CSS to that specific status page (no other page can be affected, it is not global to checkmate or to mulitple page). Editing status page is also now admin/superadmin only on the backend and frontend.

Anything that triggers a network request should be sanitized, ie `url()` and `@import` as you mentioned.  

Anything `position: fixed` or `absolute` should probably be sanitized too.

There's lots of malicious but not dangerous things to think about too.  What if a user tries to set the "down" color to a color one would expect to indicate "up"?  What about setting infinite animations?  Massive/recursive blurs?  I don't really know here, as this is not my domain, perhaps @gorkem-bwl can weigh in here.  These are probably less important since this is an admin only feature, and if they want to ruin their own status pages then :shrug:

Buco7854:
> > > > @gorkem-bwl @ajhollid I was kind of asking for your opinion if you dont mind.
> > > 
> > > 
> > > I don't really have any opinions on the frontend implementation of this as that's not my field of expertise, but the CSS has to be very strictly and carefully sanitized on the backend so we don't allow any malicious CSS to be accepted.
> > > The current validation
> > > ```
> > > customCSS: z.string().max(100000, "Custom CSS must be at most 100000 characters").optional()
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > is not sufficient.
> > > User submitted CSS must also be correctly scoped to the status page, we don't want them applying styles that are out of scope.
> > > To be honest, this feature is a huge can of worms I'm not personally interested in opening, but if it is going to be worked on then it must be done very carefully.
> > 
> > 
> > Could you give me an example of malicious CSS you'd like sanitized? Just so we are on the same page about the potential attack surface: the CSS is injected as a text node in a single <style> tag on the status page, so this isn't script injection, and anything that isn't valid CSS is just ignored by the browser. Nothing other than CSS can be executed.
> > The only thing that I cans ee as a "maybe" is url() and @import, which make the visitor's browser do a request and so leak their IP and User-Agent. Beyond that, url() with attribute selectors can read DOM attribute values, though here the page only renders our already public status page so it doesnt mater. If that is what is causing you issue the scope, I'm happy to update the PR to strip @import and off-origin url(). I'd just like to confirm the scope before working on it.
> > On scoping: we only add a <style> tag with the custom CSS to that specific status page (no other page can be affected, it is not global to checkmate or to mulitple page). Editing status page is also now admin/superadmin only on the backend and frontend.
> 
> Anything that triggers a network request should be sanitized, ie `url()` and `@import` as you mentioned.
> 
> Anything `position: fixed` or `absolute` should probably be sanitized too.
> 
> There's lots of malicious but not dangerous things to think about too. What if a user tries to set the "down" color to a color one would expect to indicate "up"? What about setting infinite animations? Massive/recursive blurs? I don't really know here, as this is not my domain, perhaps @gorkem-bwl can weigh in here. These are probably less important since this is an admin only feature, and if they want to ruin their own status pages then 🤷

Ok, I'll sanitize anything that can trigger a request from the browser. About position: fixed/absolute, I personally wouldn't touch it, I don't see the point and it's typically something you'd use legitimately.
For the other cosmetic things I'm with you: if an admin makes their own page unreadable, that's kind of on them imo.
I'll update when I can, it's night time here so I'll probably take a look tomorrow.
Thanks a lot for the feedback 😄

Buco7854:
Well ended up doing it tonight. Any external url is now rejected, went for a full reject rather than sanitization so we're upfront with the user.

server/src/utils/customCss.ts
```diff
@@ -0,0 +1,23 @@
...
+// protocol-relative. Anchoring on the start of the target keeps data: and
+// relative URLs allowed, including data: SVG that embeds the SVG namespace URL.
+const EXTERNAL_REFERENCE = /(?:@import\b|(?:url|image-set|-webkit-image-set|image|cross-fade|-webkit-cross-fade)\s*\(\s*['"]?\s*(?:https?:|\/\/))/i;
+
+export const cssReferencesExternalResource = (css?: string | null): boolean => {
+	if (typeof css !== "string" || css === "") {
+		return false;
+	}
+	return EXTERNAL_REFERENCE.test(decodeForScan(css));
+};
```
ajhollid: Unless I'm misunderstanding how this works, this hand-rolled solution allows URLs with escape chars to bypass the sanitization:

```
const decodeForScan = (css) => {
  return css
    .replace(/\/\*[\s\S]*?\*\//g, "")
    .replace(/\\([0-9a-fA-F]{1,6})\s?|\\([\s\S])/g, (_match, hex, char) => {
      if (!hex) {
        return char;
      }
      const codePoint = parseInt(hex, 16);
      return codePoint > 0 && codePoint <= 0x10ffff
        ? String.fromCodePoint(codePoint)
        : "�";
    });
};

const EXTERNAL_REFERENCE =
  /(?:@import\b|(?:url|image-set|-webkit-image-set|image|cross-fade|-webkit-cross-fade)\s*\(\s*['"]?\s*(?:https?:|\/\/))/i;

const cssReferencesExternalResource = (css) => {
  if (typeof css !== "string" || css === "") {
    return false;
  }
  return EXTERNAL_REFERENCE.test(decodeForScan(css));
};

const baseline = 'body{background:url("https://bypass/malicious")}';
const hexEscape =
  'body{background:url("htt\a ps://webhook.site/29d8a9b0-2092-4dfe-8377-51c59f150b3d")}';
const newLine =
  'body{background:url("htt\\\nps://webhook.site/29d8a9b0-2092-4dfe-8377-51c59f150b3d")}';

for (const [name, css] of [
  ["baseline", baseline],
  ["hex escaped \\a", hexEscape],
  ["new line", newLine],
]) {
  console.log(`${name}`);
  console.log(`${JSON.stringify(css)}`);
  console.log(JSON.stringify(decodeForScan(css)));
  console.log(cssReferencesExternalResource(css));
  console.log("--------------------------------------------------");
}
```

Run this snippet, it doesn't flag either of those cases:

```
baseline
"body{background:url(\"https://bypass/malicious\")}"
"body{background:url(\"https://bypass/malicious\")}"
true
--------------------------------------------------
hex-escaped \a
"body{background:url(\"htta ps://webhook.site/29d8a9b0-2092-4dfe-8377-51c59f150b3d\")}"
"body{background:url(\"htta ps://webhook.site/29d8a9b0-2092-4dfe-8377-51c59f150b3d\")}"
false
--------------------------------------------------
line continuation
"body{background:url(\"htt\\\nps://webhook.site/29d8a9b0-2092-4dfe-8377-51c59f150b3d\")}"
"body{background:url(\"htt\nps://webhook.site/29d8a9b0-2092-4dfe-8377-51c59f150b3d\")}"
false
```

The browser will happily make those requests given those values that skip sanitization.

Image

I don't know enough about CSS to feel confident hand rolling a validation here, it's probably best to outsource this to a well tested and maintained library

ajhollid (changes requested):

server/src/utils/customCss.ts
Buco7854: I wasn't sure if you would have been fine with an other dependency thats why I avoided it. If you are fine with it I'll use https://github.com/csstools/tokenizer it's actually already pulled by an other dep so it doesn't change anything.

I'll add CSP for status page so the client refuses to load those anyway if a different domain somehow manage to slip by.

Buco7854:
normlaly escape that are innofensive should still pass but anything that can be "dangerous" doesnt. In addition I updated CSP for status pages so anything that is not google fonts or current domain is blocked.

server/src/api/middleware/statusPageDocumentCsp.ts
```diff
@@ -0,0 +1,19 @@
+import type { NextFunction, Request, Response } from "express";
+
+const PUBLIC_STATUS_PAGE_DOCUMENT_PREFIX = "/status/public";
```
ajhollid, line 4: What about custom domains?  The tightened CSP won't apply to those 🤔

client/src/Validation/statusPage.ts
```diff
@@ -58,7 +58,11 @@ export const statusPageSchema = z.object({
 	showUptimePercentage: z.boolean().register(statusPageStepRegistry, { step: 1 }),
 	showAdminLoginLink: z.boolean().register(statusPageStepRegistry, { step: 1 }),
 	showInfrastructure: z.boolean().register(statusPageStepRegistry, { step: 1 }),
-	customCSS: z.string().optional().register(statusPageStepRegistry, { step: 1 }),
+	customCSS: z
+		.string()
+		.max(100000, "Custom CSS must be at most 100000 characters")
+		.optional()
+		.register(statusPageStepRegistry, { step: 1 }),
```
ajhollid, line 69: This validation doesn't match the server side validation, a user will be able to submit CSS the client considers valid, but the server will reject it

ajhollid (changes requested):
Couple of changes here and I think we're good to go.

Buco7854:
Fixed all 2 points. Should be good now.

ajhollid:
@Buco7854 this all looks good now, thanks for making changes.

One thing I'd like to confirm though with you and @gorkem-bwl .  This PR makes create/update status pages an admin only operation.  Is that the correct level of access control for status pages?

Buco7854:
Can't recall the exact thing that was or not displayed but there was definitely some admin gating (only on UI) so I suppose that is intended that only admin can edit status pages.

ajhollid (approved):

PR #3726, opened 2026-06-12
Fix SSL cert check on non-standard ports and login user enumeration

### LINKED ISSUE #2768: Login error handling issues

### LINKED ISSUE #3646: SSL/TLS certificate expiry not checked on non standard https port
**Describe the bug**
Hello, First, thanks a lot for your work.
I'm starting with checkmate (to replace checkcle that is having an issue on ressources, especially CPU) and I noticed the certificate expiry is not checked on https service with non-standard port.

**To Reproduce**
1. Go to Checkmate->Uptime->Create a monitor with a non standard https port (eg https://checkmate.example.org:54321
2. wait that the monitor is initialized and some checks are reported
3. Click on the monitor created
4. See error: Certificate Expiry is NA

**Expected behavior**
I would have expected the Certificate Expiry date to have been checked

**Screenshots**

Image

 - Browser Chrome
 - Version  147.0.7727.138

**Additional context**
Checkmate 3.7.1 running on Docker on a Synology DS1522+ (deployed using docker-compose mono with separated mongodb)

sawirricardo:
## Summary

Two small fixes for confirmed open issues:

### 1. SSL/TLS certificate expiry on non-standard HTTPS ports (#3646)

`fetchMonitorCertificate` was only passing the hostname to `ssl-checker`, ignoring the port specified in the monitor URL. This meant certificates on non-standard HTTPS ports (e.g., `https://example.com:8443`) were never checked and always showed "N/A".

**Fix:** Extract port from the monitor URL and pass it to the ssl-checker via the `SSLOptions.port` parameter (already supported by the library).

### 2. Login error handling / user enumeration (#2768)

The login flow returned different error messages and status codes depending on whether the email existed:
- Non-existent email → "User not found" (404)
- Wrong password → "Incorrect password" (401)

This leaks information about which emails are registered.

**Fix:** Catch the "user not found" case and return the same generic "Incorrect email or password" (401) message for both scenarios.

Fixes #3646
Fixes #2768

### DIFF
--- a/server/src/api/controllers/controllerUtils.ts
+++ b/server/src/api/controllers/controllerUtils.ts
@@ -7,7 +7,8 @@ type SSLCheckerType = typeof sslChecker;
 export const fetchMonitorCertificate = async (checker: SSLCheckerType, monitor: Monitor): Promise<SSLDetails> => {
 	const monitorUrl = new URL(monitor.url);
 	const hostname = monitorUrl.hostname;
-	const cert = await checker(hostname);
+	const port = monitorUrl.port ? parseInt(monitorUrl.port, 10) : undefined;
+	const cert = await checker(hostname, { ...(port ? { port } : {}) });
 	if (cert?.validTo === null || cert?.validTo === undefined) {
 		throw new Error("Certificate not found");
 	}
--- a/server/src/domain/users/user.service.ts
+++ b/server/src/domain/users/user.service.ts
@@ -220,12 +220,17 @@ export class UserService implements IUserService {
 
 	loginUser = async (email: string, password: string) => {
 		// Check if user exists
-		const user = await this.usersRepository.findByEmail(email);
+		let user;
+		try {
+			user = await this.usersRepository.findByEmail(email);
+		} catch {
+			throw new AppError({ message: "Incorrect email or password", service: SERVICE_NAME, status: 401 });
+		}
 		// Compare password
 		const match = await bcrypt.compare(password, user.password);
 
 		if (match !== true) {
-			throw new AppError({ message: "Incorrect password", service: SERVICE_NAME, status: 401 });
+			throw new AppError({ message: "Incorrect email or password", service: SERVICE_NAME, status: 401 });
 		}
 
 		// Remove password from user object.  Should this be abstracted to DB layer?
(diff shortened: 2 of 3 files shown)


server/src/domain/users/user.service.ts
```diff
@@ -220,12 +220,17 @@ export class UserService implements IUserService {
 
 	loginUser = async (email: string, password: string) => {
 		// Check if user exists
-		const user = await this.usersRepository.findByEmail(email);
+		let user;
+		try {
+			user = await this.usersRepository.findByEmail(email);
+		} catch {
+			throw new AppError({ message: "Incorrect email or password", service: SERVICE_NAME, status: 401 });
+		}
```
ajhollid, line 228: This doesn't resolve the enumeration problem.  Attackers can still determine if they have found a correct email credential, as bcrypt is still awaited after the user lookup.

If you want to fully resolve the enumeration issue then a dummy value needs to be hashed on the user not found path.

Additionally, this catch block is too broad, it classifies _all_ errors thrown by the repository as email/password errors.  This should be narrowed appropriately.

ajhollid (changes requested):
This doesn't actually resolve the enumeration issue noted in the PR title, please see comments.

In future, one issue per PR please 👍

server/src/api/controllers/controllerUtils.ts
```diff
@@ -7,7 +7,8 @@ type SSLCheckerType = typeof sslChecker;
 export const fetchMonitorCertificate = async (checker: SSLCheckerType, monitor: Monitor): Promise<SSLDetails> => {
 	const monitorUrl = new URL(monitor.url);
 	const hostname = monitorUrl.hostname;
-	const cert = await checker(hostname);
+	const port = monitorUrl.port ? parseInt(monitorUrl.port, 10) : undefined;
+	const cert = await checker(hostname, { ...(port ? { port } : {}) });
```
ajhollid, line 11: This is a pretty convoluted way of expressing that port should not be added here if not specificed.

```
checker(hostname, port ? { port } : undefined)
```

seems clearer to me 🤔

ajhollid (changes requested):

gorkem-bwl:
Closing as inactive

PR #3741, opened 2026-06-27
test: cover maintenance window monitor edits

sanmaxdev:
## Summary
- add a real-Mongo regression test for editing a maintenance window's monitor membership
- verify removed monitors no longer match the window
- verify kept and newly added monitors both resolve back to the edited window

## Context
Refs #3551. The service layer already maps the edit payload's `monitors` field to `monitorIds`; this adds repository-level coverage so the Mongo persistence path is locked down too.

## Test Plan
- `TMPDIR=/home/ubuntu/tmp npm test -- --runTestsByPath test/integration/maintenanceWindowRepository.test.ts test/unit/services/maintenanceWindowService.test.ts --runInBand`
- `npx prettier --check test/integration/maintenanceWindowRepository.test.ts`
- `npm run build`


ajhollid (approved):
Thanks for taking the time to help write tests!

PR #3756, opened 2026-07-05
Implement Request Headers on HTTP monitors

### LINKED ISSUE #3755: Add HTTP Headers options for HTTP Monitors
If the url requires any kind of header, like authentication. We can't add it. 

I think an option to add custom headers to the requested URL should do most of the job related.

I though a solution similar to what we do on postman.

**Additional context**
No context needed

danjourno-dev:
## Describe your changes

Implements Request Headers on HTTP monitors. Note, as secret was stored as plain test, I store the headers as plain text. I can change this to encrypted at rest if you'd prefer.

Fixes #3755 

```Javascript
<div>{t('add')}</div>
```

New config section added to HTTP config
image


client/src/Validation/monitor.ts
```diff
@@ -80,6 +80,10 @@ const httpSchema = baseSchema.extend({
 		.array(httpStatusCode)
 		.optional()
 		.register(monitorStepRegistry, { step: 2 }),
+	headers: z
+		.array(z.object({ key: z.string(), value: z.string() }))
+		.optional()
+		.register(monitorStepRegistry, { step: 2 }),
```
ajhollid, line 98: This is validation is not consistent with the back end validation.  Back end validation requires a value, so if I enter a key for a header but not a value, I pass front end validation but fail back end validation 

Image

client/src/Pages/CreateMonitor/index.tsx
```diff
@@ -1321,6 +1323,111 @@ const CreateMonitorPage = () => {
...
+													sx={{ visibility: "hidden" }}
+													aria-hidden
+												>
+													<Trash2 size={16} />
+												</IconButton>
+											</Stack>
+										)}
+										{headers.map((header, index) => (
+											<Stack
+												key={index}
```
ajhollid: Please don't key on index.  [Stable keys](https://react.dev/learn/rendering-lists) should be used.

server/src/service/network/HttpProvider.ts
```diff
@@ -162,8 +162,16 @@ export class HttpProvider implements IStatusProvider<HttpStatusPayload> {
 			throw new Error("URL is required for HTTP monitor");
 		}
 
+		// Build request headers: user-defined headers first, then secret-derived
+		// Authorization (secret wins to preserve existing behavior).
+		const headerMap: Record<string, string> = {};
+		for (const h of monitor.headers ?? []) {
+			if (h.key) headerMap[h.key] = h.value;
```
ajhollid, line 169: This will overwrite an existing header if a user adds duplicate headers.  There is nothing stopping them from doing so, and no indication that the last header in the list is the one that will be written.

client/src/Pages/CreateMonitor/index.tsx
```diff
@@ -1321,6 +1323,111 @@ const CreateMonitorPage = () => {
...
+													</FieldLabel>
+												</Stack>
+												<IconButton
+													size="small"
+													sx={{ visibility: "hidden" }}
+													aria-hidden
+												>
+													<Trash2 size={16} />
+												</IconButton>
+											</Stack>
```
ajhollid: This is hard to use on mobile; suggest stacking vertically on small screens.  Please see breakpoints used in other UI components.

Image

ajhollid (changes requested):
This looks pretty good in principle.  There are some issues that need to be resolved, but they are mostly minor.  Thanks for working on the issue, please see my comments in the code review!

danjourno-dev:
@ajhollid Bear with me. My rebase looks to have gone wrong. I'll fix shortly

ajhollid:
No rush, ping me when you're good to go

server/src/service/network/HttpProvider.ts
danjourno-dev: Duplicate headers have now been prevented from being configured

client/src/Validation/monitor.ts
danjourno-dev: Empty values are now not allowed by the client

client/src/Pages/CreateMonitor/index.tsx
danjourno-dev: Fixed

client/src/Pages/CreateMonitor/index.tsx
danjourno-dev: Layout updated to support mobile

danjourno-dev:
@ajhollid It's ready for you

client/src/Pages/CreateMonitor/index.tsx
```diff
@@ -1321,6 +1336,111 @@ const CreateMonitorPage = () => {
...
+							))}
+							<Button
+								type="button"
+								variant="outlined"
+								color="secondary"
+								onClick={() => appendHeader({ key: "", value: "" })}
+							>
+								{t("pages.createMonitor.form.advanced.headers.option.addButton")}
+							</Button>
+							{(errors as Record<string, { root?: { message?: string } }>).headers?.root
```
ajhollid, line 1427: Let's remove this cast and type this properly; we shouldn't have to coerce the TS compiler to give us the proper types by casting.

```
const headersRootError = form.getFieldState("headers", form.formState).error?.root;
```

You should be able to properly extract the errors from the form state without casting.

That also allows us to remove the duplication here when extracting errors.

client/src/Validation/monitor.ts
```diff
@@ -80,6 +80,22 @@ const httpSchema = baseSchema.extend({
 		.array(httpStatusCode)
 		.optional()
 		.register(monitorStepRegistry, { step: 2 }),
+	headers: z
+		.array(
+			z.object({
+				key: z.string().min(1, "Header name is required"),
+				value: z.string().min(1, "Header value is required"),
```
ajhollid, line 87: This validation works, but it _will_ allow for invalid headers.  I think we probably need more robust validation here. As is, if a user enters something like a colon into their header without knowing it is invalid, validation will pass, the monitor will be scheduled, and the check will be executed, resulting in a "down" monitor, which may or may not be accurate.

Let's tighten this up so invalid headers won't be accepted. I think that's a more reasonable approach.

client/src/locales/ar.json
```diff
@@ -1,4 +1,4 @@
-{
+﻿{
```
ajhollid, line 1: This diff appears to be an editor artifact, I assume from MS Visual Studio as there is also a `/vs` added to the gitignore. 

There's a byte order mark being added to all of these files, if we curl and read the raw bytes we can see it there:

```
curl -s https://raw.githubusercontent.com/danjourno-dev/Checkmate/39fae9d81e9bd64097123143ef49c2597369b290/client/src/locales/ar.json | xxd | head
00000000: efbb bf7b 0a09 2263 6f6d 6d6f 6e22 3a20  ...{.."common": 
...
```

These artifacts cause endless headaches with linters/formatters, so let's please remove all of these from the translation files.  It appears each one has a BOM added.

ajhollid (changes requested):
Thanks for making changes.  There's a few other things that I think are worth addressing, most importantly that invalid headers can be accepted right now, which could lead to misleading monitor status.

There are some editor artifcats that have been committed, we should really remove those to keep the PR clean.  

Other than that I think we're good to go here once those are resolved!

ajhollid:
Closing as inactive

PR #3763, opened 2026-07-09
fix(status-page): inject sanitized customCSS on the public page

### LINKED ISSUE #3762: Custom CSS on status pages is not applied and lacks stable selectors
### Summary

The `customCSS` field on status pages is saved to the database but has no effect on the public page. Even if it were applied, the current markup exposes no stable selectors, so operators cannot target elements.

### Details

**1. `customCSS` is never injected into the public status-page document.**

- Server side: the field is defined in `server/src/db/models/StatusPage.ts` (`customCSS`), read by `server/src/repositories/status-pages/MongoStatusPagesRepository.ts`, and typed in `server/src/types/statusPage.ts`.
- Client side: `grep -rn 'customCSS\|<style' client/src/Pages/StatusPage --include='*.tsx' --include='*.ts'` returns **zero hits**. There is no `<style>{customCSS}</style>` (or equivalent stylesheet injection) anywhere in the public status page render path.

Result: valid CSS an operator saves is stored but never rendered.

**2. Status page markup lacks stable selectors.**

- `grep -rn 'className=\|class=\|data-' client/src/Pages/StatusPage/Status --include='*.tsx'` returns **zero hits**.
- The status page (`client/src/Pages/StatusPage/Status/**`, including all themes under `themes/refined`, `themes/modern`, `themes/bold`, `themes/editorial`, `themes/minimal`, and `themes/shared`) is built with ~77 MUI `styled(...)` / `sx={...}` components, which emit hash-based class names (e.g. `.css-1abcd2e`) that change between builds and per component instance.

Result: even after fixing (1), operators would have nothing reliable to target.

### Acceptance criteria

- [ ] `customCSS` is injected into the public status page document (e.g. as a sanitized `<style>` tag in the page head, respecting the existing CSP).
- [ ] Existing external-resource protection (CSS-syntax tokenizer rejecting `@import`, `url(...)`, etc.) still applies.
- [ ] Stable, documented selectors are added to the primary status-page regions and elements — options include:
  - Semantic `data-checkmate-*` attributes on landmark regions (header, hero, overall status, monitor list, monitor card, incident timeline, footer, etc.), OR
  - Stable `className` values (not MUI hashes) on the same regions.
- [ ] Selectors are applied consistently across all themes (Refined, Modern, Bold, Editorial, Minimal).
- [ ] Public-facing docs list the available selectors with a copyable example.
- [ ] The admin preview iframe reflects the applied CSS so operators can iterate before publishing.

### Scope note

This is a follow-up to the initial custom CSS work. Backend validation, storage, length cap, external-resource rejection, and CSP tightening are already in place — this issue tracks the client-side rendering + selector-surface gap.

gorkem-bwl:
## Summary

Closes #3762. The \`customCSS\` field on status pages was persisted and validated end-to-end but never rendered on the public page, so operator-supplied styles had no effect.

## Root cause

Grep verified: no \`<style>\` tag, no \`customCSS\` reference, and no \`dangerouslySetInnerHTML\` anywhere under \`client/src/Pages/StatusPage/**\`. Data was flowing all the way to \`statusPage.customCSS\` on the client but nothing was rendering it.

## Fix

- In \`client/src/Pages/StatusPage/Status/index.tsx\`, on the public branch, render \`<style>{statusPage.customCSS}</style>\` inside the \`StatusPageThemeProvider\`.
- Reuse the existing \`cssReferencesExternalResource\` check from \`client/src/Utils/customCss.ts\` (uses \`@csstools/css-tokenizer\`) so CSS containing \`@import\`, \`url(http…)\`, \`url(//…)\`, or similar external references is dropped defensively on the client, matching the server-side check.
- The admin preview iframe already points at the public URL, so this single insertion covers both the live page and the in-app preview — no separate render path needed.

## Out of scope

Adding stable selectors (\`data-checkmate-*\` or non-hashed class names) so operators have something reliable to target. That is the second half of #3762 and will be a separate, larger PR.

## Test plan

- [ ] Save a status page with \`customCSS: "body { background: papayawhip; }"\` — public page background turns papayawhip.
- [ ] Save \`customCSS: "@import url(https://evil.example/x.css);"\` — CSS is dropped, no \`<style>\` tag rendered.
- [ ] Save \`customCSS: ""\` or leave unset — no \`<style>\` tag rendered.
- [ ] Admin preview iframe reflects the applied CSS (since it points at the public URL).
- [ ] TypeScript, ESLint, and Prettier all pass.

### DIFF
--- a/client/src/Pages/StatusPage/Status/index.tsx
+++ b/client/src/Pages/StatusPage/Status/index.tsx
@@ -21,6 +21,7 @@ import {
 	getStatusPagePublicUrl,
 	isCustomDomainHost,
 } from "@/Utils/statusPageUrl";
+import { cssReferencesExternalResource } from "@/Utils/customCss";
 import { HeaderStatusPageControls } from "@/Pages/StatusPage/Status/Components/HeaderStatusPageControls";
 import { StatusPageThemeProvider } from "@/Pages/StatusPage/Status/themes/StatusPageThemeProvider";
 import {
@@ -132,6 +133,10 @@ const StatusPageView = () => {
 	// Public route: render directly on the viewport, themed background covers everything.
 	if (isPublic) {
 		const themeConfig = THEME_CONFIGS[resolveStatusPageTheme(statusPage.theme)];
+		const customCss =
+			statusPage.customCSS && !cssReferencesExternalResource(statusPage.customCSS)
+				? statusPage.customCSS
+				: "";
 		return (
 			<StatusPageThemeProvider
 				theme={statusPage.theme}
@@ -139,6 +144,7 @@ const StatusPageView = () => {
 				timezone={statusPage.timezone}
 				paintBody
 			>
+				{customCss && <style>{customCss}</style>}
 				<BaseStatusPage
 					statusPage={statusPage}
 					monitors={monitors}


ajhollid (approved):
Looks good!

Issue #3765, opened 2026-07-10
Infrastructure network throughput is zero when monitor interval ≥ 1 aggregation bucket


gorkem-bwl:
### Summary

The Infrastructure > Network charts show zero throughput (and the "Capture is reporting zero traffic … container … loopback interfaces" notice) even when Capture is running on a real host and the interface counters are actively growing. This happens because the server computes per-second throughput as `(last check bytes_sent − first check bytes_sent) / (last − first)` **within a single aggregation bucket**, and buckets frequently contain only one sample.

### Details

**Aggregation:** `server/src/domain/checks/check.repository.mongo.ts` — `getHardwareStats` (~lines 657–798) buckets checks by `$dateToString` with a format string like `%Y-%m-%dT%H:%M:00Z` (one bucket per minute for the default `recent` range).

**Bucket format map:** `server/src/domain/monitors/monitor.service.ts:157–166`.

```
recent:  %Y-%m-%dT%H:%M:00Z   (per minute)
day:     %Y-%m-%dT%H:00:00Z   (per hour)
week:    %Y-%m-%dT00:00:00Z   (per day)
month:   %Y-%m-%dT00:00:00Z   (per day)
all:     %Y-%m-%dT00:00:00Z   (per day)
```

**Delta computation (per interface):**

```js
bytesSentPerSecond = {
  $let: {
    vars: {
      tDiff: { $divide: [{ $subtract: [{ $last: "$createdAts" }, { $first: "$createdAts" }] }, 1000] },
      f:     { $arrayElemAt: [{ $map: { input: { $first: "$net" }, as: "i", in: "$$i.bytes_sent" } }, "$$nIdx"] },
      l:     { $arrayElemAt: [{ $map: { input: { $last: "$net" },  as: "i", in: "$$i.bytes_sent" } }, "$$nIdx"] },
    },
    in: { $cond: [{ $gt: ["$$tDiff", 0] }, { $divide: [{ $subtract: ["$$l", "$$f"] }, "$$tDiff"] }, 0] },
  },
}
```

When a bucket contains **one** check, `$first === $last` → `l - f = 0` → `bytesSentPerSecond = 0`. Same for every `deltaX` field.

### Repro

- Monitor interval ≥ 60 s (this is the default).
- Range = "recent" (default, 2-hour view).
- Backend produces per-minute buckets; each bucket has one sample; every interface reports 0 throughput regardless of what the underlying `/proc/net/dev` counters show.

### Impact

[...]

PR #3766, opened 2026-07-10
fix(infrastructure): trigger network 'no traffic' notice on interface name only

gorkem-bwl:
## Summary

Related to #3765.

The Infrastructure > Network tab was firing a "Capture is reporting zero traffic on every interface … container … loopback interfaces" notice whenever the aggregated per-second delta was zero across every interface. That is a **false positive** on real hosts whenever the monitor interval is ≥ the aggregation bucket size (default: 60 s interval, 1-minute bucket). Single-sample buckets always produce zero per-second deltas even though the underlying interface has plenty of real traffic — see #3765 for the root cause.

## Fix

Base the notice on **interface names**, not on delta values.

- Only show the notice when the *only* interfaces Capture reports are loopback / container-bridge names (\`lo\`, \`lo0\`, \`docker0\`).
- Real NICs (\`enp0s6\`, \`eth0\`, \`wlan0\`, \`tunN\`, \`vethN\`, \`brN\`, …) → notice hidden, even if the current buckets read zero.
- Empty checks array → notice hidden (already gated by \`chartConfigs.length > 0\`).

The underlying zero-throughput bug (single-sample buckets) is tracked in #3765 as a separate, larger fix.

## Diff

- \`client/src/Pages/Infrastructure/Details/Components/ChartsNetwork.tsx\`
- +16 / -5

## Test plan

- [ ] Host running Capture directly, real NIC present → notice hidden.
- [ ] Container-only run reporting only \`lo\` → notice shown.
- [ ] Container with \`lo\` + \`docker0\` and no real NIC → notice shown.
- [ ] Empty checks / no data → notice hidden.
- [ ] TypeScript, ESLint, Prettier pass.

### DIFF
--- a/client/src/Pages/Infrastructure/Details/Components/ChartsNetwork.tsx
+++ b/client/src/Pages/Infrastructure/Details/Components/ChartsNetwork.tsx
@@ -92,14 +92,14 @@ export const InfraNetworkCharts = ({
 		() => getChartConfigs(theme, checks, t),
 		[theme, checks, t]
 	);
-	const showNoTrafficNotice =
+	const showLoopbackOnlyNotice =
 		chartConfigs.length > 0 && onlyContainerVisibleInterfaces(checks);
 
 	return (
 		<Stack gap={theme.spacing(LAYOUT.XS)}>
-			{showNoTrafficNotice && (
+			{showLoopbackOnlyNotice && (
 				<NoticeBanner severity="warning">
-					{t("pages.infrastructure.charts.labels.netNoTraffic")}
+					{t("pages.infrastructure.charts.labels.loopbackOnlyNotice")}
 				</NoticeBanner>
 			)}
 			<Grid
--- a/client/src/Pages/Infrastructure/Details/Components/ChartsNetwork.tsx
+++ b/client/src/Pages/Infrastructure/Details/Components/ChartsNetwork.tsx
@@ -63,10 +63,20 @@ const getChartConfigs = (
 	return configs;
 };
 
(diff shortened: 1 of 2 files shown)


ajhollid (changes requested):
The purpose of this PR is to change the behaviour so the trigger for displaying or not displaying is now network adapter name, whereas it used to be network traffic.

This is fine, but the UI still shows:

```
<NoticeBanner severity="warning">
	{t("pages.infrastructure.charts.labels.netNoTraffic")}
</NoticeBanner>
```

Which is a message showing "Capture is reporting zero traffic on every interface..."

The var `showNoTrafficNotice` is also misleading now, as it has nothing to do with traffic anymore.

TLDR, var should be renamed, string key in the translation file should be renamed to something like `loopbackOnlyNotice`, and the string content should be updated to match.

gorkem-bwl:
Good points, all three addressed in the latest commit:

- Variable: `showNoTrafficNotice` → `showLoopbackOnlyNotice`
- i18n key: `pages.infrastructure.charts.labels.netNoTraffic` → `pages.infrastructure.charts.labels.loopbackOnlyNotice`
- English copy rewritten to describe the loopback-only condition rather than "zero traffic on every interface"

The 16 non-English locales still hold the old `netNoTraffic` copy under the old key. Since it was a real translated string in each locale, I did not machine-translate them here — instead opened #3770 to route them through the usual PoEditor pipeline. Until that lands, non-English users will see the English string as the fallback, which is a step forward from seeing a translated-but-inaccurate one.

ajhollid (approved):
Awesome, thanks for the changes! 🚀

Issue #3768, opened 2026-07-10
Add Italian language support


edobusy:
**Is your feature request related to a problem? Please describe.**
I noticed Italian isn't among the supported languages, and I thought you might be interested in having a curated translation for both the README and the UI.
I'm a native Italian speaker, so I am happy to work on both.

**Describe the solution you'd like**
Add Italian across both surfaces:
1. `docs/translations/README.it.md` + the 🇮🇹 link in the README language bar.
2. `client/src/locales/it.json` + Italian in the language selector.

I've prepared the README translation and will open a PR for it shortly. The
app-locale part can follow.

**Describe alternatives you've considered**
For the app strings: I understand the team generates translations with an LLM.
I can either hand-translate `it.json` or simply review and polish the
generated Italian so the tone/UI phrasing reads naturally, whichever the team
prefers.

**Additional context**
Glad to maintain the Italian strings going forward. README PR incoming and will link here.

edobusy:
One small thing I noticed while translating: the K8s install link is broken in all the translated READMEs. They use `[K8s](./charts/helm/checkmate/INSTALLATION.md)`, but the translations live in `docs/translations/`, meaning that the path resolves to a folder that doesn't exist. It should be `../../charts/helm/checkmate/INSTALLATION.md`. 

The root README's `./charts/...` is correct because it sits at the repo root. 

It affects all 16 (+1 if we count my proposed translation) versions, happy to fix it in a separate PR if useful.

Glazzze:
Hi @edobusy,

I'd like to work on this issue.

I have checked the required changes and it looks like the main tasks are:
- adding the Italian README translation under `docs/translations/README.it.md`
- adding the Italian locale file under `client/src/locales/it.json`
- updating the language selector/configuration if needed

I've already contributed several small fixes to open-source projects and would like to help add this language support.

I'll start by preparing the translation and checking the existing i18n structure. If everything looks good, I'll submit a PR.

Thanks!

edobusy:
Hi @Glazzze, we already implemented it, I am just keeping this open until I finish reviewing the translation Gorkem implemented. 

Thank you for offering though!

gorkem-bwl:
All PRs linking to this issue are merged, so closing this issue.

PR #3778, opened 2026-07-14
fix(docs): correct K8s install link path in translated READMEs

edobusy:
## Summary

Fixes a broken **K8s installation** link in all 17 translated READMEs under `docs/translations/`.

## The bug

The translated READMEs link to the Kubernetes install guide with a path copied verbatim from the root `README.md`:

    [K8s](./charts/helm/checkmate/INSTALLATION.md)

Because these files live in `docs/translations/` (two levels below the repo root), `./charts/...` resolves to `docs/translations/charts/helm/checkmate/INSTALLATION.md`, which doesn't exist, so the link 404s in every translation.

The root `README.md` uses the same `./charts/...` and is correct, because it sits at the repo root.

## The fix

Point the path up two levels in the 17 translated files:

    [K8s](../../charts/helm/checkmate/INSTALLATION.md)

This resolves to `charts/helm/checkmate/INSTALLATION.md` at the repo root (verified present).

## Scope / safety

- Only the K8s link is touched, one line per file, 17 files.
- The other two relative links in these files (`../custom-ca-trust.md` and `../`) were already correct and are left untouched.
- The root `README.md` is intentionally not changed.

Noticed while working on #3768.


ajhollid (approved):
Thanks for taking care of this!

PR #3797, opened 2026-07-22
Fix monitoring status stuck

### LINKED ISSUE #3772: UI Bug and endless initializing
**Describe the bug**
The connection was successful, and the log shows that everything is fine. However, the lifecycle status remains "initializing."

And in German, the text sticks out of the box for RAM usage under "Infrastructure" (see screenshot graph)

**To Reproduce**
1. Update to v3.9.2 (just rise the versions in docker compose)

```yaml
services:
  checkmate:
    image: 'ghcr.io/bluewave-labs/checkmate-backend-mono-multiarch:v3.9.2'
    environment:
      - SERVICE_URL_CHECKMATE_52345
      - 'UPTIME_APP_API_BASE_URL=${SERVICE_URL_CHECKMATE}/api/v1'
      - 'UPTIME_APP_CLIENT_HOST=${SERVICE_URL_CHECKMATE}'
      - 'DB_CONNECTION_STRING=mongodb://mongodb:27017/uptime_db'
      - 'CLIENT_HOST=${SERVICE_URL_CHECKMATE}'
      - 'JWT_SECRET=${SERVICE_PASSWORD_64_JWT}'
    depends_on:
      - mongodb
    healthcheck:
      test:
        - CMD
        - node
        - '-e'
        - "require('http').get('http://127.0.0.1:52345/health', r => process.exit(r.statusCode === 200 ? 0 : 1)).on('error', () => process.exit(1))"
      interval: 30s
      timeout: 5s
      retries: 3
  mongodb:
    image: 'ghcr.io/bluewave-labs/checkmate-mongo:v3.9.2'
    command:
      - mongod
      - '--quiet'
      - '--bind_ip_all'
    volumes:
      - 'checkmate-mongo:/data/db'
    healthcheck:
      test:
        - CMD
        - mongosh
        - '--eval'
        - "db.adminCommand('ping')"
        - '--quiet'
      interval: 5s
      timeout: 30s
      start_period: 10s
```

**Expected behavior**
lifecycle is working

**Screenshots**

Image

 - Browser Chrome
 - Version 150.0.7871.46 (Offizieller Build) (64-Bit)

LvckyAPI:
## Describe your changes                                                                                                                                                                                                         
                                                                                                                                                                                                                                   
  **Bug fix: monitors permanently stuck in "Initializing" after upgrading to v3.9.x**                                                                                                                                              
                                                                                                                                                                                                                                   
  Monitors carrying a legacy status value (e.g. the string `"true"` from older versions that stored `monitor.status` as a boolean) can never recover: `computeReachability` only transitions `down → up`, so any value outside the 
  `MonitorStatuses` enum is written back unchanged on every evaluation. Checks, `lastEvaluatedAt` and the status window all keep working — only the status stays frozen, and the UI renders the unknown value with the             
  "Initializing" fallback label forever (while the summary widgets, which aggregate on the real `initializing` value, count 0).                                                                                                    
                                                                                                                                                                                                                                   
  Older versions masked this because the warmup branch used to hard-overwrite the status on every check; that overwrite was intentionally removed in v3.9.x, which surfaced the stuck state.                                       
                                                                                                                                                                                                                                   
  Changes:                                                                                                                                                                                                                         
                                                                                                                                                                                                                                   
  - **`statusService.updateMonitorStatus`**: status values outside the `MonitorStatuses` enum are now resolved the same way as `initializing`/`maintenance`, so affected monitors self-heal on the next evaluated check. No        
  notification storm: `statusChanged` only surfaces when the resolved status is `down` (so a real outage still opens an incident).                                                                                                 
  - **Regression tests**: legacy `"true"` resolves to `up` without a status-change notification; legacy `"false"` resolves to `down` and surfaces `statusChanged` so an incident is opened.                                        
  - **UI**: infrastructure gauge titles no longer overflow their card (`overflowWrap` on the `BaseChart` title) and the title color now references the theme path (`theme.palette.text.secondary`). Shortened/fixed German gauge   
  and chart labels (typo "Termperatur", clarified temperature chart label to "CPU temperature") so they fit the cards.                                                                                                             
                                                                                                                                                                                                                                   
  Verified against a live instance: monitors stuck on `"true"` since the v3.9.2 upgrade, evaluate jobs running with zero failures — pausing/resuming (which resets the status to `initializing`) was the only workaround. With this
  fix the status resolves automatically.                                                                                                                                                                                           
                                                                                                                                                                                                                                   
  Fixes #3772                                                                                                                                                                                                                          
                                                                                                                                                                                                                                   
  ```Javascript                                                                                                                                                                                                                    
  <div>{t('add')}</div>                                   
  ``` 

- [x] I have not included any files that are not related to my pull request, including package-lock and package-json if dependencies have not changed                                                                            
  - [x] I ran npm run format in server and client directories, which automatically formats your code.                                                                                                                              

### DIFF
--- a/server/src/service/statusService.ts
+++ b/server/src/service/statusService.ts
@@ -220,11 +220,11 @@ export class StatusService implements IStatusService {
 			// Resolve "initializing" and "maintenance" up front, incidents should be created on initialization && down
 			// Unknown values (e.g. legacy boolean statuses from old versions) are resolved the same way,
 			// otherwise computeReachability can never transition them and they stay stuck forever.
-			const isKnownStatus = (MonitorStatuses as readonly string[]).includes(monitor.status);
+			  const isUnknownStatus = !MonitorStatuses.includes(monitor.status);
 			let newStatus: MonitorStatus = monitor.status;
 			let statusChanged = false;
-			if (monitor.status === "initializing" || monitor.status === "maintenance" || !isKnownStatus) {
-				newStatus = status === true ? "up" : "down";
+			if (monitor.status === "initializing" || monitor.status === "maintenance" || isUnknownStatus) {
+				newStatus = status ? "up" : "down";
 				patch.status = newStatus;
 				statusChanged = newStatus === "down";
 			}
--- a/server/src/service/statusService.ts
+++ b/server/src/service/statusService.ts
@@ -220,7 +220,7 @@ export class StatusService implements IStatusService {
 			// Resolve "initializing" and "maintenance" up front, incidents should be created on initialization && down
 			// Unknown values (e.g. legacy boolean statuses from old versions) are resolved the same way,
 			// otherwise computeReachability can never transition them and they stay stuck forever.
-			  const isUnknownStatus = !MonitorStatuses.includes(monitor.status);
+			const isUnknownStatus = !MonitorStatuses.includes(monitor.status);
 			let newStatus: MonitorStatus = monitor.status;
 			let statusChanged = false;
 			if (monitor.status === "initializing" || monitor.status === "maintenance" || isUnknownStatus) {
--- a/server/src/service/statusService.ts
+++ b/server/src/service/statusService.ts
@@ -1,7 +1,7 @@
 import { IMonitorStatsRepository } from "@/domain/monitor-stats/monitor-stats.repository.interface.js";
 import { IMonitorsRepository } from "@/domain/monitors/monitor.repository.interface.js";
 import type { Check, CheckDiskInfo, CheckSnapshot } from "@/domain/checks/check.type.js";
-import type { Monitor, MonitorStatus } from "@/domain/monitors/monitor.type.js";
+import { MonitorStatuses, type Monitor, type MonitorStatus } from "@/domain/monitors/monitor.type.js";
 import type {
 	DockerStatusPayload,
 	GameStatusPayload,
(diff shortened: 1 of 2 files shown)


client/src/Components/design-elements/BaseChart.tsx
```diff
@@ -39,7 +39,8 @@ export const BaseChart = ({
 			>
 				<Typography
 					variant="eyebrow"
-					color="text.secondary"
+					color={theme.palette.text.secondary}
+					sx={{ overflowWrap: "anywhere" }}
```
ajhollid, line 43: This is unrelated to the monitor status PR.  Let's remove this and open a new PR if it's an issue to resolve.

server/src/service/statusService.ts
```diff
@@ -222,9 +222,12 @@ export class StatusService implements IStatusService {
 			const patch: Partial<Monitor> = {};
 
 			// Resolve "initializing" and "maintenance" up front, incidents should be created on initialization && down
+			// Unknown values (e.g. legacy boolean statuses from old versions) are resolved the same way,
+			// otherwise computeReachability can never transition them and they stay stuck forever.
+			const isKnownStatus = (MonitorStatuses as readonly string[]).includes(monitor.status);
```
ajhollid: This cast is unnecessary and can be removed.

I also think it makes more sense semantically to invert this check as well, so that we have `isUnknownStatus` instead of `!isKnownStatus`.

```
  const isUnknownStatus = !MonitorStatuses.includes(monitor.status);
  if (monitor.status === "initializing" || monitor.status === "maintenance" || isUnknownStatus) {
     ...
  }
```
I think that matches the intent of the fix a bit better and is a bit easier to read.

ajhollid (changes requested):
The fix here is solid, but there is an unnecessary type cast, not sure why it was included, but it should be removed.

The PR description is also not correct, it claims a German locale fix, but I don't see that anywhere here.

There's also an unrelated chart fix included in the PR, let's remove that.

Other than that looks good to me, thanks for the fix!

LvckyAPI:
Sorry i build it on the my ui bug fix branch. I will fix it later. Thanks for the quick review

ajhollid:
> Sorry i build it on the my ui bug fix branch. I will fix it later. Thanks for the quick review

Yeah I saw your other PR just after I left the review here :P  No problem, just leave it in since the other PR is merged anyway.  Once the other fixes are in we're good to go :+1:

LvckyAPI:
> Once the other fixes are in we're good to go

Should be done

ajhollid (approved):
Thanks for making changes, looks good!

PR #3785, opened 2026-07-17
feat: Add bearer/basic auth option to webhooks

smz202000:
## Description

Closes #2369

This PR adds optional authentication support for webhook notifications. Users can now select an authentication type when configuring a webhook:

- **None** (default) - No authentication
- **Basic Auth** - Provides username and password fields, sends Authorization: Basic <base64> header
- **Bearer Token** - Provides a token field, sends Authorization: Bearer <token> header

### Backend
- Added webhookAuthType, webhookUsername, webhookPassword, webhookToken fields to the Notification interface
- Added corresponding Mongoose schema fields
- Added optional validation fields for the webhook discriminated union
- Implemented buildAuthHeaders method in WebhookProvider to send the appropriate Authorization header based on auth type

### Frontend
- Added webhook auth UI section (auth type dropdown + conditional fields) to the notification create/edit page
- Updated Notification type, Zod validation schema, form hook defaults, and English translations

## Testing
- The auth fields are optional; existing webhook configurations continue to work without changes
- Tested in both sendMessage and sendTestAlert code paths

PR #3803, opened 2026-07-27
fix: stop returning notification credentials to the client

### LINKED ISSUE #3771: Notification secrets (accessToken) are returned to the client in API responses
**Describe the bug**

Notification credentials are returned to the client in API responses. When a notification is created, fetched, or edited, the response includes the stored `accessToken` (the provider secret) in plaintext for **matrix, telegram, twilio, and pushover**. So a stored secret round-trips to the browser on every fetch, and the edit form even relies on reading it back to repopulate the token field. Secrets should be write-only: stored and used to send, but never returned.

**To Reproduce**
1. Create a notification with a secret (e.g. a **telegram** channel with a bot token), via the UI or `POST /api/v1/notifications`.
2. Fetch it: `GET /api/v1/notifications/{id}` (or `GET /api/v1/notifications/team`).
3. In the response JSON, `data.accessToken` contains the bot token in plaintext.
4. Same in the browser: opening the notification's **edit** form triggers that GET and loads the secret into the client.

**Expected behavior**

The API should never return credential fields (`accessToken`, and any future secret) in responses. They should be store-and-use only: used when sending notifications, but never sent back to the client.

**Screenshots**

Image

Not applicable. This is server-side API behavior, reproducible with any client (curl or Swagger `/api-docs`), and also observable in-browser via the edit form's GET response.

**Additional context**
- **Severity:** authenticated and team-scoped (only members who manage the notification receive it), so the practical risk is client-side exposure such as XSS, the Network tab, or logs. It is a defense-in-depth hardening, not a remote exploit.
- **Context from #3717:** while discussing ntfy auth, @ajhollid asked that provider secrets not be returned to the client from the API. @mospell then noted that this already happens for the existing providers (the Matrix access token is returned alongside the other notification settings) and suggested filing it as its own issue. This is that issue.
- **The fix is coordinated fullstack:** the edit form currently depends on reading the token back to repopulate, so redacting the API also needs a frontend change (don't prefill the secret; a "re-enter to change / leave to keep" pattern). 

Happy to help with this!

P.S. I saw a similar credential-exposure issue was recently handled as a security advisory; if you'd prefer this to go through that process too, happy to close this and re-report it privately.

edobusy:
## Describe your changes

`POST /notifications`, `GET /notifications/team`, `GET /notifications/{id}` and
`PATCH /notifications/{id}` all returned the stored `accessToken` as plain text. That affects
matrix, telegram, twilio and pushover. Any logged in team member could read every channel
credential from the API. #3771 has screenshots of the returned fields on current `develop`.

The Configure form depended on this. It read the token back to prefill its field, so the leak
could not be fixed on the server alone.

This continues #3559 by @egeoztass, which has been idle since early May. It takes the approach
@ajhollid asked for there, rebases it onto the backend changes that have landed since, and adds
the tests that were still missing.

cc @ajhollid

### Credentials are no longer returned

Responses are built from a list of public fields, and each credential is reported as
`accessTokenSet: boolean` instead. This is the same approach the settings page already uses for
`pagespeedKeySet` and `emailPasswordSet`.

Step_1
(the same request on this branch: accessTokenSet, and no token)

### The form no longer prefills

When you edit a channel, the credential field is replaced by "credential is set" and a Reset
button. The input only appears after you press Reset. If you do not replace the credential, it is
left out of the request, and the stored one is kept. The four duplicated token fields are now one
`CredentialField` component, and credential inputs are masked, which they were not before.

Step_1 2
(editing a channel: the credential is not prefilled)

Step_4
(after pressing Reset: a masked input for the new value)

### Edit validation is derived, not duplicated

Each channel schema is now declared once. `editNotificationBodyValidation` is built from the
create schemas with `.partial()`, as asked for in #3559. The behaviour follows from that: leave
the credential out to keep it, send a value to replace it, and an empty string is rejected so a
required credential cannot be wiped by mistake.

Step_3 2
(editing another field while leaving the credential out: the stored credential is unchanged)

Step_5
(Reset, then Save with an empty field: rejected before the request is sent)

### Test still works on a saved channel

Hiding the credential would otherwise break the Test button, because it sends the form's current
values. `POST /notifications/{id}/test` fills in any credential the client left out from the
stored record.

### The part worth reviewing closely

A credential that the server fills in can only be sent to the destination it is already stored
against. If the request changes where the notification is delivered and leaves the credential
out, it is rejected with a 400. Otherwise the endpoint could be used to read a stored credential
once: send `{id, homeserverUrl: attacker.example}` with the token left out, and the server would
deliver it there. `NotificationDestinationFields` lists, per channel, which fields decide the
outbound host. Providers that post to a fixed host have no such field, and are listed with an
empty array and a comment.

Step_6_guardrail
(changing the Matrix home server and pressing Test, with the credential left out: rejected)

To be clear about what this does and does not buy: someone who can already PATCH the channel can
point it anywhere and wait for an alert. The difference is that a PATCH is saved, visible in the
channel's configuration, and stops alerts from reaching the team. A silent test is none of those
things. I did not want the fix to hand back a quieter version of the problem it removes.

### Safe by default

The list of public fields is written out by hand rather than derived as "everything that is not a
credential". A field added to `Notification` later is not returned until someone classifies it.
Leaving a field unclassified fails the build, and the error names the field. This follows the fix
for the status page leak (GHSA-3m74-8cg9-rp8j) and the drift check already in
`openapi/routes/notification.ts`.

This matters now, because #3785 adds `webhookToken` and `webhookPassword`. If the code simply
stripped the credentials it knew about, those two would be returned the day that PR merges. With
this approach they cannot be, and the build asks which they are.

### Two decisions I would like your call on

1. **`accountSid` is still returned.** #3559 hid it. It is Twilio's Basic auth username and it
   travels in the request URL by design, so it identifies an account rather than authenticating
   one. Hiding it behind a Reset button costs the user something and gains nothing. Happy to
   change it, it is one line plus its flag.
2. **The redaction lives in the controller**, next to `settingsController.buildAppSettings`. The
   response shape is an HTTP concern, and the service has to keep returning full records for the
   send path. The alternative is the service layer. I followed the existing precedent.

### One question for later: where the credential sits in the form

`CredentialField` renders in the same place the old token input was, so its position depends on
the provider. It is first for telegram and pushover, second of four for twilio (between the
Account SID and the phone numbers), and last for matrix. I left the order alone on purpose,
because it is the order the fields were already in. Reordering fields is a UX change with no
security value, and Account SID and Auth Token sit next to each other in the Twilio console, so
splitting them would make the create flow worse.

It is worth deciding before #3785 lands, because that PR adds two credentials to a single
channel. Two credentials are easier to read as one block: a single "credentials are set" state
with one Reset that shows both inputs, rather than two separate Reset buttons that let a pair be
half replaced without anyone noticing. A block that size does look odd in the middle of a
provider's fields.

So the rule I would suggest, if you want one:

> A single credential stays where its input was. A group of credentials is shown as one block at
> the end of its provider's fields.

Happy to do that here, in a follow up, or not at all. Just say which. Nothing in the backend
depends on it, because leaving a field out keeps it per field, so grouping is only a UI
convention.

### Notes

- The send path is untouched. Providers still read full records straight from the repository, and
  there are tests covering that.
- **API change:** `accessToken` is gone from notification responses, and `accessTokenSet` is new.
  The OpenAPI spec now documents a real response schema for these endpoints instead of an untyped
  payload, and a test parses a real response against that schema strictly, so the two cannot
  drift apart.
- `openapi.json` is not regenerated here. Regenerating it on current `develop` already produces a
  diff of about 25k lines that has nothing to do with this change, so I left it to the build.
- API clients that send the full body on edit, credential included, keep working.

### Out of scope

- Switching a saved channel to a different provider type leaves the old provider's credential in
  the document. That is not new. Today the form prefills the old token into the new provider's
  field and saves it, so this PR does not change the behaviour.
- `usePost` logs the request body when a request fails, which includes a credential on a failed
  create or test. That is separate and already there. Happy to open an issue.
- Encryption at rest, and forcing HTTPS.

### Checks

`npm run lint`, `npm run format-check` and `tsc` are clean in both packages. Server suite: 1354
tests passing, 27 of them new. Client builds.

Fixes #3771

```Javascript
<div>{t('add')}</div>
```

**P.S. Corrections after rebasing onto `develop`** (the new commits are explained in the
comment below)

Some of the above is now out of date. Rather than rewrite the description and lose the
screenshots, here are the corrections:

- **"The send path is untouched" is no longer true.** One provider file changed: `roomId` is
  now URL-encoded in the Matrix provider, because it was going into the request path raw.
- **`openapi.json` is now regenerated.** My note that regenerating produced a ~25k line diff
  was wrong. That was tabs versus spaces, not content.
- **The first "Out of scope" bullet no longer applies.** Switching a channel to a different
  provider type no longer leaves the old provider's credential in the document. It could not
  stay out of scope once the credential became invisible, since the old behaviour relied on
  the user seeing it prefilled into the new provider's field.
- **The guard now covers saving, not just testing.** The paragraph about "someone who can
  already PATCH the channel can point it anywhere and wait for an alert" is superseded:
  `PATCH` is guarded too, because otherwise the guard on the test endpoint was bypassable in
  two requests.
- **"Destination" means host, not recipient.** A fixed-host channel's credential can still be
  repointed at another chat id, user key or phone number without being re-entered. That
  discloses later alerts to whoever owns the new recipient, but never the credential itself.
  The comment in the code says so now.
- **Checks:** 1401 tests across 80 suites, up from 1339 across 79 on `develop`, so 62 new
  rather than 27. Server `tsc`, lint and format are clean. On the client, `tsc -b` builds and
  lint and format are clean for the files this PR touches, though `npm run lint` reports 12
  errors that are already on `develop`, in files this PR does not open.
- **#3559 is now closed** by @egeoztass in favour of this one, so it is no longer idle.
- **The screenshots predate #3829's form refactor.** The form's layout changed with it; the
  flow they show did not.
- Minor: the request-body logging note applies to `usePatch` and `usePut` as well as
  `usePost`.

### DIFF
--- a/client/src/Hooks/useNotificationForm.ts
+++ b/client/src/Hooks/useNotificationForm.ts
@@ -1,10 +1,14 @@
 import { useMemo } from "react";
-import { notificationSchema } from "@/Validation/notifications";
+import { zodResolver } from "@hookform/resolvers/zod";
+import type { Resolver } from "react-hook-form";
+import { notificationEditSchema, notificationSchema } from "@/Validation/notifications";
 import type { NotificationFormData } from "@/Validation/notifications";
 import type { Notification } from "@/Types/Notification";
 
 interface UseNotificationFormOptions {
 	data?: Notification | null;
+	/** True once the user has asked to replace the stored credential, which reveals its field. */
+	isCredentialReset?: boolean;
 }
 
 function buildDefaults(data: Notification | null): NotificationFormData {
@@ -12,7 +16,9 @@ function buildDefaults(data: Notification | null): NotificationFormData {
 		type: data?.type ?? "email",
 		notificationName: data?.notificationName || "",
 		address: data?.address || "",
-		accessToken: data?.accessToken || "",
+		// Never seeded from the API: credentials are not returned, and an unchanged one is kept by
+		// leaving it out of the payload.
+		accessToken: "",
 		accountSid: data?.accountSid || "",
 		phone: data?.phone || "",
 		twilioPhoneNumber: data?.twilioPhoneNumber || "",
@@ -22,9 +28,30 @@ function buildDefaults(data: Notification | null): NotificationFormData {
 	};
 }
 
-export const useNotificationForm = ({ data = null }: UseNotificationFormOptions = {}) => {
-	return useMemo(() => {
-		const defaults = buildDefaults(data);
-		return { schema: notificationSchema, defaults };
-	}, [data]);
+export const useNotificationForm = ({
+	data = null,
+	isCredentialReset = false,
+}: UseNotificationFormOptions = {}) => {
+	// Only the fetched notification may rebuild the defaults. Asking to replace a credential swaps
+	// the schema but must not hand the form a new defaults object, which would reset it and discard
+	// whatever the user has already typed.
+	const defaults = useMemo(() => buildDefaults(data), [data]);
+
+	/**
+	 * Whether the stored credential is being kept, meaning its field stays hidden and it is left out
+	 * of the payload. It is kept only while the user has not asked to replace it *and* the selected
+	 * channel type still matches the one it was stored against: a credential belongs to the provider
+	 * it was issued by, so switching type has to ask for a new one rather than carry the old one over.
+	 */
+	const isCredentialKept = (type: NotificationFormData["type"]) =>
+		Boolean(data?.accessTokenSet) && !isCredentialReset && type === data?.type;
+
+	// The schema depends on the type currently selected in the form, so it is resolved per validation
+	// run rather than fixed at render time.
+	const resolver: Resolver<NotificationFormData> = (values, context, options) =>
+		zodResolver(
+			isCredentialKept(values.type) ? notificationEditSchema : notificationSchema
+		)(values, context, options);
+
+	return { resolver, defaults, isCredentialKept };
 };
--- a/client/src/Pages/Notifications/create/CredentialField.tsx
+++ b/client/src/Pages/Notifications/create/CredentialField.tsx
@@ -1,0 +1,51 @@
+import Box from "@mui/material/Box";
+import { Button, FieldLabel } from "@/Components/inputs";
+import { FormTextField } from "@/Components/inputs/forms/FormTextField";
+
+interface CredentialFieldProps {
+	fieldLabel: string;
+	placeholder: string;
+	storedLabel: string;
+	resetLabel: string;
+	isEditable: boolean;
+	onReset: () => void;
+}
+
+/**
+ * Input for a stored credential. The API never returns the credential itself, so an existing one
+ * cannot be prefilled: the field is replaced by a Reset button until the user chooses to change it,
+ * the same flow the PageSpeed API key uses on the settings page.
+ */
+export const CredentialField = ({
+	fieldLabel,
+	placeholder,
+	storedLabel,
+	resetLabel,
+	isEditable,
+	onReset,
+}: CredentialFieldProps) => {
+	if (!isEditable) {
+		return (
+			<Box>
+				<FieldLabel>{storedLabel}</FieldLabel>
+				<Button
+					onClick={onReset}
+					variant="contained"
+					color="error"
+				>
+					{resetLabel}
+				</Button>
+			</Box>
+		);
+	}
+
+	return (
+		<FormTextField
+			name="accessToken"
+			type="password"
+			autoComplete="new-password"
+			fieldLabel={fieldLabel}
+			placeholder={placeholder}
+		/>
+	);
+};
--- a/client/src/Pages/Notifications/create/index.tsx
+++ b/client/src/Pages/Notifications/create/index.tsx
@@ -3,7 +3,7 @@ import { Button } from "@/Components/inputs";
 import Stack from "@mui/material/Stack";
 import { useTheme } from "@mui/material/styles";
 
-import { useEffect, useMemo, useState } from "react";
+import { useEffect, useMemo, useRef, useState } from "react";
 import { useParams } from "react-router-dom";
 import { useNavigate } from "react-router-dom";
 import { FormProvider, useForm } from "react-hook-form";
@@ -44,17 +44,27 @@ const NotificationsCreatePage = () => {
 		defaultValues: defaults,
 	});
 
-	const { watch, reset, handleSubmit, clearErrors, trigger, getValues } = form;
+	const { watch, reset, handleSubmit, clearErrors, trigger, getValues, setValue } = form;
 
 	useEffect(() => {
 		reset(defaults);
 	}, [defaults, reset]);
 
 	const watchedType = watch("type");
+	const previousType = useRef(watchedType);
 
 	useEffect(() => {
 		clearErrors();
-	}, [watchedType, clearErrors]);
+
+		if (previousType.current === watchedType) return;
+		previousType.current = watchedType;
+
+		// A credential belongs to the provider it was issued by. The field is shared across providers,
+		// and react-hook-form keeps a value when its input unmounts, so without this the credential
+		// typed for one provider silently becomes the next one's, hidden behind a masked field.
+		setValue("accessToken", "");
+		setCredentialHasBeenReset(false);
+	}, [watchedType, clearErrors, setValue]);
 
 	const addressConfig = useMemo(() => {
 		if (watchedType === "pager_duty") {
--- a/client/src/Pages/Notifications/create/index.tsx
+++ b/client/src/Pages/Notifications/create/index.tsx
@@ -3,11 +3,10 @@ import { Button } from "@/Components/inputs";
 import Stack from "@mui/material/Stack";
 import { useTheme } from "@mui/material/styles";
 
-import { useEffect, useMemo } from "react";
+import { useEffect, useMemo, useState } from "react";
 import { useParams } from "react-router-dom";
 import { useNavigate } from "react-router-dom";
 import { FormProvider, useForm } from "react-hook-form";
-import { zodResolver } from "@hookform/resolvers/zod";
 import { useGet, usePost, usePatch } from "@/Hooks/UseApi";
 import { useNotificationForm } from "@/Hooks/useNotificationForm";
 import type { NotificationFormData } from "@/Validation/notifications";
@@ -16,6 +15,7 @@ import { useTranslation } from "react-i18next";
 import { NotificationChannels } from "@/Types/Notification";
 import { FormTextField } from "@/Components/inputs/forms/FormTextField";
 import { FormSelectField } from "@/Components/inputs/forms/FormSelectField";
+import { CredentialField } from "./CredentialField";
 
 const NotificationsCreatePage = () => {
 	const { t } = useTranslation();
@@ -32,10 +32,15 @@ const NotificationsCreatePage = () => {
 	const { patch, loading: isPatching } = usePatch<NotificationFormData, Notification>();
 	const { post: testPost, loading: isTesting } = usePost<NotificationFormData, void>();
 
-	const { schema, defaults } = useNotificationForm({ data: existingNotification });
+	const [credentialHasBeenReset, setCredentialHasBeenReset] = useState(false);
+
+	const { resolver, defaults, isCredentialKept } = useNotificationForm({
+		data: existingNotification,
+		isCredentialReset: credentialHasBeenReset,
+	});
 
 	const form = useForm<NotificationFormData>({
-		resolver: zodResolver(schema),
+		resolver,
 		defaultValues: defaults,
 	});
 
@@ -76,10 +81,22 @@ const NotificationsCreatePage = () => {
 		};
 	}, [watchedType, t]);
 
+	// A credential the user has not chosen to replace is left out of the payload entirely, which the
+	// server reads as "keep the stored value".
+	const buildPayload = (data: NotificationFormData): NotificationFormData => {
+		if (!isCredentialKept(data.type)) {
+			return data;
+		}
+		const payload = { ...data };
+		Reflect.deleteProperty(payload, "accessToken");
+		return payload;
+	};
+
 	const onSubmit = async (data: NotificationFormData) => {
+		const payload = buildPayload(data);
 		const result = isEditMode
-			? await patch(`/notifications/${notificationId}`, data)
-			: await post("/notifications", data);
+			? await patch(`/notifications/${notificationId}`, payload)
+			: await post("/notifications", payload);
 		if (result) {
 			navigate("/notifications");
 		}
@@ -88,8 +105,19 @@ const NotificationsCreatePage = () => {
 	const handleTest = async () => {
 		const isValid = await trigger();
 		if (!isValid) return;
-		const data = getValues();
-		await testPost("/notifications/test", data);
+		const payload = buildPayload(getValues());
+		// Testing a saved channel goes through its own endpoint so the server can supply the stored
+		// credential. Anything else — an unsaved channel, or one being switched to another provider,
+		// whose stored credential no longer applies — carries what it needs in the body.
+		const isSavedChannel = isEditMode && watchedType === existingNotification?.type;
+		const endpoint = isSavedChannel
+			? `/notifications/${notificationId}/test`
+			: "/notifications/test";
+		await testPost(endpoint, payload);
+	};
+
+	const handleCredentialReset = () => {
+		setCredentialHasBeenReset(true);
 	};
 
 	return (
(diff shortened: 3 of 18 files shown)


YoinkingFishy (approved):
Very well-crafted security fix. Key strengths:

1. **`NotificationDestinationFields` + `destinationChanged` guard** in `sendTestNotificationById` is a nuanced mitigation against the silent-test exfiltration vector — clearly thought through.
2. **Type-level drift detection** (`NotificationFieldsAreClassified`) elegantly prevents future fields from being accidentally exposed.
3. **Deriving edit schema from create via `.partial()`** avoids duplicated validation logic.

Minor suggestion: the `destinationChanged` rejection path in `sendTestNotificationById` doesn't have a dedicated unit test on the service layer. Worth adding one to pin the behaviour, since it's a security-sensitive guard.

ashu130698:
Verified this manually — deployed locally, created a Matrix channel, then tried changing the Home server to an attacker-controlled URL and hitting Test without re-entering the credential. Got blocked exactly as expected ("Enter the credentials again to test this channel with a different destination"). Also confirmed via a direct DB query that editing unrelated fields (e.g. chat ID) leaves the stored credential untouched. Agree this guard needs a dedicated unit test — happy to put one up, @edobusy / @ajhollid, let me know if that overlaps with anything in progress.

edobusy:
Rebased onto current `develop`: merges cleanly now, all five checks green. Five new commits
on top of the original four, and I have updated the description where it had gone stale.

**The rebase** had two things to deal with: #3829's form refactor, which rewrote both client
files this touches (re-applied on top of the new `FormProvider` idiom rather than merged, so
it does not quietly revert that work), and the new Rocket.Chat channel, which the
classification in this PR refused to compile without. Good to see that part earn its keep
rather than just be asserted.

**The new commits, in short:**

- The destination guard now covers `PATCH`, not just the test endpoint. It was bypassable in
  two requests: save the new destination first so stored and incoming agree, then test. It
  also now only guards credentials a channel can actually use, and a type change drops the
  stored credential, or asks for the new provider's if that provider needs one, rather than
  carrying the old one across.
- `roomId` is URL-encoded in the Matrix provider. It went into the request path raw, so
  escaped slashes could move the authenticated PUT elsewhere on the caller's own homeserver.
  Predates this PR, but the new test endpoint is what makes it reachable without knowing the
  token. Happy to split those two lines out if you would rather review them separately.
- The client clears the credential field on a type change. It is one shared field and
  react-hook-form keeps unmounted values, so a Telegram token could end up submitted as the
  Matrix one. Masking the input, which this PR adds, is what made that silent.
- Test hardening: the new endpoint had no controller test, and the per-channel claims were
  sampled rather than table-driven.
- `openapi.json` is regenerated. My earlier note that this produced a ~25k line diff was
  wrong: that was tabs versus spaces, not content.

**One behaviour change worth stating plainly:** changing a saved Matrix channel's home server
now requires re-entering the token. Intended, and a Matrix token is homeserver-specific
anyway, but you should not have to discover it yourself.

1401 tests across 80 suites, up from 1339 across 79 on `develop`.

@ashu130698 thanks for deploying it and actually trying the attack. The guard does already
have dedicated coverage though, in `server/test/unit/services/notificationsService.test.ts`,
including the exact case you reproduced: home server changed, credential omitted, provider
never called. Your DB check is covered by the edit tests. I suspect the Files-changed tab hid
the test file.

If you still want something, there is a real gap. `NotificationDestinationFields` has twelve
entries and only two are exercised, so a wrong one would not fail anything. A test that reads
the map and asserts what it says would pass with a wrong entry too, so the useful version
cross-checks against the providers: drive each with a marker value per candidate field against
the mocked `got`, and assert anything landing in the outgoing host is listed for that channel.
`test/helpers/notificationProviderContract.ts` is the natural home, since every channel goes
through it.

@ajhollid the two decisions in the description are still open, both still one-liners. #3559 is
now closed, so this is the single fix for #3771.

ashu130698:
Hey @edobusy, thanks for the detailed explanation! I took your advice and built the cross-check test for NotificationDestinationFields in notificationProviderContract.ts.

To make it fully robust, the test:

Injects a distinct marker into candidate fields one by one (to prevent strict provider validations from short-circuiting the HTTP requests for other fields).
Parses the intercepted got request and uses an exact match against the hostname to avoid substring collision false-positives (like url vs webhookUrl).
Runs a bi-directional assertion at the end: proving that any field that landed in the host is listed in the map, AND that every field listed in the map actually did dictate the host.
I've updated the 11 HTTP providers to pass their channel type and mock. The entire test suite passes perfectly locally.

Here is the patch if you'd like to drop it into your branch. Let me know if you need any tweaks!



```import { describe, expect, it } from "@jest/globals";
 import type { INotificationProvider } from "../../src/domain/notifications/providers/INotificationProvider.ts";
-import type { Notification } from "../../src/domain/notifications/notification.type.ts";
+import type { Notification, NotificationChannel } from "../../src/domain/notifications/notification.type.ts";
 import type { NotificationMessage } from "../../src/domain/notifications/notification.type.ts";
+import { NotificationDestinationFields } from "../../src/domain/notifications/notification.type.ts";
 
 const makeMessage = (): NotificationMessage => ({
 	type: "monitor_down",
 export const testNotificationProviderContract = (
 	opts: {
 		create: () => INotificationProvider;
 		makeNotification: () => Partial<Notification>;
+		type?: NotificationChannel;
+		getInterceptedRequests?: () => { url: string }[];
 	}
 ) => {
 	describe(`INotificationProvider contract: ${name}`, () => {
 export const testNotificationProviderContract = (
 			const result = await provider.sendMessage(opts.makeNotification() as Notification, makeMessage());
 			expect(typeof result).toBe("boolean");
 		});
+
+		if (opts.type && opts.getInterceptedRequests) {
+			it("destination host is determined strictly by NotificationDestinationFields", async () => {
+				const baseNotification = opts.makeNotification() as Record<string, any>;
+				
+				const allDestinationFields = Array.from(new Set(Object.values(NotificationDestinationFields).flat()));
+				const candidateFields = Array.from(new Set([...allDestinationFields, "accessToken", "webhookPassword", "chatId", "phone", "notificationName", "username", "password", "routingKey"]));
+				
+				const allowedDestinationFields = (NotificationDestinationFields[opts.type!] || []) as string[];
+				const fieldsThatDictatedHost = new Set<string>();
+
+				// Test one field at a time to isolate validation errors
+				for (const field of candidateFields) {
+					const notificationWithMarker = { ...baseNotification };
+					const markerHost = `marker-${field.toLowerCase()}.example.com`;
+					notificationWithMarker[field] = `https://${markerHost}/path`;
+
+					// create() also resets the HTTP mock in our tests!
+					const provider = opts.create();
+
+					try {
+						await provider.sendTestAlert(notificationWithMarker as any);
+					} catch (e) {
+						// Providers might throw validation or connection errors. We ignore them.
+					}
+
+					const requests = opts.getInterceptedRequests!();
+					
+					for (const req of requests) {
+						const url = new URL(req.url);
+						// Exact match to prevent substring collisions
+						if (url.hostname === markerHost) {
+							fieldsThatDictatedHost.add(field);
+						}
+					}
+				}
+
+				// 1. Assert no unauthorized fields dictated the host
+				for (const field of fieldsThatDictatedHost) {
+					expect(allowedDestinationFields).toContain(field);
+				}
+
+				// 2. Assert every authorized field actually DOES dictate the host (Prevents dead/wrong entries)
+				for (const field of allowedDestinationFields) {
+					expect(Array.from(fieldsThatDictatedHost)).toContain(field);
+				}
+			});
+		}
 	});
 };
const createProvider = () => {
 
 testNotificationProviderContract("DiscordProvider", {
 	create: () => {
-		mockGotPost.mockResolvedValue({});
+		mockGotPost.mockReset().mockResolvedValue({});
 		return createProvider().provider;
 	},
 	makeNotification: () => makeNotification(),
+	type: "discord",
+	getInterceptedRequests: () => mockGotPost.mock.calls.map(c => ({ url: c[0] as string })),
 });
 
 describe("DiscordProvider", () => {
 const createProvider = () => {
 
 testNotificationProviderContract("MatrixProvider", {
 	create: () => {
-		mockGotPut.mockResolvedValue({});
+		mockGotPut.mockReset().mockResolvedValue({});
 		return createProvider().provider;
 	},
 	makeNotification: () => makeNotification(),
+	type: "matrix",
+	getInterceptedRequests: () => mockGotPut.mock.calls.map(c => ({ url: c[0] as string })),
 });
 
 describe("MatrixProvider", () => {
const makeNtfyNotification = (overrides = {}) =>
 
 testNotificationProviderContract("NtfyProvider", {
 	create: () => {
-		mockGotPost.mockResolvedValue({});
+		mockGotPost.mockReset().mockResolvedValue({});
 		return createProvider().provider;
 	},
 	makeNotification: () => makeNtfyNotification(),
+	type: "ntfy",
+	getInterceptedRequests: () => mockGotPost.mock.calls.map(c => ({ url: c[0] as string })),
 });
 
 describe("NtfyProvider", () => {
const createProvider = () => {
 
 testNotificationProviderContract("PagerDutyProvider", {
 	create: () => {
-		mockGotPost.mockResolvedValue({});
+		mockGotPost.mockReset().mockResolvedValue({});
 		return createProvider().provider;
 	},
 	makeNotification: () => makeNotification(),
+	type: "pager_duty",
+	getInterceptedRequests: () => mockGotPost.mock.calls.map(c => ({ url: c[0] as string })),
 });
 
 describe("PagerDutyProvider", () => {
 const createProvider = () => {
 
 testNotificationProviderContract("PushoverProvider", {
 	create: () => {
-		mockGotPost.mockResolvedValue({});
+		mockGotPost.mockReset().mockResolvedValue({});
 		return createProvider().provider;
 	},
 	makeNotification: () => makeNotification(),
+	type: "pushover",
+	getInterceptedRequests: () => mockGotPost.mock.calls.map(c => ({ url: c[0] as string })),
 });
 
 describe("PushoverProvider", () => {
 const createProvider = () => {
 
 testNotificationProviderContract("RocketChatProvider", {
 	create: () => {
-		mockGotPost.mockResolvedValue({});
+		mockGotPost.mockReset().mockResolvedValue({});
 		return createProvider().provider;
 	},
 	makeNotification: () => makeNotification({ type: "rocket_chat" }),
+	type: "rocket_chat",
+	getInterceptedRequests: () => mockGotPost.mock.calls.map(c => ({ url: c[0] as string })),
 });
 
 describe("RocketChatProvider", () => {
const createProvider = () => {
 
 testNotificationProviderContract("SlackProvider", {
 	create: () => {
-		mockGotPost.mockResolvedValue({});
+		mockGotPost.mockReset().mockResolvedValue({});
 		return createProvider().provider;
 	},
 	makeNotification: () => makeNotification(),
+	type: "slack",
+	getInterceptedRequests: () => mockGotPost.mock.calls.map(c => ({ url: c[0] as string })),
 });
 
 describe("SlackProvider", () => {
const createNotification = (overrides?: Partial<Notification>): Notification =>
 	...overrides,
 });
 
+import { testNotificationProviderContract } from "../../../helpers/notificationProviderContract.ts";
+
+testNotificationProviderContract("TeamsProvider", {
+	create: () => {
+		mockPost.mockReset().mockResolvedValue({});
+		return new TeamsProvider(createMockLogger() as any);
+	},
+	makeNotification: () => createNotification(),
+	type: "teams",
+	getInterceptedRequests: () => mockPost.mock.calls.map(c => ({ url: c[0] as string })),
+});
+
 const createMessage = (overrides?: Partial<NotificationMessage>): NotificationMessage => ({
 	type: "monitor_down",
 	severity: "critical",
 const createProvider = () => {
 
 testNotificationProviderContract("TelegramProvider", {
 	create: () => {
-		mockGotPost.mockResolvedValue({});
+		mockGotPost.mockReset().mockResolvedValue({});
 		return createProvider().provider;
 	},
 	makeNotification: () => makeNotification(),
+	type: "telegram",
+	getInterceptedRequests: () => mockGotPost.mock.calls.map(c => ({ url: c[0] as string })),
 });
 
 describe("TelegramProvider", () => {
 const makeTwilioNotification = (overrides?: Record<string, unknown>) =>
 
 testNotificationProviderContract("TwilioProvider", {
 	create: () => {
-		mockGotPost.mockResolvedValue({});
+		mockGotPost.mockReset().mockResolvedValue({});
 		return createProvider().provider;
 	},
 	makeNotification: () => makeTwilioNotification(),
+	type: "twilio",
+	getInterceptedRequests: () => mockGotPost.mock.calls.map(c => ({ url: c[0] as string })),
 });
 
 describe("TwilioProvider", () => {
 const createProvider = () => {
 
 testNotificationProviderContract("WebhookProvider", {
 	create: () => {
-		mockGotPost.mockResolvedValue({});
+		mockGotPost.mockReset().mockResolvedValue({});
 		return createProvider().provider;
 	},
 	makeNotification: () => makeNotification(),
+	type: "webhook",
+	getInterceptedRequests: () => mockGotPost.mock.calls.map(c => ({ url: c[0] as string })),
 });
 
 describe("WebhookProvider", () => { ```

Issue #3805, opened 2026-07-28
Tag filtering always return no result


CoreKit-glitch:
**Describe the bug**
On monitor /uptime page, using tag filtering doesn't return a list of monitor that own specified tag

<img width="192" height="181" alt="Image" src="https://github.com/user-attachments/assets/20d36141-03c9-4de2-bee7-782593c93031" />

**To Reproduce**
1. Go to tags menu and create tag
2. Edit a monitor and apply this tag
3. go to uptime menu and apply taf filter

**Expected behavior**
I expect to have a list of monitor that own specified tag(s)

**Screenshots**

<img width="192" height="181" alt="Image" src="https://github.com/user-attachments/assets/b1c964ce-a119-44a7-9c1f-3d23ce1cdb9f" />

**Desktop (please complete the following information):**
 - Browser Firefox 153

**Additional context**
Checkmate v3.9.2

anjupathak03:
Hi, I'd like to work on this issue. I'll reproduce the tag filter failure locally, trace the selected tag values through the React filter state and API query, then add a focused fix and regression coverage. Could you please assign it to me?

anjupathak03:
Update after tracing this: the reported v3.9.2 release () still compares raw tag strings inside the aggregation query, which explains the empty result. Current  already converts tag query values to  in  (introduced in commit ). Since contributions must target , I don't want to submit a duplicate/no-op fix. This appears resolved for the next release; maintainers may want to confirm with a regression test or close it after release.

anjupathak03:
Correction to my previous comment: shell formatting stripped the identifiers. The reported v3.9.2 release on master compares raw tag strings inside the aggregation query. Current develop converts tag query values to mongoose.Types.ObjectId in monitor.repository.mongo.ts, introduced by commit 70f10a73d8669b17c82e5a28f8ef83c8d07300c8. Because contributions must target develop, a new pull request would duplicate the existing fix. This appears resolved for the next release; a regression test could still be useful.

gorkem-bwl:
This is resolved.

PR #3808, opened 2026-07-28
test: cover monitor tag filtering

anjupathak03:
## Describe your changes

Adds a real-Mongo regression test for monitor tag filtering.

The test seeds monitors with matching and non-matching ObjectId tag references, including a monitor from another team, then verifies that:

- the aggregation-backed monitor query returns only the selected team's matching monitor
- the aggregation-backed summary reports the same filtered total

The behavior reported in #3805 is already fixed on `develop` by converting query-string tag IDs to ObjectIds before aggregation. This test locks in that behavior and prevents the raw-string comparison from returning an empty result again.

## Issue

Refs #3805

## Verification

- `npm test -- --runTestsByPath test/integration/monitorRepository.test.ts --runInBand --coverage=false`
- `npm run test:integration -- --runInBand` (11 suites, 81 tests)
- `npm run typeCheck`
- `npm run lint`
- `npm run format-check`

The focused test was also checked against the previous raw-string query implementation and failed as expected with an empty monitor array.

## Checklist

- [ ] I deployed the application locally. This is a draft and a test-only change.
- [x] i18n is not applicable because there are no user-visible strings.
- [x] No production values or styles were hardcoded.
- [x] Theme requirements are not applicable because there is no UI change.
- [x] The PR is granular and targeted to one behavior.
- [x] Server formatting was verified with `npm run format-check`.
- [x] A screenshot is not applicable because there is no UI change.

### DIFF
--- a/server/test/integration/monitorRepository.test.ts
+++ b/server/test/integration/monitorRepository.test.ts
@@ -1,0 +1,54 @@
+import { afterAll, beforeAll, beforeEach, describe, expect, it } from "@jest/globals";
+import mongoose from "mongoose";
+import { MongoMemoryServer } from "mongodb-memory-server";
+import { MonitorModel } from "../../src/domain/monitors/monitor.model.ts";
+import MongoMonitorsRepository from "../../src/domain/monitors/monitor.repository.mongo.ts";
+
+let mongod: MongoMemoryServer;
+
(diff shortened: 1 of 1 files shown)


ajhollid (approved):
Thank you for the tests!

PR #3812, opened 2026-07-29
fix: restore status page custom domains

### LINKED ISSUE #3811: [Bug] Custom domain for status page is broken
**Describe the bug**
The custom domain feature does not work anymore since 3.10
When accessing a configured domain it does not show the linked status page anymore. Instead we are redirected to login page as if it simply did not exist.

**To Reproduce**
1. Create a status page
2. Add custom domain name for status page (status.example.com)
3. Go to public url (status.example.com)
4. Get redirected to login page

**Expected behavior**
Status page should show up when accessing with custom domain name.

 - Browser Any
 - Version 3.10.0

anjupathak03:
## Describe your changes

Restores custom-domain status page routing when `CLIENT_CONFIG_CLIENT_HOST` is not explicitly configured.

The runtime-config migration omitted `clientHost` from `/config.js` by default. On a custom domain, the client therefore fell back to that domain's own `window.location.origin`, treated it as the main Checkmate host, and rendered the login routes instead of resolving the published status page.

This change:

- always exposes the canonical `CLIENT_HOST` through the generated client config
- preserves `CLIENT_CONFIG_CLIENT_HOST` as the explicit higher-priority override
- adds regression coverage for fallback and override behavior
- updates configuration comments and README guidance to describe the actual precedence

Fixes #3811

## Verification

- Deployed locally and verified that the custom-domain status page resolves correctly when `CLIENT_CONFIG_CLIENT_HOST` is not configured.
- `npm test -- --runTestsByPath test/unit/services/settingsService.test.ts --runInBand --coverage=false` — 13 tests passed
- `npm test -- --runInBand --coverage=false` — 74 suites, 1,328 tests passed
- `npm run typeCheck` in `server`
- `npm run lint` in `server`
- `npm run build` in `server` and `client`
- `npm run format-check` in `server` and `client`
- `git diff --check`

## Checklist

- [x] I deployed the application locally and verified custom-domain status-page routing.
- [x] i18n is not applicable because no visible strings were added.
- [x] No production values or styles are hardcoded.
- [x] Theme requirements are not applicable because there is no UI styling change.
- [x] The PR is granular and targeted to one regression.
- [x] Formatting was verified in both server and client.
- [x] A screenshot is not applicable because the change restores routing/configuration behavior without changing the UI.

### DIFF
--- a/README.md
+++ b/README.md
@@ -105,10 +105,10 @@ The web client needs no configuration by default: it calls the API on the same o
 | Variable | Description |
 |---|---|
 | `CLIENT_CONFIG_API_BASE_URL` | Full base URL the client calls the API at, e.g. `https://api.example.com/api/v1`; defaults to same-origin `/api/v1` |
-| `CLIENT_CONFIG_CLIENT_HOST` | Origin used when the client builds absolute links (invites, status pages); defaults to the browser's current origin |
+| `CLIENT_CONFIG_CLIENT_HOST` | Overrides the canonical origin used for absolute links and custom-domain detection; defaults to `CLIENT_HOST` |
 | `CLIENT_CONFIG_LOG_LEVEL` | Browser console log level: `error`, `warn`, `info`, or `debug` (default `error`) |
 
-> **Upgrading from an older image?** The `UPTIME_APP_*` variables (`UPTIME_APP_API_BASE_URL`, `UPTIME_APP_CLIENT_HOST`, `UPTIME_APP_LOG_LEVEL`) are no longer read. In most setups no replacement is needed — the same-origin defaults cover them; if you pointed the client at a different origin, use the `CLIENT_CONFIG_*` equivalents above. The `checkmate-client`, `checkmate-backend`, `checkmate-mongo`, and `checkmate-backend-mono-multiarch` images are no longer updated — switch to `ghcr.io/bluewave-labs/checkmate`, keeping your existing MongoDB service and data volume.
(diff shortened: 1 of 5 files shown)


YoinkingFishy (approved):
Clean fix for the custom domain regression. The fallback to `CLIENT_HOST` is the right approach — keeps the client config always populated without requiring an explicit override. Test coverage for the fallback and explicit-override cases is solid.

One minor note: the PR description says "draft pending" deployment verification but the PR isn't marked as draft. Might want to mark it draft if you can't verify locally yet.

anjupathak03:
Custom-domain verification is complete. I confirmed that the published status page resolves correctly when CLIENT_CONFIG_CLIENT_HOST is not configured. The PR description and checklist have been updated, and this is ready for maintainer review.

m3nu:
The `CLIENT_HOST` fallback is the right fix for #3811, but it also makes `isCustomDomainHost()` live for the first time on every deployment, and the branch it gates is broader than the bug being fixed.

`Routes/index.tsx:60` replaces the *entire* route table with `<Status/>` on a hostname mismatch alone — before anything has checked whether a status page actually exists for that host. So any instance reachable at more than one hostname loses the app on the secondary one: `www` vs apex, or a platform-provided default URL that stays reachable next to a custom domain.

The failure mode is a blank page rather than an error. `Status/index.tsx:104` returns `null` when `/status-page/resolve?domain=…` finds nothing, and that sits above the loading and empty-state branches, so every route on that hostname renders an empty document. It reads as an outage, not a misconfiguration.

There's also no way to opt out once this merges: `clientHost` is now always populated, so no combination of `CLIENT_HOST` / `CLIENT_CONFIG_CLIENT_HOST` restores the current behaviour.

Would it work to gate the status-page routes on the resolve actually returning a status page, and fall back to the normal route table when it doesn't? That keeps #3811 fixed for genuine custom domains, and turns the unknown-host case into the normal app instead of a blank screen.

Context: we host Checkmate for users at PikaPods, where each instance keeps its default `*.pikapod.net` hostname alongside any custom domain — so this would affect every instance with a custom domain configured.

PR #3814, opened 2026-07-29
fix: normalize constant response times

### LINKED ISSUE #3813: [Bug] Equal response times normalize to NaN
### Describe the bug

When two or more checks all have the same response time, both normalization helpers return `NaN` for the chart value.

`NormalizeData` and `NormalizeDataUptimeDetails` calculate the p0 and p95 anchors and then divide by `max - min`. For a constant series those anchors are equal, so the range has zero width. When the result is serialized to JSON, `NaN` becomes `null`, leaving chart consumers without a numeric value.

The existing unit tests already characterize this as a known edge case, but intentionally expect the broken result.

### To reproduce

```ts
NormalizeData(
  [{ responseTime: 50 }, { responseTime: 50 }],
  10,
  100
);
```

Both normalized `responseTime` values are `NaN`.

The same happens to `avgResponseTime` in `NormalizeDataUptimeDetails`.

### Expected behavior

Constant response-time series should produce a finite chart value while retaining the real value in `originalResponseTime` or `originalAvgResponseTime`.

Mapping a zero-width series to `rangeMin` is consistent with how the minimum anchor is normally mapped and keeps the bars visible without inventing relative variation.

anjupathak03:
## Describe your changes

Prevents constant response-time series from normalizing to `NaN`.

Both normalization paths use p0 and p95 as their display anchors. When every value is equal, those anchors are identical and the rescale denominator is zero. The resulting `NaN` becomes `null` when serialized to JSON, so chart consumers receive no numeric height.

This change maps a zero-width series to `rangeMin`, which is the same value the minimum anchor normally receives and keeps the chart visible without inventing variation. Raw response times remain available through `originalResponseTime` and `originalAvgResponseTime`.

The existing characterization tests are updated to assert finite lower-bound values for both raw checks and grouped uptime details.

Fixes #3813

## Verification

- `npm test -- --runTestsByPath test/unit/utils/dataUtils.test.ts --runInBand --coverage=false` — 26 tests passed
- `npm test -- --runInBand --coverage=false` — 74 suites, 1,327 tests passed
- `npm run typeCheck` in `server`
- `npm run lint` in `server`
- `npm run build` in `server` and `client`
- `npm run format-check` in `server` and `client`
- `git diff --check`

## Checklist

- [ ] I deployed the application locally. This is a draft; the pure normalization change is covered by focused and full automated tests.
- [x] i18n is not applicable because no visible strings were added.
- [x] No production values or styles are hardcoded.
- [x] Theme requirements are not applicable because there is no UI styling change.
- [x] The PR is granular and targeted to one edge case.
- [x] Formatting was verified in both server and client.
- [x] A screenshot is not applicable because there is no UI change.

### DIFF
--- a/server/src/utils/dataUtils.ts
+++ b/server/src/utils/dataUtils.ts
@@ -44,6 +44,7 @@ const percentileBy = <T>(arr: T[], percentile: number, value: (item: T) => numbe
 };
 
 const rescale = (raw: number, min: number, max: number, rangeMin: number, rangeMax: number): number => {
+	if (min === max) return rangeMin;
 	const scaled = rangeMin + ((raw - min) * (rangeMax - rangeMin)) / (max - min);
 	return Math.max(rangeMin, Math.min(rangeMax, scaled));
 };
--- a/server/test/unit/utils/dataUtils.test.ts
+++ b/server/test/unit/utils/dataUtils.test.ts
@@ -113,15 +113,11 @@ describe("NormalizeData", () => {
 		expect(result[0]!.originalResponseTime).toBe(0);
 	});
 
-	// Characterization, not endorsement: when every value is equal the p0–p95 window
-	// has zero width and the rescale divides by zero. Fixing this is a planned
-	// follow-up (docs/planning/dataUtils-generic-normalization-refactor.md); this
-	// test pins the current behavior so the fix is a deliberate, visible change.
-	it("produces NaN when all values are equal (division by zero — known edge case)", () => {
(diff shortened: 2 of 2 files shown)


ajhollid:
Normalization is being deprecated in favour of allowing values that exceed p95 to just clip off the graph.  This preserves the natural shape of the response time graph, which is important as we're now graphing the timing details and not just TTFB.

I'm going to close this as we're not using normalization here anymore, but thank you for your work!

PR #3817, opened 2026-07-30
fix: propagate SMTP transport errors in EmailService.sendEmail

### LINKED ISSUE #3757: EmailService.sendEmail swallows SMTP transport errors, making callers' .catch() blocks dead code
**Describe the bug**
EmailService.sendEmail catches transport/SMTP errors internally and never rethrows or returns a rejected promise. Because of this, any caller wrapping the call in .catch() ,for example UserService in user.service.ts, never receives the failure. The .catch() blocks are effectively dead code, and email send failures are silently suppressed instead of being surfaced.

**To Reproduce**
1. Trigger a flow that sends an email through UserService (e.g. registration, password reset, notification).
2. Force an SMTP/transport failure (bad credentials, unreachable host, timeout, etc.).
3. Observe that the .catch() around the sendEmail call in UserService never fires, despite the email failing to send.

**Expected behavior**
EmailService.sendEmail should propagate transport errors , either by rethrowing the caught error or returning a promise that rejects with it ,so calling modules can properly catch and handle email failures (logging, retries, user-facing error messages, etc.).

**Additional context**
I'd like to work on this issue myself if it's approved, happy to submit a PR with the fix (rethrowing/propagating the error in EmailService.sendEmail and verifying UserService's catch blocks handle it correctly).

YoinkingFishy:
## Description

`EmailService.sendEmail` was catching SMTP transport errors and silently swallowing them. Callers wrapping the call in `.catch()` (e.g., `UserService.createUser`) never received the failure, making those catch blocks dead code.

## Changes

- **`sendMail()` catch block** — rethrows the caught transport error after logging it, so callers can properly handle email failures
- **`transporter.verify()` catch block** — rethrows instead of returning `false`, for consistent error propagation
- **Return type** — updated from `Promise<string | false | undefined>` to `Promise<string>` since the method now always returns a `messageId` on success or throws on failure

## Related Issue

Closes #3757

### DIFF
--- a/server/src/service/emailService.ts
+++ b/server/src/service/emailService.ts
@@ -20,7 +20,7 @@ type TemplateCompiler = (template: string) => (context: Record<string, unknown>)
 export interface IEmailService {
 	init(): void;
 	buildEmail(template: string, context: Record<string, unknown>): Promise<string | undefined>;
-	sendEmail(to: string, subject: string, html: string, transportConfig?: EmailTransportConfig): Promise<string | false | undefined>;
+	sendEmail(to: string, subject: string, html: string, transportConfig?: EmailTransportConfig): Promise<string>;
 }
 
 export class EmailService implements IEmailService {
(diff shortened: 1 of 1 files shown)


ajhollid (changes requested):
This PR breaks notification behaviour.  Please review how notifications work before requesting changes.  Thank you!

PR #3822, opened 2026-08-02
feat: Rocket.Chat notifications

### LINKED ISSUE #3821: Add Rocket.Chat incoming webhook notifications

Checkmate supports several notification channels, but Rocket.Chat is not available as a first-class provider. Rocket.Chat users must rely on the generic webhook option, which is less discoverable and does not give Checkmate ownership of a Rocket.Chat-specific payload contract.

**Describe the solution you would like**

Add a Rocket.Chat notification provider configured with a single incoming webhook URL.

For the initial implementation:

- expose Rocket.Chat in the notification channel selector and API schema
- validate HTTP and HTTPS webhook URLs, including self-hosted instances
- support test notifications
- send a text-only JSON payload containing the notification title, summary, monitor name, type, status, URL, optional details, threshold breaches, and a monitor-scoped incident link
- keep the provider consistent with the existing timeout, retry, error handling, and logging conventions

**Describe alternatives you have considered**

The generic webhook provider can target a Rocket.Chat URL, but it does not provide a dedicated UI option or a stable Rocket.Chat-specific integration contract.

mdiniz97:
## Changes
- [x] Add Rocket.Chat incoming webhook provider with text-only payload
- [x] Register and validate the channel across backend and persistence
- [x] Expose Rocket.Chat in the notification UI
- [x] Update OpenAPI and README
- [x] Add provider and validation tests

Fixes #3821

### DIFF
--- a/client/src/Pages/Notifications/components/NotificationsTable.tsx
+++ b/client/src/Pages/Notifications/components/NotificationsTable.tsx
@@ -65,13 +65,7 @@ export const NotificationsTable = ({
 				id: "type",
 				content: t("common.table.headers.type"),
 				render: (row) => {
-					return (
-						<Typography textTransform={"capitalize"}>
-							{row?.type === "rocket_chat"
-								? t("pages.notifications.form.type.rocketChat")
-								: row?.type}
-						</Typography>
-					);
+					return <Typography textTransform={"capitalize"}>{row?.type}</Typography>;
 				},
 			},
 			{
--- a/client/src/Pages/Notifications/components/NotificationsTable.tsx
+++ b/client/src/Pages/Notifications/components/NotificationsTable.tsx
@@ -65,7 +65,13 @@ export const NotificationsTable = ({
 				id: "type",
 				content: t("common.table.headers.type"),
 				render: (row) => {
-					return <Typography textTransform={"capitalize"}>{row?.type}</Typography>;
+					return (
+						<Typography textTransform={"capitalize"}>
+							{row?.type === "rocket_chat"
+								? t("pages.notifications.form.type.rocketChat")
+								: row?.type}
+						</Typography>
+					);
 				},
 			},
 			{
--- a/client/src/Pages/Notifications/create/index.tsx
+++ b/client/src/Pages/Notifications/create/index.tsx
@@ -139,7 +139,11 @@ const NotificationsCreatePage = () => {
 										key={type}
 										value={type}
 									>
-										<Typography textTransform="capitalize">{type}</Typography>
+										<Typography textTransform="capitalize">
+											{type === "rocket_chat"
+												? t("pages.notifications.form.type.rocketChat")
+												: type}
+										</Typography>
 									</MenuItem>
 								))}
 							</Select>
--- a/client/src/Pages/Notifications/create/index.tsx
+++ b/client/src/Pages/Notifications/create/index.tsx
@@ -139,11 +139,7 @@ const NotificationsCreatePage = () => {
 										key={type}
 										value={type}
 									>
-										<Typography textTransform="capitalize">
-											{type === "rocket_chat"
-												? t("pages.notifications.form.type.rocketChat")
-												: type}
-										</Typography>
+										<Typography textTransform="capitalize">{type}</Typography>
 									</MenuItem>
 								))}
 							</Select>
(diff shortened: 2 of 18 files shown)


client/src/Pages/Notifications/components/NotificationsTable.tsx
```diff
@@ -65,7 +65,13 @@ export const NotificationsTable = ({
 				id: "type",
 				content: t("common.table.headers.type"),
 				render: (row) => {
-					return <Typography textTransform={"capitalize"}>{row?.type}</Typography>;
+					return (
+						<Typography textTransform={"capitalize"}>
+							{row?.type === "rocket_chat"
+								? t("pages.notifications.form.type.rocketChat")
+								: row?.type}
+						</Typography>
```
ajhollid: This isn't necessary, notification channel types are not translated

client/src/Pages/Notifications/create/index.tsx
```diff
@@ -139,7 +139,11 @@ const NotificationsCreatePage = () => {
 										key={type}
 										value={type}
 									>
-										<Typography textTransform="capitalize">{type}</Typography>
+										<Typography textTransform="capitalize">
+											{type === "rocket_chat"
+												? t("pages.notifications.form.type.rocketChat")
+												: type}
```
ajhollid: Same here

server/src/domain/notifications/providers/rocketChat.ts
```diff
@@ -0,0 +1,88 @@
...
+			lines.push(
+				"",
+				"Thresholds:",
+				...message.content.thresholds.map(
+					(threshold) => `- ${threshold.metric.toUpperCase()}: ${threshold.formattedValue} (threshold: ${threshold.threshold}${threshold.unit})`
+				)
+			);
+		}
+
+		lines.push("", `Incident: ${message.clientHost}/infrastructure/${message.monitor.id}`);
```
ajhollid: Check how the other providers implement this.  Appending an incident should be put behind a conditional, otherwise this will display when a monitor recovers as well.

server/src/api/validation/notificationValidation.ts
```diff
@@ -23,6 +23,14 @@ export const createNotificationBodyValidation = z.discriminatedUnion("type", [
 		roomId: z.union([z.string(), z.literal("")]).optional(),
 		accessToken: z.union([z.string(), z.literal("")]).optional(),
 	}),
+	// Rocket.Chat notification
+	z.object({
+		notificationName: z.string().min(1, "Notification name is required"),
+		type: z.literal("rocket_chat"),
+		address: z
+			.url({ message: "Please enter a valid Rocket.Chat webhook URL" })
+			.refine((url) => /^https?:\/\//i.test(url), { message: "Rocket.Chat webhook URL must use HTTP or HTTPS" }),
```
ajhollid: Protocol validation is [supported natively](https://zod.dev/api#urls):


```
const schema = z.url({ protocol: /^https$/ });
 
schema.parse("https://example.com"); // ✅
schema.parse("http://example.com"); // ❌
```

ajhollid (changes requested):
Looks good for the most part.  We can remove the localization for the channel type though, there's no need to provide translations for those, at least at this time.

A couple of minor issues here and there to take a look at as well and then we should be good to go!

Solid implementation, thank you for your contribution!

client/src/Pages/Notifications/create/index.tsx
mdiniz97: Removed here as well. The selector now uses the same raw type rendering as the other channels.

server/src/api/validation/notificationValidation.ts
mdiniz97: Replaced the refinement with the native Zod protocol option using /^https?$/. Both server and client still accept HTTPS and self-hosted HTTP while rejecting unsupported protocols.

server/src/domain/notifications/providers/rocketChat.ts
mdiniz97: Guarded the incident line with message.content.incident, matching the other providers. Added regression coverage for messages without incident data.

client/src/Pages/Notifications/components/NotificationsTable.tsx
mdiniz97: Removed the special-case localization. The table now renders the raw channel type like the other providers.

ajhollid (approved):
Awesome, thank you for taking the time to make the requested changes 🚀   This looks good, thank you for your contribution to the project!

mdiniz97:
Thank you for the thoughtful review and for merging this! I’m glad to contribute to Checkmate 🚀

PR #3828, opened 2026-08-07
docs: clarify Docker installation and add VPS deployment guide

### LINKED ISSUE #3823: Deployment documentation is inconsistent regarding MongoDB after the single-image migration
## Description

The latest release notes state:

> Checkmate now ships as a single image: `ghcr.io/bluewave-labs/checkmate`. The old per-service images (`checkmate-client`, `checkmate-backend`, `checkmate-mongo`, `checkmate-backend-mono`) are deprecated.

This suggests that the deployment has been simplified to a single container.

However, the `docker-compose.yaml` provided for the new deployment still defines a separate `mongodb` service, and the `checkmate` container depends on it:

```yaml
environment:
  - DB_CONNECTION_STRING=mongodb://mongodb:27017/uptime_db

depends_on:
  mongodb:
    condition: service_healthy
```

So while the application itself is now distributed as a single image, the deployment still requires a MongoDB container.

## Why this is confusing

Reading the release notes, it is easy to interpret "ships as a single image" as meaning that the whole stack is now self-contained.

The provided compose file contradicts that expectation by still requiring an external MongoDB service.

## Suggestion

Could the documentation/release notes be clarified to explicitly state that:

* only the **Checkmate application** has been consolidated into a single image;
* **MongoDB is still required** as a separate service (unless an external MongoDB instance is used).

Alternatively, if the intention is for MongoDB to be embedded in the future, the compose example should be updated accordingly.

Thanks for the great work!

Glazzze:
## Describe your changes

- Clarified that the deprecated `docker/dist-mono`, `docker/dist`, and `docker/dist-arm` directories must not be used.
- Documented the current single-image Checkmate deployment, including the separate MongoDB requirement.
- Documented that legacy `UPTIME_APP_*` variables are ignored.
- Added a VPS-focused Docker Compose example with Caddy automatic HTTPS.
- Kept port `52345` internal for the VPS stack; only Caddy exposes ports `80` and `443`.

Fixes #3823

## Validation

- Ran `docker compose --env-file <test-env> -f docker/docker-compose.vps.yaml config`.
- Ran `git diff --check`.
- Performed a self-review of the documentation and Compose configuration.

## Checklist

- [ ] I deployed the application locally.
- [x] N/A — this documentation-only change adds no application UI strings.
- [x] N/A — no application UI styling or application configuration values were added.
- [x] My PR is granular and targeted to one issue.
- [ ] I ran `npm run format` in server and client directories.
- [x] N/A — there is no UI change, so no screenshot or video is needed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,7 +89,7 @@ curl -O https://raw.githubusercontent.com/bluewave-labs/checkmate/master/docker/
 JWT_SECRET="$(openssl rand -hex 32)" docker compose up -d
 ```
 
-Then open http://localhost:52345. This is the only current Compose file for a direct HTTP installation; the retired `docker/dist-mono/`, `docker/dist/`, and `docker/dist-arm/` directories must not be used. If the app is reached at another origin (domain or LAN IP), set `CLIENT_HOST` accordingly. To build the image yourself, run `docker build -f docker/Dockerfile -t checkmate .` from a checkout.
+Then open http://localhost:52345. If the app is reached at another origin (domain or LAN IP), set `CLIENT_HOST` accordingly. To build the image yourself, run `docker build -f docker/Dockerfile -t checkmate .` from a checkout.
 
 ### VPS deployment with HTTPS (Caddy)
 
--- a/README.md
+++ b/README.md
@@ -89,7 +89,38 @@ curl -O https://raw.githubusercontent.com/bluewave-labs/checkmate/master/docker/
 JWT_SECRET="$(openssl rand -hex 32)" docker compose up -d
 ```
 
-Then open http://localhost:52345. If the app is reached at another origin (domain or LAN IP), set `CLIENT_HOST` accordingly. To build the image yourself, run `docker build -f docker/Dockerfile -t checkmate .` from a checkout. For TLS, put any reverse proxy (Caddy, Traefik, nginx) in front of port 52345.
+Then open http://localhost:52345. This is the only current Compose file for a direct HTTP installation; the retired `docker/dist-mono/`, `docker/dist/`, and `docker/dist-arm/` directories must not be used. If the app is reached at another origin (domain or LAN IP), set `CLIENT_HOST` accordingly. To build the image yourself, run `docker build -f docker/Dockerfile -t checkmate .` from a checkout.
+
+### VPS deployment with HTTPS (Caddy)
+
+For a public VPS, use the dedicated Compose file below. It runs the same all-in-one Checkmate image and MongoDB, adds Caddy for automatic HTTPS, and deliberately does **not** publish Checkmate's internal port `52345` to the host.
+
+Before starting, point an A/AAAA record for your domain at the VPS and allow inbound TCP ports `80` and `443`. Then download the two VPS deployment files into an empty directory:
+
+```bash
+mkdir checkmate && cd checkmate
+curl -O https://raw.githubusercontent.com/bluewave-labs/checkmate/master/docker/docker-compose.vps.yaml
+curl -O https://raw.githubusercontent.com/bluewave-labs/checkmate/master/docker/Caddyfile
+```
+
+Create `.env`, replacing `checkmate.example.com` with your domain:
+
+```bash
+export CHECKMATE_DOMAIN=checkmate.example.com
+export CLIENT_HOST="https://${CHECKMATE_DOMAIN}"
+export JWT_SECRET="$(openssl rand -hex 32)"
+printf 'CHECKMATE_DOMAIN=%s\nCLIENT_HOST=%s\nJWT_SECRET=%s\n' \
+  "$CHECKMATE_DOMAIN" "$CLIENT_HOST" "$JWT_SECRET" > .env
+```
+
+Start the stack and check its status:
+
+```bash
+docker compose -f docker-compose.vps.yaml up -d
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -80,14 +80,47 @@ Usage instructions can be found [here](https://checkmate.so/docs).
...
+
+> **What “all-in-one” means:** the Checkmate application is packaged in a single image; MongoDB is not embedded in that image and remains required. The reference Compose file starts MongoDB for you. For custom deployments, configure `DB_CONNECTION_STRING` to use an external MongoDB instance.
 
 ```bash
 curl -O https://raw.githubusercontent.com/bluewave-labs/checkmate/master/docker/docker-compose.yaml
 JWT_SECRET="$(openssl rand -hex 32)" docker compose up -d
 ```
 
-Then open http://localhost:52345. If the app is reached at another origin (domain or LAN IP), set `CLIENT_HOST` accordingly. To build the image yourself, run `docker build -f docker/Dockerfile -t checkmate .` from a checkout. For TLS, put any reverse proxy (Caddy, Traefik, nginx) in front of port 52345.
+Then open http://localhost:52345. This is the only current Compose file for a direct HTTP installation; the retired `docker/dist-mono/`, `docker/dist/`, and `docker/dist-arm/` directories must not be used. If the app is reached at another origin (domain or LAN IP), set `CLIENT_HOST` accordingly. To build the image yourself, run `docker build -f docker/Dockerfile -t checkmate .` from a checkout.
```
ajhollid: These docker files/directories no longer exist, this warning isn't needed anymore I don't think

docker/docker-compose.vps.yaml
```diff
@@ -0,0 +1,59 @@
...
+        [
+          "CMD",
+          "node",
+          "-e",
+          "require('http').get('http://localhost:52346/livez',r=>process.exit(r.statusCode===200?0:1)).on('error',()=>process.exit(1))",
+        ]
+      interval: 15s
+      timeout: 3s
+      start_period: 60s
+      retries: 3
```
ajhollid, line 25: Any reason why `start_interval` was removed here?

README.md
```diff
@@ -80,14 +80,47 @@ Usage instructions can be found [here](https://checkmate.so/docs).
 <a id="installation"></a>
 ## Installation
 
-The quickest way to run Checkmate is the reference Docker Compose file, which runs the all-in-one image (`ghcr.io/bluewave-labs/checkmate`) plus MongoDB:
+The quickest way to run Checkmate is the reference Docker Compose file. It starts two services: the all-in-one Checkmate application image (`ghcr.io/bluewave-labs/checkmate`) and a separate MongoDB service.
+
+> **What “all-in-one” means:** the Checkmate application is packaged in a single image; MongoDB is not embedded in that image and remains required. The reference Compose file starts MongoDB for you. For custom deployments, configure `DB_CONNECTION_STRING` to use an external MongoDB instance.
```
ajhollid: This content is already merged from your other PR, suggest you rebase this PR so its diff is accurate

ajhollid (changes requested):
Looks good for the most part.  The old Docker directories/compose files no longer exist so the warning there isn't really relevant anymore and can be removed safely.  Thanks for the contribution!

Glazzze:
Rebased onto the latest develop, so the already-merged README change is no longer included in this PR. I also restored the existing healthcheck start_interval values (2s for Checkmate and 1s for MongoDB).

Glazzze:
Thanks for the review! I’ve removed the obsolete warning about the old Docker directories/compose files in d94347ef4. All formatting, build, and server-test checks are passing now. Could you please take another look when you have a chance?

gorkem-bwl:
@ajhollid everything you asked for is in 👍

ajhollid (changes requested):
I think most of the issues are resolved, one thing that apppears to b emissing is the `NODE_ENV` var though 🤔   If the application runs in dev mode rate limiting is disabled, so it should probably be set.

Other than that, I'm not a devops guy, but won't all requests forwarded from Caddy share the same IP?  If you have multiple users they'll run into the rate limit very quicky.

Glazzze:
Addressed in a43c3cc.

- Set NODE_ENV=production in the VPS Compose example so production rate limiting is enabled.
- Configured Express to trust the single Caddy proxy hop, allowing the rate limiter to use the client IP forwarded by Caddy.

Validation: git diff --check and Docker Compose config expansion (using a temporary compatibility copy because the local Compose v2.10.2 does not support the existing start_interval field).

server/src/app.ts
```diff
@@ -31,6 +31,7 @@ export const createApp = ({
 }) => {
 	const allowedOrigin = envSettings.clientHost;
 	const app = express();
+	app.set("trust proxy", 1);
```
ajhollid: This shouldn't be set by default as it makes an assumption about deployment.  This should default to false by default and be handled by an env var that is validated by the env var validation schema.

ajhollid (changes requested):
Thanks for making the requested changes.

The addition of trust proxy should be configurable and false by default however, as we don't want to make assumptions about what kind of deployment a user has.

Glazzze:
Addressed in 895f4c87d. TRUST_PROXY is now validated, defaults to false, and is explicitly enabled only for the VPS Compose deployment.

PR #3832, opened 2026-08-09
fix: map zero-width response time series to min height (fixes #3813)

### LINKED ISSUE #3813: [Bug] Equal response times normalize to NaN

kash-gg:
**(Please remove this line only before submitting your PR. Ensure that all relevant items are checked before submission.)**

## Describe your changes

Fixes #123

```Javascript
<div>{t('add')}</div>
```

### DIFF
--- a/client/src/Utils/DataUtils.ts
+++ b/client/src/Utils/DataUtils.ts
@@ -109,7 +109,9 @@ export const computeBarHeights = (
 
 	const sorted = values.slice().sort((a, b) => a - b);
 	const cap = percentile(sorted, CAP_PERCENTILE);
-	if (cap <= 0) return values.map(() => MIN_HEIGHT_PCT);
+	if (cap <= 0 || (sorted.length > 0 && sorted[0] === sorted[sorted.length - 1])) {
+		return values.map(() => MIN_HEIGHT_PCT);
+	}
 	return values.map((value) => {
 		return Math.max(MIN_HEIGHT_PCT, Math.min(100, (value / cap) * 100));
 	});


sanmaxdev (changes requested):
The code change handles the current client-side zero-width path correctly, and all five repository checks are green.

Before this is ready to merge, please replace the untouched PR template content and `Fixes #123` with a specific summary, testing details, and `Fixes #3813`. Since this changes chart rendering, please also attach the screenshot or video required by the contribution guide.

ajhollid:
Thanks for the PR here, but this has been resolved server-side already.

PR #3833, opened 2026-08-09
feat(status-page): display active maintenance banner and monitor maintenance alert (#2679)

### LINKED ISSUE #2679: Change status to maintenance mode
There is no way to see that maintenance mode is starting.
It does not show on the status page nor in the uptime monitor itself.
only on the Maintenance page status

I come from uptime kuma
And where maintenance mode starts, all monitors that are set to maintenance mode change state to "maintenance mode"
And also on the status page, you can see this very clearly. And the admin can also add a message
like ETA of return to service, etc.
I would be happy if there was a possibility, for example, that maintenance mode starts, the monitor will change to "maintenance mode - up" so that it will still check but will not send notifications

Thanks!!

Shivam-Kapure:
### Summary
This PR implements maintenance mode visibility on public status pages and monitor detail pages as requested in #2679.

- **Status Pages**: Displays a themed `MaintenanceBanner` when any monitors attached to the status page are undergoing active scheduled maintenance, detailing affected services and estimated completion time (ETA).
- **Monitor Detail Page**: Displays an informative warning alert banner when the viewed monitor is currently in maintenance mode.
- **Backend**:
  - Implemented `getActiveWindowEnd()` utility to calculate end times for one-time and recurring maintenance windows.
  - Updated `StatusPageService.getPublicStatusPagePayload` to query `maintenanceWindowsRepository` for active windows covering the status page monitors.
  - Injected `IMaintenanceWindowsRepository` in `services.api.ts`.
- **Localization**: Added translation strings to `en.json`.

Fixes #2679

### PR Checklist
- [x] Tested the app locally and confirmed changes work as intended.
- [x] Code reviewed and no debug logs or leftover code.
- [x] Included related issue (`Fixes #2679`).
- [x] All user-facing strings use `t(...)` translations with keys in `en.json`.
- [x] Shared themes/tokens used for styling.
- [x] Code passes `npm run lint` and all unit tests (67 test suites, 1262 tests passing).
- [x] Opened against the `develop` branch.

### DIFF
--- a/client/src/Pages/StatusPage/Status/themes/shared/MaintenanceBanner.tsx
+++ b/client/src/Pages/StatusPage/Status/themes/shared/MaintenanceBanner.tsx
@@ -1,0 +1,164 @@
+import Box from "@mui/material/Box";
+import Stack from "@mui/material/Stack";
+import Typography from "@mui/material/Typography";
+import { Wrench, Clock } from "lucide-react";
+import { useTranslation } from "react-i18next";
+import type { Monitor } from "@/Types/Monitor";
+import type { ActiveMaintenanceInfo } from "@/Types/StatusPage";
+import { useStatusPageTheme } from "@/Pages/StatusPage/Status/themes/StatusPageThemeProvider";
+import { formatDateWithTz } from "@/Utils/TimeUtils";
+
+interface Props {
+	activeMaintenances?: ActiveMaintenanceInfo[];
+	monitors: Monitor[];
+	timezone?: string;
+}
+
+export const MaintenanceBanner = ({ activeMaintenances, monitors, timezone }: Props) => {
+	const { t } = useTranslation();
+	const { tokens, timezone: themeTimezone } = useStatusPageTheme();
+	const effectiveTimezone = timezone || themeTimezone;
+
+	if (!activeMaintenances || activeMaintenances.length === 0) {
+		return null;
+	}
+
+	const monitorMap = new Map(monitors.map((m) => [m.id, m.name]));
+
+	return (
+		<Stack
+			spacing={2}
+			sx={{
+				width: "100%",
+				mb: 4,
+			}}
+		>
+			{activeMaintenances.map((mw) => {
+				const affectedNames = mw.monitorIds
+					.map((id) => monitorMap.get(id))
+					.filter(Boolean);
+
+				const formattedEnd = formatDateWithTz(
+					mw.end,
+					"MMM D, YYYY h:mm A",
+					effectiveTimezone
+				);
+
+				const etaText = t("pages.statusPages.maintenanceBanner.eta", {
+					time: `${formattedEnd} (${effectiveTimezone})`,
+				});
+
+				return (
+					<Box
+						key={mw.id}
+						sx={{
+							p: 2.5,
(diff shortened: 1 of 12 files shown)


server/src/domain/status-pages/status-page.service.ts
```diff
@@ -26,7 +29,8 @@ export class StatusPageService implements IStatusPageService {
 	constructor(
 		private statusPagesRepository: IStatusPagesRepository,
 		private settingsService: ISettingsService,
-		private monitorsRepository: IMonitorsRepository
+		private monitorsRepository: IMonitorsRepository,
+		private maintenanceWindowsRepository?: IMaintenanceWindowsRepository
```
ajhollid, line 33: This should be required rather than optional

server/src/utils/maintenanceWindow.ts
```diff
@@ -26,3 +26,35 @@ export const isWindowActive = (window: MaintenanceWindow, now: Date = new Date()
 
 	return false;
 };
+
+export const getActiveWindowEnd = (window: MaintenanceWindow, now: Date = new Date()): Date | null => {
```
ajhollid, line 30: This is really the same logic as `isWindowActive`.  We don't want to duplicate this math as we introduce the possibility of drift, so why don't we just reimplement `isActiveWindow` as `getActiveWindowEnd(window, now) !== null`?

I think that resolves the duplication issue here.

client/src/Pages/Uptime/Details/index.tsx
```diff
@@ -189,6 +190,18 @@ const UptimeDetailsPage = () => {
 				isAdmin={isAdmin}
 				refetch={refetchMonitor}
 			/>
+			{monitor.status === "maintenance" && (
+				<Alert
+					severity="warning"
+					icon={<Wrench size={20} />}
+					sx={{
+						borderRadius: "8px",
```
ajhollid, line 198: This should be `theme.shape.borderRadius`

client/src/Pages/StatusPage/Status/themes/shared/MaintenanceBanner.tsx
```diff
@@ -0,0 +1,164 @@
...
+import { useStatusPageTheme } from "@/Pages/StatusPage/Status/themes/StatusPageThemeProvider";
+import { formatDateWithTz } from "@/Utils/TimeUtils";
+
+interface Props {
+	activeMaintenances?: ActiveMaintenanceInfo[];
+	monitors: Monitor[];
+	timezone?: string;
+}
+
+export const MaintenanceBanner = ({ activeMaintenances, monitors, timezone }: Props) => {
```
ajhollid, line 17: This component is full of hardcoded arbitrary values for dimensions, colors etc.

The theme should be used as much as possible for colors and dimensions, as well as the SPACING and LAYOUT util objects.  Please see other components for implementation reference

ajhollid (changes requested):
The backend changes are solid, however there has been some duplication introduced around calculating maintenance winodws that I think can be cleany resolved.

The front end needs additional work as there are many hardcoded arbitrary values that should be replaced with references to the theme and utility files.  Please see my comments in the code review for details.  Thank you!

PR #3834, opened 2026-08-10
feat: add domain registration expiry monitoring

### LINKED ISSUE #2094: Domain monitoring
A nice addition to the features supported by checkmate

https://github.com/nwesterhausen/domain-monitor

colinjtsen2010:
Fixes #2094

## What changed
- New endpoint GET /monitors/domain/:monitorId that returns the domain registration expiry date from WHOIS (walks up from subdomains, e.g. www.example.com -> example.com)
- Uptime details page shows a "Domain expiry" box next to "Certificate expiry"
- OpenAPI docs and unit tests included

## Checklist
- [x] npm run build (server + client)
- [x] npm run lint
- [x] npm run format-check
- [x] Unit tests pass
- [x] Screenshot of the UI change(shown in screenshot below)

Screenshot

### DIFF
--- a/server/src/api/controllers/controllerUtils.ts
+++ b/server/src/api/controllers/controllerUtils.ts
@@ -3,6 +3,7 @@ import { Monitor, MonitorTypes, type MonitorType } from "@/domain/monitors/monit
 import { UserRole } from "@/domain/users/user.type.js";
 import sslChecker, { SSLDetails } from "ssl-checker";
 import * as whoiser from "whoiser";
+import { parse as parseDomain } from "tldts";
 type SSLCheckerType = typeof sslChecker;
 type WhoisModule = typeof whoiser;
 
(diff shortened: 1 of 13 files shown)


gorkem-bwl:
This looks good to me! CC @ajhollid

server/src/api/controllers/controllerUtils.ts
```diff
@@ -13,6 +15,65 @@ export const fetchMonitorCertificate = async (checker: SSLCheckerType, monitor:
...
+		const firstServerData = Object.values(result)[0];
+		const expiryDate = extractDomainExpiryDate(
+			typeof firstServerData === "object" && firstServerData !== null && !Array.isArray(firstServerData) ? firstServerData : {}
+		);
+		if (expiryDate) {
+			return { domain: candidate, expiryDate };
+		}
+	}
+	throw new Error("Domain expiry not found");
+};
```
ajhollid: This will throw errors on all bare IP addresses.  Not necessarily wrong, but it's polluting the logs something fierce.  That case should probably be handled since it will be very common

server/src/api/controllers/controllerUtils.ts
```diff
@@ -13,6 +15,65 @@ export const fetchMonitorCertificate = async (checker: SSLCheckerType, monitor:
...
+	}
+	return null;
+};
+
+export const fetchMonitorDomain = async (client: WhoisModule, monitor: Monitor): Promise<{ domain: string; expiryDate: string }> => {
+	const monitorUrl = new URL(monitor.url);
+	const labels = monitorUrl.hostname.split(".");
+
+	// WHOIS registries only answer for the registrable domain, so try the full
+	// hostname first and walk up to the parent domain (www.example.com -> example.com).
```
ajhollid: This walk up could potentially produce an incorrect expiry date if a domain doesn't parse correctly.  We could inadvertantly skip it and walk up to a parent domain and get an expiry date for that one 🤔

server/src/api/controllers/controllerUtils.ts
```diff
@@ -13,6 +15,65 @@ export const fetchMonitorCertificate = async (checker: SSLCheckerType, monitor:
...
+	}
+
+	const date = new Date(raw);
+	if (!Number.isNaN(date.getTime())) {
+		return date;
+	}
+	return null;
+};
+
+export const fetchMonitorDomain = async (client: WhoisModule, monitor: Monitor): Promise<{ domain: string; expiryDate: string }> => {
```
ajhollid: There should be some sort of caching here, else we'll be making a fresh request every time we visit a details page.

ajhollid (changes requested):
The happy path here looks decent, but there are some issues with the less than happy path.  Please have a look at my comments in the code review and see what you think.

Thanks for your hard work here!

server/src/api/controllers/controllerUtils.ts
colinjtsen2010: Fixed: bare IP addresses, localhost, and IPv6 hosts now return { domain: null, expiryDate: null } without calling WHOIS at all (tldts detects them), so no more error-log spam. The client already renders N/A for a null expiry.

server/src/api/controllers/controllerUtils.ts
colinjtsen2010: Fixed: replaced the label walk-up with tldts to extract the exact registrable domain (e.g. www.example.co.uk -> example.co.uk). We now query only the registrable domain, so a subdomain can never accidentally fall back to a parent domain's expiry date.

server/src/api/controllers/controllerUtils.ts
colinjtsen2010: Fixed: added a TTL cache keyed by registrable domain (1 hour for results, 10 minutes for no-expiry results), so visiting a details page reuses the previous lookup instead of re-querying the registry. Unit tests cover cache hit and expiry.

colinjtsen2010:
@ajhollid I've addressed all three review comments in 75516a2:

1. Bare IPs / localhost / IPv6 now return a null expiry without querying WHOIS (no more error-log spam).
2. Replaced the label walk-up with tldts so we only ever query the exact registrable domain (www.example.co.uk -> example.co.uk), preventing an incorrect parent-domain fallback.
3. Added a TTL cache keyed by registrable domain (1h positive, 10min negative) so details pages reuse lookups.

Unit tests (9/9), typecheck, lint, and the server build all pass locally. Could you take another look when you get a chance?

ajhollid (approved):
Looks good, thank you for making changes.  The npm lockfile was pointing at a mirror, I assume this is an artefact of the local dev env you've got set up.  I've pointed the lockfile back at the official registry and commited the changes.

Thanks for your contribution to the project!

PR #3837, opened 2026-08-14
fix: propagate SMTP transport errors from EmailService.sendEmail

### LINKED ISSUE #3757: EmailService.sendEmail swallows SMTP transport errors, making callers' .catch() blocks dead code

egeoztass:
Closes #3757

## Problem

`EmailService.sendEmail` caught transport/SMTP errors internally and resolved to `false` or `undefined` instead of rejecting. Every caller that wrapped it in `.catch()` was therefore dead code, and delivery failures were silently suppressed.

The most visible symptom is password recovery: with a broken SMTP config the endpoint still responds `200 "Password recovery email sent successfully"` while the user never receives anything.

## Change

`sendEmail` now throws an `AppError` carrying the underlying cause, for both failure paths (`transporter.verify()` and `sendMail()`). The signature tightens from `Promise<string | false | undefined>` to `Promise<string>`.

Existing logging is unchanged — the throw is added alongside it, not in place of it.

## Call sites

All five were updated for the new contract:

| Call site | Behaviour |
|---|---|
| `UserService.registerUser` | Existing `.catch()` now actually fires; still fire-and-forget, still logs a warning. |
| `UserService.requestRecovery` | Failure now reaches the client instead of a misleading success response. |
| `InviteService.sendInviteEmail` | Keeps its existing user-facing message, attaches the SMTP cause in `details`. |
| `settingsController` test email | Appends the underlying SMTP error to the message — diagnosing mail settings is the point of that endpoint. |
| `EmailProvider` | Catches and returns `false`. |

The `EmailProvider` case is the one worth a second look: `NotificationsService` fans providers out through `Promise.all`, so letting the rejection escape would abort the *other* channels' alerts whenever email failed. Providers must keep resolving to a boolean.

## Tests

Seven existing tests asserted the old falsy-return contract and were rewritten to assert rejection. New coverage added for:

- the underlying cause being preserved on the thrown `AppError`
- `registerUser` logging a warning when the welcome email fails
- `requestRecovery` propagating the failure
- `EmailProvider` returning `false` rather than rejecting

`npm test` — 79 suites, 1342 tests passing. `tsc --noEmit`, `eslint`, and `prettier --check` all clean.

### DIFF
--- a/server/src/api/controllers/settingsController.ts
+++ b/server/src/api/controllers/settingsController.ts
@@ -93,24 +93,30 @@ class SettingsController implements ISettingsController {
 		if (!html) {
 			throw new AppError({ message: "Failed to build email template.", status: 500 });
 		}
-		const messageId = await this.emailService.sendEmail(to, subject, html, {
-			systemEmailHost,
-			systemEmailPort,
-			systemEmailUser,
-			systemEmailAddress,
-			systemEmailDisplayName,
-			systemEmailPassword,
-			systemEmailConnectionHost,
-			systemEmailSecure,
-			systemEmailPool,
-			systemEmailIgnoreTLS,
-			systemEmailRequireTLS,
-			systemEmailRejectUnauthorized,
-			systemEmailTLSServername,
-		});
-
-		if (!messageId) {
-			throw new AppError({ message: "Failed to send test email.", status: 500 });
+		let messageId: string;
+		try {
+			messageId = await this.emailService.sendEmail(to, subject, html, {
+				systemEmailHost,
+				systemEmailPort,
+				systemEmailUser,
+				systemEmailAddress,
+				systemEmailDisplayName,
+				systemEmailPassword,
+				systemEmailConnectionHost,
+				systemEmailSecure,
+				systemEmailPool,
+				systemEmailIgnoreTLS,
+				systemEmailRequireTLS,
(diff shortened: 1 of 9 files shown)


ajhollid (approved):
Good changes here, appreciate the hard work!

PR #3838, opened 2026-08-14
fix: resolve Matrix room aliases and encode room IDs in the send path

### LINKED ISSUE #3478: Matrix notifications are not going through

egeoztass:
Closes #3478

## Problem

Matrix notifications fail for anyone who configures the channel with a **room alias** rather than a raw room ID.

The send endpoint takes a room ID (`!id:server`), but Element displays the alias (`#name:server`), so that's the value most people paste into the form. An unresolved alias doesn't just get rejected by the homeserver — it breaks the URL before the request is even sent, because the leading `#` opens a URI fragment:

```
input   #alerts:example.com
url     https://matrix.example.com/_matrix/client/v3/rooms/#alerts:example.com/send/m.room.message/<txn>
parsed  path = /_matrix/client/v3/rooms/
        hash = #alerts:example.com/send/m.room.message/<txn>
```

The PUT lands on `/rooms/` instead of the send endpoint. The provider catches the resulting error, logs a warning, and returns `false` — so the channel just silently never delivers.

## Change

Three fixes to `MatrixProvider`, all in the request-building path:

1. **Resolve aliases.** An identifier starting with `#` is resolved through `GET /_matrix/client/v3/directory/room/{roomAlias}` before sending, per the client-server spec. Anything else is treated as a room ID and passed through, so existing working configs are unaffected.
2. **Percent-encode the room ID** in the send path. `:` was previously sent raw.
3. **Strip trailing slashes** from the homeserver URL, so `https://matrix.example.com/` no longer produces `//_matrix/...`.

Alias resolution failures (unknown alias, or a directory response with no `room_id`) return `false` and log a warning, matching how the provider already handles send failures.

## Tests

Eight new cases covering alias resolution, path encoding, the no-unencoded-`#` invariant, resolution failures, and trailing-slash normalisation — for both `sendTestAlert` and `sendMessage`. All eight fail against the current provider and pass with the fix.

`npm test` — 79 suites, 1347 tests passing. `tsc --noEmit`, `eslint`, and `prettier --check` all clean.

## Note

The original reporter mentioned they'd open a PR with their findings back in April but none appeared, so I've picked it up. If they were working from a different root cause, happy to compare notes.

### DIFF
--- a/server/src/domain/notifications/providers/matrix.ts
+++ b/server/src/domain/notifications/providers/matrix.ts
@@ -38,10 +38,13 @@ export class MatrixProvider extends NotificationProvider {
 
 	private sendMessageWithBody = async (notification: Partial<Notification>, body: unknown): Promise<boolean> => {
 		const { homeserverUrl, accessToken, roomId } = notification;
-		const txnId = randomUUID();
-		const url = `${homeserverUrl}/_matrix/client/v3/rooms/${roomId}/send/m.room.message/${txnId}`;
+		const baseUrl = this.normalizeHomeserverUrl(homeserverUrl as string);
 
 		try {
+			const resolvedRoomId = await this.resolveRoomId(baseUrl, accessToken as string, roomId as string);
+			const txnId = randomUUID();
+			const url = `${baseUrl}/_matrix/client/v3/rooms/${encodeURIComponent(resolvedRoomId)}/send/m.room.message/${txnId}`;
+
 			await got.put(url, {
 				json: body,
 				headers: {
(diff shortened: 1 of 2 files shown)


ajhollid (approved):
Looks good to me, thanks for working on the project!

PR #3844, opened 2026-08-17
fix(security): redact sensitive fields from client-side logs

### LINKED ISSUE #3843: [Security] Login password logged in plaintext to browser console on failed request
**Describe the bug**
When a login request fails (for example, when the backend returns 502 because the database is unreachable), the frontend writes the full request body to the browser console, including the user's plaintext password. Credentials should never be written to logs at any layer, including client-side, as browser consoles are routinely captured by error-reporting tools, screen-shares, and support sessions.

**To Reproduce**
1. Stop the MongoDB service so the backend cannot serve auth requests (the frontend still loads).
2. Restart the checkmate-server service; it stays up and serves the frontend despite the DB being down.
3. Open the browser dev tools console.
4. Attempt to log in with any credentials.
5. Observe the failed POST to /api/v1/auth/login, and note the logged error object contains body: { email, password } with the password in plaintext.

**Observed**
The error log entry for the failed POST includes:
```
"context":{"endpoint":"/auth/login","body":{"email":"[REDACTED]","password":"[REDACTED]"}}
```

**Expected behavior**
Sensitive fields (password, tokens, secrets) should be redacted or omitted from any client-side logging. Common approaches: strip known sensitive keys before logging the request body, or log only metadata (endpoint, status) rather than the full payload for auth endpoints.

 - Reproduced in Chrome; the logging is in the frontend bundle so it is browser-independent.
 - Checkmate version v3.10.0

gorkem-bwl:
Fixes #3843.

A failed request wrote the full request body to the browser console, so a failing POST to `/auth/login` printed the user's plaintext password where it could be picked up by error-reporting tools, screen shares, or support sessions.

## Scope

The reported path was login, but the same unredacted body reached the console from several places:

| Path | In the report | Notes |
|---|---|---|
| POST → `/auth/login`, `/auth/register` | yes | plaintext password |
| PATCH / PUT | no | carry monitor `secret` values |
| `error()` / `warn()` debug branch | no | printed raw `context`, bypassing the JSON path |
| `debug()` / `info()` | no | printed their raw `data` argument |
| The `Error` object itself | no | see below |

GET and DELETE only ever logged `endpoint`, so they were already clean.

The last row is the one worth flagging. An `AxiosError` carries the outgoing request on `config` — `config.data` (the serialized body) and `config.headers.Authorization` (set for every request by the interceptor in `ApiClient.ts`). So redacting the `context` alone left the same password reachable through the error argument. Errors now go through `toSafeError()`. Because Axios serializes the body to a *string*, which key-based redaction cannot see into, it is parsed before redaction; a body that is not JSON is replaced wholesale rather than logged raw.

## Approach

Redaction lives in the logger rather than at the three `UseApi.ts` call sites, so every current and future caller is covered by construction.

Keys are matched case-insensitively on a substring after stripping dashes and underscores, so `password`, `newPassword`, `X-Api-Key` and `Authorization` are all caught. The walk copies rather than mutates the caller's object, handles class instances as well as object literals, cuts genuine cycles without mislabelling repeated siblings, and renders `FormData`/`Blob` as a type marker instead of walking them.

## Note on severity

Error-level logging is on by default (`configuredLevel` falls back to `"error"`), so this was **not** limited to debug builds. `configuredLevel` also reads `runtimeConfig.logLevel`, which is runtime config — meaning `debug` is reachable on a deployed instance, and the debug-branch leaks above were reachable in production too.


ajhollid (approved):
I removed the tests since we don't have an established test routine yet for the client. Other than that looks good!

PR #3848, opened 2026-08-18
fix(monitors): check TLS certificate on the monitor's own port

### LINKED ISSUE #3646: SSL/TLS certificate expiry not checked on non standard https port

gorkem-bwl:
Fixes #3646.

The certificate expiry box showed "N/A" for any HTTPS monitor on a non-standard port.

## Cause

`fetchMonitorCertificate` parsed the monitor URL but passed only the hostname to `ssl-checker`, which then defaults to port 443. A monitor on `https://host:54321` therefore had the certificate of whatever answered on 443 inspected — or nothing at all, surfacing as "N/A".

## Fix

Pass the port from the monitor URL. `URL.port` is an empty string for the scheme's default port, so that case is mapped to 443 explicitly rather than through `Number()`, which would yield `NaN`. `ssl-checker` already accepts a port via `SSLOptions`, so there is no dependency or signature change.

Behaviour changes **only** for URLs that explicitly name a port:

| URL | before | after |
|---|---|---|
| `https://host` | 443 | 443 |
| `https://host:54321` | 443 | **54321** |
| `http://host` | 443 | 443 |
| `http://host:8080` | 443 | rejected (see below) |

## Second commit: plain-HTTP guard

Review surfaced an adjacent gap. Once the real port is passed, a plain-HTTP monitor on a non-default port (`http://host:8080`) would get a TLS handshake attempted against a port serving no TLS, returning a 500 carrying a raw OpenSSL error (`packet length too long`) instead of the previous silent "N/A".

The endpoint never checked the scheme — it was already wrong for `http://` monitors, just quietly so, since every such lookup went to 443. It now guards on the protocol and fails with a clean 400 before opening a socket.

## Verification

- 8 regression tests: the reported non-standard-port case, the default-port fallback, an explicit `:443`, a NaN guard, and both plain-HTTP cases (each asserting the checker is never invoked).
- Verified by reverting: the 3 port assertions fail without the port fix; both HTTP assertions fail without the scheme guard.
- Edge cases checked — IPv6 literals, embedded credentials, and trailing-dot hosts all derive correctly.
- `npm run build` clean; full server suite **1379 passing across 82 suites**.

### DIFF
--- a/server/src/api/controllers/controllerUtils.ts
+++ b/server/src/api/controllers/controllerUtils.ts
@@ -11,9 +11,14 @@ const DEFAULT_HTTPS_PORT = 443;
 
 export const fetchMonitorCertificate = async (checker: SSLCheckerType, monitor: Monitor): Promise<SSLDetails> => {
 	const monitorUrl = new URL(monitor.url);
+	// A plain-HTTP monitor serves no certificate. Reject before dialing, so the
+	// caller gets this instead of a raw OpenSSL handshake error from the socket.
+	if (monitorUrl.protocol !== "https:") {
+		throw new AppError({ message: "Certificate is only available for HTTPS monitors", status: 400 });
+	}
 	const hostname = monitorUrl.hostname;
-	// URL.port is "" for default ports, so fall back to 443; without this the
-	// certificate of a monitor on a non-standard port was probed on 443.
+	// URL.port is "" for the scheme's default port, so fall back to 443; without
+	// this the certificate of a monitor on a non-standard port was probed on 443.
 	const port = monitorUrl.port === "" ? DEFAULT_HTTPS_PORT : Number(monitorUrl.port);
 	const cert = await checker(hostname, { port });
 	if (cert?.validTo === null || cert?.validTo === undefined) {
--- a/server/src/api/controllers/controllerUtils.ts
+++ b/server/src/api/controllers/controllerUtils.ts
@@ -8,6 +8,9 @@ type SSLCheckerType = typeof sslChecker;
 type WhoisModule = typeof whoiser;
 
 const DEFAULT_HTTPS_PORT = 443;
+// Ports are user-defined, so a probe can dial something that accepts the
+// connection and never completes the handshake. Bound the wait explicitly.
+const CERTIFICATE_TIMEOUT_MS = 10_000;
 
 export const fetchMonitorCertificate = async (checker: SSLCheckerType, monitor: Monitor): Promise<SSLDetails> => {
 	const monitorUrl = new URL(monitor.url);
@@ -20,7 +23,7 @@ export const fetchMonitorCertificate = async (checker: SSLCheckerType, monitor: 
 	// URL.port is "" for the scheme's default port, so fall back to 443; without
 	// this the certificate of a monitor on a non-standard port was probed on 443.
 	const port = monitorUrl.port === "" ? DEFAULT_HTTPS_PORT : Number(monitorUrl.port);
-	const cert = await checker(hostname, { port });
+	const cert = await checker(hostname, { port, timeout: CERTIFICATE_TIMEOUT_MS });
 	if (cert?.validTo === null || cert?.validTo === undefined) {
 		throw new Error("Certificate not found");
 	}
--- a/server/src/api/controllers/controllerUtils.ts
+++ b/server/src/api/controllers/controllerUtils.ts
@@ -7,10 +7,15 @@ import { parse as parseDomain } from "tldts";
 type SSLCheckerType = typeof sslChecker;
 type WhoisModule = typeof whoiser;
 
+const DEFAULT_HTTPS_PORT = 443;
+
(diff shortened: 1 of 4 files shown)


server/src/api/controllers/controllerUtils.ts
```diff
@@ -7,10 +7,20 @@ import { parse as parseDomain } from "tldts";
...
+	// A plain-HTTP monitor serves no certificate. Reject before dialing, so the
+	// caller gets this instead of a raw OpenSSL handshake error from the socket.
+	if (monitorUrl.protocol !== "https:") {
+		throw new AppError({ message: "Certificate is only available for HTTPS monitors", status: 400 });
+	}
 	const hostname = monitorUrl.hostname;
-	const cert = await checker(hostname);
+	// URL.port is "" for the scheme's default port, so fall back to 443; without
+	// this the certificate of a monitor on a non-standard port was probed on 443.
+	const port = monitorUrl.port === "" ? DEFAULT_HTTPS_PORT : Number(monitorUrl.port);
```
ajhollid, line 25: If we're going to allow arbitrary user-defined ports here we should add a timeout to the ssl checker request or we could hang indefinitely

server/test/unit/utils/monitorCertificate.test.ts
```diff
@@ -0,0 +1,97 @@
+import { describe, expect, it } from "@jest/globals";
+import { fetchMonitorCertificate } from "../../../src/api/controllers/controllerUtils.ts";
+import type { Monitor } from "../../../src/domain/monitors/monitor.type.ts";
+import type { SSLDetails, SSLOptions } from "ssl-checker";
+
+// Regression cover for #3646: the certificate probe parsed the monitor URL but
+// dropped its port, so a monitor on a non-standard port had its certificate
+// checked on 443 and surfaced as "N/A" in the UI.
```
ajhollid: This comment can be removed

server/src/api/controllers/controllerUtils.ts
```diff
@@ -7,10 +7,20 @@ import { parse as parseDomain } from "tldts";
...
 
+const DEFAULT_HTTPS_PORT = 443;
+
 export const fetchMonitorCertificate = async (checker: SSLCheckerType, monitor: Monitor): Promise<SSLDetails> => {
 	const monitorUrl = new URL(monitor.url);
+	// A plain-HTTP monitor serves no certificate. Reject before dialing, so the
+	// caller gets this instead of a raw OpenSSL handshake error from the socket.
+	if (monitorUrl.protocol !== "https:") {
+		throw new AppError({ message: "Certificate is only available for HTTPS monitors", status: 400 });
+	}
```
ajhollid, line 21: The details page currently makes a request to fetch monitor certificates on 443 regardless of monitor type.  Now that we're guaranteed to throw an error on non https type monitors we'll get an error on every visit to the details page.

The client should have the same gating on protocol added as well so we don't make requests to fetch certificates for non-https monitors

ajhollid (changes requested):
Looks good for the most part, some minor changes needed and we're good to go!

server/src/api/controllers/controllerUtils.ts
gorkem-bwl: Added. `CERTIFICATE_TIMEOUT_MS = 10_000` is now passed to the checker.

`ssl-checker` already wires a `timeout` listener that rejects with "Timed Out", but its defaults set no timeout so it never fired — passing one activates it. The bundled type declaration is a hand-maintained subset of `https.RequestOptions` and had omitted `timeout`, so it is added there too. Covered by a new assertion in the test file.

server/test/unit/utils/monitorCertificate.test.ts
gorkem-bwl: Removed.

server/src/api/controllers/controllerUtils.ts
gorkem-bwl: Good catch — you were right that the details page would 400 on every visit.

Gated client side to match. The subtlety is that `http` is a single monitor type covering both schemes, so an `http://` monitor passed the existing `monitor?.type !== "http"` check and requested a certificate anyway. `isHttpsUrl` now parses the monitor URL and the request is skipped unless the scheme is `https:`, so the stat box falls back to "N/A" as it did before rather than surfacing an error.

PR #3857, opened 2026-08-20
feat(monitors): route checks through an outbound HTTP proxy

gorkem-bwl:
Adds outbound HTTP proxy support for monitor checks, for deployments whose egress is restricted.

Raised in Discord: checks could not reach anything through a corporate proxy, and setting `https_proxy` / `http_proxy` / `NODE_USE_ENV_PROXY` had no effect.

## Why the environment variables did nothing

Two independent reasons, so it was not a misconfiguration:

1. **got does not read proxy environment variables.** Unlike curl or axios, got has never honoured `HTTP_PROXY`/`HTTPS_PROXY` — it requires an explicit agent. `NODE_USE_ENV_PROXY` applies to Node's built-in `fetch`, which got does not use.
2. **Any agent would have been discarded anyway.** `HttpProvider` overwrote `options.agent` on every single request.

## Approach

Proxy selection is resolved per check and applied at the point the agent is chosen. `NO_PROXY` handling is delegated to `proxy-from-env` rather than reimplemented — wildcards, host lists and port-qualified entries then behave exactly as operators already expect from curl.

Both dependencies were already present transitively, so this adds 76K and no new install: `https-proxy-agent` (52K) and `proxy-from-env` (24K).

## Infrastructure monitors are excluded by default

This is the part worth reviewing closely. `hardware` and `pagespeed` checks **both delegate to `HttpProvider`**:

```
HttpProvider ──┬── http        (public sites → proxy)
               ├── hardware    (Capture agents on the LAN → direct)
               └── pagespeed   (Google API → proxy)
```

A global proxy would therefore also capture Capture agents, which normally sit on the internal network — proxying them fails or hairpins. So the default is `http,pagespeed`, and infrastructure stays direct.

`PROXY_MONITOR_TYPES` is the escape hatch for deployments whose agents genuinely are only reachable through the proxy.

## Configuration

```bash
HTTP_PROXY=http://proxy.corp.example:3128
HTTPS_PROXY=http://user:pass@proxy.corp.example:3128   # credentials supported
NO_PROXY=localhost,127.0.0.1,*.internal,10.0.0.5
PROXY_MONITOR_TYPES=http,pagespeed                      # default; "none" disables
```

Invalid entries are dropped and logged rather than silently ignored, so a typo cannot quietly disable proxying:

- unknown types (`htpp`) → warned
- socket-based types (`ping`, `port`, `docker`, `grpc`, `websocket`, `dns`, `game`) → warned, since an HTTP proxy cannot carry them at all

## Agent caching

Agents are cached by proxy URL **and** TLS mode. Keying on TLS matters: sharing one agent between a strict monitor and an `ignoreTlsErrors` monitor would leak the relaxed setting across monitors. Caching also preserves connection pooling, so 50 monitors share tunnels rather than dialling the proxy on every check.

## TLS handling (second commit)

Review caught a real bug in the first commit: `ignoreTlsErrors` was silently dropped for proxied monitors. `rejectUnauthorized` was passed to the `HttpsProxyAgent` constructor, where it governs the socket **to the proxy**, not the tunnelled handshake with the target.

The consequence was worse than a no-op — a monitor with `ignoreTlsErrors` against a self-signed endpoint works today, and configuring a proxy would have made it fail verification, flip to down, and fire a false incident with notifications. The option is now stated per request.

Two other review findings, both about failing loudly rather than silently:

- A fully-invalid `PROXY_MONITOR_TYPES` (e.g. a typo) disabled proxying entirely, which in a restricted-egress deployment takes every HTTP monitor down at once. It now falls back to the defaults and warns.
- `none` only matched as the sole entry, so `none,http` half-enabled proxying. It is now honoured wherever it appears.

## Verification

22 policy tests plus 10 provider-level tests covering the type policy, `NO_PROXY` (hosts and wildcards), authenticated proxies, scheme-specific selection, the `none` sentinel, invalid input, and agent-cache identity.

Behaviour was also verified end to end through `HttpProvider` with a stub transport, asserting which agent is actually selected:

| scenario | result |
|---|---|
| http monitor, default | **proxied** |
| hardware monitor, default | direct |
| hardware monitor, opted in | **proxied** |
| http monitor, `PROXY_MONITOR_TYPES=none` | direct |
| http monitor, `NO_PROXY` match | direct |
| proxied monitor, `ignoreTlsErrors: true` | `rejectUnauthorized: false` |
| proxied monitor, strict | `rejectUnauthorized: true` |

The TLS assertions were verified by reverting the fix — both fail without it.

`npm run build` clean; full suite **1437 passing across 82 suites**.

## Not covered

Notification providers (Discord, Slack, Matrix, Telegram, PagerDuty, …) still call out directly and are unaffected. In a network locked down enough to need a proxy for checks, alerts will likely fail too — worth a follow-up, but kept out of scope here.


ajhollid:
Closing as already implemented

PR #3878, opened 2026-08-26
perf(docker): reduce MongoDB healthcheck overhead

### LINKED ISSUE #3877: Reduce MongoDB healthcheck overhead in Docker Compose
**Describe the bug**

The MongoDB health check in both Docker Compose configurations starts a new `mongosh` process every five seconds. Because the health-check process runs inside the MongoDB container, its startup cost is included in the container's CPU usage and produces recurring CPU spikes even when database activity is low.

**To Reproduce**

1. Start Checkmate with Docker Compose.
2. Run a small set of active monitors.
3. Observe the MongoDB container with `docker stats`.
4. Compare the periodic container CPU spikes with MongoDB operation counters or `mongotop` output.

In a local test with six active monitors, the MongoDB container periodically reached 56.64% and 47.10% CPU. During a 30-second sample, MongoDB averaged 0.43 inserts/s, 6.87 queries/s, 2.33 updates/s, and 4.43 commands/s, while collection activity measured by `mongotop` remained within a few milliseconds per second.

**Expected behavior**

The health check should retain the semantic `mongosh` ping while avoiding unnecessary steady-state process startups. Startup checks should remain frequent so `depends_on: condition: service_healthy` is not delayed.

- Environment: Docker Desktop on macOS
- MongoDB image: `mongo:8.0`
- Checkmate branch: `develop`

**Additional context**

A 30-second steady-state interval with a 30-second start period and a two-second start interval preserves fast startup detection while reducing steady-state `mongosh` executions by approximately 83%. A five-second timeout and five retries retain a reasonable failure-detection window.

**Self-hosted reproduction**

The same behavior was reproduced on a Dokploy-managed self-hosted Checkmate deployment running on a 2-vCPU, 2-GB RAM server. The MongoDB container was limited to 0.75 CPU and 512 MB RAM.

With the current five-second healthcheck interval, a 20-sample observation recorded:

- 4.56% median MongoDB container CPU
- 77.85% p95 MongoDB container CPU
- 78.44% maximum MongoDB container CPU

The application container remained comparatively quiet during the same observation. This supports the local finding that the recurring spikes are dominated by healthcheck process startup rather than monitor write volume.

Recurring MongoDB CPU spikes with the five-second healthcheck

A proposed fix and before/after validation are available in #3878.

eralpozcan:
## Describe your changes

Replace the MongoDB healthchecks that start a new `mongosh` process with a lightweight TCP connection check.

- Keep the existing five-second healthcheck cadence and failure-detection settings.
- Check whether MongoDB accepts a connection on `127.0.0.1:27017` using Bash's built-in `/dev/tcp` support.
- Apply the same healthcheck to the production, development, and `docs/hosting/checkmate-so` Compose configurations.

### Review-driven revision

The initial version reduced the expensive `mongosh` probe frequency from every 5 seconds to every 30 seconds. Following maintainer review, the implementation was revised to remove the expensive process startup instead of running it less often.

| Area | Initial proposal | Revised implementation |
| --- | --- | --- |
| Healthcheck command | `mongosh` ping | Bash TCP connection check |
| Steady-state interval | 30 seconds | Existing 5 seconds retained |
| Startup/failure timings | Adjusted | Existing values retained |
| Compose coverage | Production and development | Production, development, and both `docs/hosting/checkmate-so` MongoDB services |

All benchmark results below were rerun against the revised TCP implementation.

Fixes #3877

### Local Docker reproduction

- Reproduced the issue with six active monitors; five generated test monitors ran at 10-second intervals.
- With the original five-second `mongosh` healthcheck, MongoDB container CPU showed recurring peaks of 56.64%, 47.10%, and 27.48%. A later comparable sample observed a 46.93% peak.
- During a 30-second observation window, application activity remained modest: 0.43 inserts/s, 6.87 queries/s, 2.33 updates/s, and 4.43 commands/s.
- `mongotop` showed only a few milliseconds of activity per second on the `checks`, `jobs`, and `monitors` collections, separating normal database work from healthcheck process startup.
- Verified that the TCP command exits successfully against MongoDB on port 27017 and fails against a closed port.
- Started MongoDB with a new empty volume and confirmed that the container reached `healthy` using the TCP healthcheck.
- With the TCP healthcheck still running every five seconds, the comparable local sample recorded a 7.44% maximum MongoDB container CPU.
- Verified all three Compose files with `docker compose config --quiet`.

### Self-hosted validation

Validated on a Dokploy-managed self-hosted Checkmate deployment running on a 2-vCPU, 2-GB RAM server. The MongoDB container was limited to 0.75 CPU and 512 MB RAM. Both measurements use the same five-second healthcheck cadence.

| Metric | Before (`mongosh`) | After (TCP) | Change |
| --- | ---: | ---: | ---: |
| Healthcheck executions per minute | 12 | 12 | unchanged |
| MongoDB CPU median (20 samples) | 4.56% | 0.83% | -81.8% |
| MongoDB CPU p95 (20 samples) | 77.85% | 6.23% | -92.0% |
| MongoDB CPU maximum (20 samples) | 78.44% | 7.08% | -91.0% |

Before:

MongoDB CPU before the healthcheck change

After:

MongoDB CPU after the TCP healthcheck change

- [x] I have added i18n support to visible strings (not applicable; no user-facing strings changed).
- [x] I have **not** included any files that are not related to my pull request, including package-lock and package-json if dependencies have not changed.
- [x] I didn't use any hardcoded application values (port 27017 is the MongoDB service port used by these Compose configurations).
- [x] I made sure font sizes, color choices etc. are all referenced from the theme (not applicable; no UI changes).
- [x] I ran the relevant formatting and validation checks (no server or client source files changed; all affected Compose files pass `docker compose config --quiet`).
- [x] I took a screenshot or a video and attached to this PR if there is a UI change (not applicable; no UI changes).

### DIFF
--- a/docker/docker-compose.yaml
+++ b/docker/docker-compose.yaml
@@ -34,7 +34,7 @@ services:
     volumes:
       - mongo-data:/data/db
     healthcheck:
-      test: ["CMD-SHELL", "bash -c 'exec 3<>/dev/tcp/127.0.0.1/27017'"]
+      test: ["CMD", "bash", "-c", "exec 3<>/dev/tcp/127.0.0.1/27017"]
       interval: 5s
       timeout: 30s
       start_period: 0s
--- a/docker/docker-compose.yaml
+++ b/docker/docker-compose.yaml
@@ -34,12 +34,12 @@ services:
     volumes:
       - mongo-data:/data/db
     healthcheck:
-      test: ["CMD", "mongosh", "--eval", "db.adminCommand('ping')", "--quiet"]
-      interval: 30s
-      timeout: 5s
-      start_period: 30s
-      start_interval: 2s
-      retries: 5
+      test: ["CMD-SHELL", "bash -c 'exec 3<>/dev/tcp/127.0.0.1/27017'"]
+      interval: 5s
+      timeout: 30s
+      start_period: 0s
+      start_interval: 1s
+      retries: 30
 
 volumes:
   mongo-data:
--- a/docker/docker-compose.yaml
+++ b/docker/docker-compose.yaml
@@ -35,11 +35,11 @@ services:
       - mongo-data:/data/db
     healthcheck:
       test: ["CMD", "mongosh", "--eval", "db.adminCommand('ping')", "--quiet"]
-      interval: 5s
-      timeout: 30s
-      start_period: 0s
-      start_interval: 1s
-      retries: 30
+      interval: 30s
+      timeout: 5s
+      start_period: 30s
+      start_interval: 2s
+      retries: 5
 
 volumes:
   mongo-data:
--- a/docker/dev/docker-compose.yaml
+++ b/docker/dev/docker-compose.yaml
@@ -48,8 +48,8 @@ services:
       - ./mongo/data:/data/db
     healthcheck:
       test: ["CMD", "mongosh", "--eval", "db.adminCommand('ping')", "--quiet"]
-      interval: 5s
-      timeout: 30s
-      start_period: 0s
-      start_interval: 1s
-      retries: 30
+      interval: 30s
+      timeout: 5s
+      start_period: 30s
+      start_interval: 2s
+      retries: 5
--- a/docker/dev/docker-compose.yaml
+++ b/docker/dev/docker-compose.yaml
@@ -47,9 +47,9 @@ services:
     volumes:
       - ./mongo/data:/data/db
     healthcheck:
-      test: ["CMD", "mongosh", "--eval", "db.adminCommand('ping')", "--quiet"]
-      interval: 30s
-      timeout: 5s
-      start_period: 30s
-      start_interval: 2s
-      retries: 5
+      test: ["CMD-SHELL", "bash -c 'exec 3<>/dev/tcp/127.0.0.1/27017'"]
+      interval: 5s
+      timeout: 30s
+      start_period: 0s
+      start_interval: 1s
+      retries: 30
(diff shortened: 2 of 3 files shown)


ajhollid (changes requested):
I think a better solution here is to just not use mongosh and just see if 27017 accepts a connection.

There's no replica set here, so if 27017 accepts a connection then the service can considered ready.

In essence, make the health check very cheap instead of carrying out the existing expensive check less often.

Please note there is another docker compose file in the docs/hosting dir that also needs to be updated.

Thanks!

eralpozcan:
I had not considered approaching it this way. Making the health check itself cheap instead of reducing its frequency makes sense. I will revise the change to use a TCP connection check and update the Compose file under docs/hosting as well, then rerun the validation against the revised implementation.

Updated Version: 

TCP Screenshot

docker/docker-compose.yaml
```diff
@@ -34,7 +34,7 @@ services:
     volumes:
       - mongo-data:/data/db
     healthcheck:
-      test: ["CMD", "mongosh", "--eval", "db.adminCommand('ping')", "--quiet"]
+      test: ["CMD-SHELL", "bash -c 'exec 3<>/dev/tcp/127.0.0.1/27017'"]
```
ajhollid: This is fine, but we should be able to skip the extra shell by just doing ` ["CMD", "bash", "-c", "exec 3<>/dev/tcp/127.0.0.1/27017"] `

ajhollid (changes requested):
Check looks good, but there's an extra shell in there that I think we can safely skip

docker/docker-compose.yaml
eralpozcan: Ty! Working on this PR was a great learning experience for catching some subtle nuances. ^^

Issue #2593, opened 2025-07-11
Feature Request: Website monitoring improvements


FirefoxNL:
## Is your feature request related to a problem? Please describe.
The HTTP(S) monitoring options are now quite limited. I would like to suggest some improvements. Checkmate is already an amazing tool but I feel like the following changes would elevate the website monitoring to the next level.
Thank you to all the contributors. 😄

## Describe the solution you'd like
**1. Authentication support** 
Some applications require authentication in the form of an Authorization header. Adding support for authentication in the form of basic authentication or other authentication or token headers would be nice. mTLS or OAuth2 support would be even better.

**2. Custom header support**
This could go hand-in-hand with the authentication support. Some applications require additional headers to be present before giving back a succesful (20x) status.

**3. Other methods**
Some endpoints I would like to monitor only support POST, PUT or DELETE methods.

**4. Retries**
When monitoring public resources, an incidental failed request happens. I would like to be able to set the amount of failures before an incident is created and notification is sent.

**5. Uptime overview**
In the Uptime overview it is now shown as *HTTP(S)*. Showing HTTP or HTTPS there would be a nice bonus.

**6. Custom body**
Adding custom body to request. Could be useful, other monitoring solutions support this as well.

**7. Custom intervals**
Right now, only 1 through 5 minute intervals are available. It would be nice to be able to set more precise intervals. For example by changing it from a drop-down to an input of seconds.

gorkem-bwl:
Those are great suggestions, thank you @FirefoxNL 

@karenvicent @mohicody  we can discuss this during our call this week.

PR #3881, opened 2026-08-26
Add custom monitor interval options

### LINKED ISSUE #2633: Custom Monitor Intervals
## Custom Monitor Intervals

> This issue addresses one of the feature requests mentioned in issue #2593 (point 7: Custom intervals)

Currently, HTTP(S) monitoring intervals are limited to a fixed set of options (1-5 minutes) in a dropdown menu. This limitation prevents users from setting more precise monitoring frequencies for different types of endpoints. Critical services might need more frequent checks (e.g., every 30 seconds), while less critical endpoints could be monitored less frequently to reduce system load.

Replace the current dropdown selection with an input field that allows users to specify custom monitoring intervals in seconds. The implementation would:

- Allow input of custom interval values in seconds
- Include validation for reasonable bounds (e.g., minimum 10 seconds, maximum 1 hour)
- Update the UI to display the input in seconds rather than minutes
- Maintain backward compatibility with existing monitors

fabiangigler:
## Describe your changes

Adds custom monitor intervals across all monitor types using dynamic autocomplete suggestions for seconds, minutes, hours, and days to the UI. The API + monitoring tasks already support arbitrary intervals.

autocomplete-interval-1

Fixes #2633
Also Custom Interval portion of #2593

```Javascript
<div>{t('add')}</div>
```

---


client/src/Pages/CreateMonitor/index.tsx
```diff
@@ -454,16 +491,16 @@ const CreateMonitorPage = () => {
...
+	const intervalOptions = useMemo(() => {
+		const options = MonitorIntervalOptions.map(({ value }) => toIntervalOption(value));
+		if (
+			typeof watchedInterval === "number" &&
+			!options.some((option) => option.id === watchedInterval)
+		) {
+			options.push(toIntervalOption(watchedInterval));
+		}
+		return options;
+	}, [watchedInterval]);
```
ajhollid, line 503: This is a localization regression.  These values will only display in English, whereas they were localized before.  We'll need to implement proper localization support here

client/src/Pages/CreateMonitor/index.tsx
```diff
@@ -689,12 +726,16 @@ const CreateMonitorPage = () => {
...
+							<FormAutocompleteField
 								name="interval"
 								fieldLabel={t(
 									"pages.createMonitor.form.frequency.option.frequency.label"
 								)}
 								options={intervalOptions}
+								filterOptions={(options, { inputValue }) =>
+									filterIntervalOptions(options, inputValue)
+								}
+								getOptionDisabled={(option) => option.id < MIN_MONITOR_INTERVAL_MS}
```
ajhollid, line 738: Clearing the field produces an error that isn't helpful for the user:

Image

This needs to be addressed, probably by ignoring the clear in `onChange`.  `disableClearable` should also probably be set here, which necessitates adding the prop to `FormAutoCompleteField` and threading it through to the `Autocomplete` component

server/src/api/validation/monitorValidation.ts
```diff
@@ -9,12 +9,14 @@ import {
...
 	MonitorStatuses,
 	MonitorTypes,
+	MIN_MONITOR_INTERVAL_MS,
 	PageSpeedStrategies,
 	ProxyModes,
 } from "@/domain/monitors/monitor.type.js";
 import { DateRanges, SortOrders } from "@/types/query.js";
 
 const httpStatusCode = z.number().refine((code) => HttpStatusCodeSet.has(code), { message: "Must be a valid HTTP status code" });
+const monitorIntervalSchema = z.number().int().min(MIN_MONITOR_INTERVAL_MS);
```
ajhollid: Monitors created via API could previously be created with any value for interval.  If a user exports those monitors and expects to be able to import them again, their import operation will fail.

I think the most reasonable approach would be to clamp minimum values in the validation schema.  You can probably use Zod's [transform API](https://zod.dev/api#transform) to handle this

ajhollid (changes requested):
This is a nice addition, but there are some issues to work through before merging, please see my comments in the code review.

client/src/Pages/CreateMonitor/index.tsx
fabiangigler: @ajhollid I agree this is a regression, however the function wraps the formatDuration helper which is used to display the interval durations i.e. on the monitor detail page, where the duration + helper copy is still hardcoded in English as well.

I'd be happy to tackle i18n on the formatDuration / underlying time formatting helpers, but thought that should be part of a seperate issue / PR, right?

server/src/api/validation/monitorValidation.ts
fabiangigler: That's essentially my motivation for this PR, I migrated 50 monitors from uptime robot to checkmate, and found all my durations be working, just not editable with the same flexibility in the UI 😅

Now that I think about it, maybe it's best to just limit the frontend to a minimum interval (is 10seconds ok instead of 15?) and leave the API validation as is.

client/src/Pages/CreateMonitor/index.tsx
fabiangigler: @ajhollid I created an issue here:
https://github.com/bluewave-labs/Checkmate/issues/3891

client/src/Pages/CreateMonitor/index.tsx
fabiangigler: fixed via follow up commit!

fabiangigler:
@ajhollid addressed the feedback + opened a seperate issue regarding the global duration i18n

server/src/api/validation/monitorValidation.ts
ajhollid: @fabiangigler I think you were on the right track here, frontend validation and backend validation should generally be in sync.  We shouldn't be able to craft monitors via API or JSON import that don't match what should be allowed via UI.

Unless there's a compelling reason not to, I think that clamping to the validation enforced minimum is a reasonable strategy rather than just flat out rejecting :thinking:   Thoughts?

fabiangigler: @ajhollid mainly I did not want to silently break existing functionality (i.e. creating 3s monitors via the API, would not throw any errors, but just fetch at the minimum clamp value instead of what the API call requested), especially as people technical enough to use the API might know the drawbacks of super fast monitoring intervals.

This is really your judgement call to make here, I don't personally have a strong need to monitor in less than 15s intervals, so happy to tighten the validation here if that's inline with the project.

gorkem-bwl:
Loved the dynamic autocomplete suggestions!

client/src/Pages/CreateMonitor/index.tsx
fabiangigler: opened a PR tof ix this here:
https://github.com/bluewave-labs/Checkmate/pull/3913

will need to update L83 to enable localized "long" formatting after it's merged.

fabiangigler:
@gorkem-bwl anything missing to go ahead with approving/merging this one?
just rebased against latest develop to resolve conflicts.

ajhollid:
Hey @fabiangigler,

Apologies, the review backlog is pretty deep! 

I will take a look at this PR first thing tomorrow, thanks for your patience!

ajhollid (approved):
Looks good here, thank you for your contribution!

PR #3890, opened 2026-08-26
Email auth not required

### LINKED ISSUE #3866: Email auth required but should not be
**Describe the bug**
the UI and Sent test requires that a password is entered, but Google Workspace uses IP whitelists instead of email auth, and does not allow smtp auth anymore.  I cannot configure email notification at this time.

**Expected behavior**
auth and password should be optional.

shlokmestry:
## Describe your changes

The email settings form required an SMTP password before a test email could be sent or the "Send Test Email" button enabled, even though the backend already treats `systemEmailPassword` as optional. Providers like Google Workspace that use IP allowlisting instead of SMTP auth have no password to enter, so the form blocked a valid configuration.

- Removed the password requirement from the pre-flight check in `handleSendTestEmail` and from the "Send Test Email" button's `disabled` condition — only host, port, and address remain required.
- Fixed a regression this introduced: when a password is already saved server-side, the (hidden) password field defaults to `""`, so the test-email request would silently send an empty password and could break a working authenticated config. `systemEmailPassword` is now only included in the payload when the user actually enters one, matching the existing pattern used for the other optional email fields.

Fixes #3866

- [x] I have added i18n support to visible strings (instead of `<div>Add</div>`, use): n/a — no new UI strings added
- [ ] I ran `npm run format` in server and client directories, which automatically formats your code.
- [ ] I took a screenshot or a video and attached to this PR if there is a UI change.****

### DIFF
--- a/client/src/Pages/Settings/index.tsx
+++ b/client/src/Pages/Settings/index.tsx
@@ -197,12 +197,14 @@ export const SettingsPage = () => {
 			systemEmailHost: formValues.systemEmailHost,
 			systemEmailPort: formValues.systemEmailPort,
 			systemEmailAddress: formValues.systemEmailAddress,
-			systemEmailPassword: formValues.systemEmailPassword,
 			systemEmailSecure: formValues.systemEmailSecure,
 			systemEmailPool: formValues.systemEmailPool,
 			systemEmailIgnoreTLS: formValues.systemEmailIgnoreTLS,
 			systemEmailRequireTLS: formValues.systemEmailRequireTLS,
 			systemEmailRejectUnauthorized: formValues.systemEmailRejectUnauthorized,
+			...(formValues.systemEmailPassword && {
+				systemEmailPassword: formValues.systemEmailPassword,
+			}),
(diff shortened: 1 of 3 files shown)


sanmaxdev (changes requested):
Blocking: this removes the client-side password requirement, but `EmailService` still always passes an `auth` object with `user: systemEmailUser || systemEmailAddress` and the optional password. Nodemailer treats that shape as `LOGIN` even when the password is absent, so an IP-allowlisted relay still attempts SMTP authentication.

Please omit `auth` from the transport configuration when no credentials are configured, and add a unit test for the no-auth transport shape in `emailService.test.ts`.

shlokmestry:
@sanmaxdev Thanks for the catch - fixed in 46a822d. `auth` is now only attached to the nodemailer transport config when `systemEmailPassword` is actually set, so an IP-allowlisted relay (no password configured) no longer triggers LOGIN auth. Also added unit tests covering both the no-auth and with-auth transport shapes in `emailService.test.ts`. Ready for another look.

PR #3910, opened 2026-09-02
fix(client): don't navigate away when selecting text in a table row

### LINKED ISSUE #3907: Failed health check: Error message is difficult to copy
When clicking on a health monitor, a list of entries showing "status | date & time | message | status code" is displayed. However, clicking on a single entry causes all this information to disappear in the detail view.

This is particularly frustrating when a health check fails and you want to copy the error message to your clipboard. Although you can highlight the text, releasing the mouse button immediately navigates to the next view, which no longer contains the message.

Here are two potential solutions:
1. Add a "Copy to Clipboard" button/icon directly next to the message.
2. Preserve and display all previously shown information in the detailed view as well.

I talk about this view: 

Image

gorkem-bwl:
Closes #3907.

## Problem

Selecting the text of a failed check's error message navigates away before you can copy it. The mouseup that ends a drag-select lands on the table row, which fires the row's `onClick`.

## Change

`DataTable` ignores a row click when the text selection is inside the row that was clicked. One file, +25/−2.

The selection is scoped to the clicked row deliberately: a document-wide check would let a leftover selection anywhere on the page swallow the first click on any row.

Because the guard lives in the shared table, this fixes drag-select on all 13 tables that pass `onRowClick` — incidents, monitors, notifications, status pages and the rest — not just checks. Both the desktop row and the mobile card layout are covered.

## What this replaces

An earlier revision added a copy button on the message column. It worked, but it was ~140 lines for a second answer to a problem the guard already solves, and it brought a shared component, an HTTP clipboard fallback, four translation keys and a timer along with it. Dropped in favour of the smaller fix. A reusable copy button is still worth having for monitor URLs, tokens and incident ids — it should just land on its own merits rather than inside a bug fix.

## A related bug, deliberately not fixed here

The row click on the uptime details checks table navigates to `/checks/<check id>`, but that route's parameter is `:monitorId` (`client/src/Routes/index.tsx`). A check id never matches a monitor, so `selectedMonitorType` is `undefined`, the fetch URL resolves to `null`, and the destination page renders empty.

That is the literal "all this information disappears" in the report. It needs its own decision about where that row should link — the monitor, a real check detail view, or nowhere — so it is out of scope here. Happy to follow up.

## Testing

- `npx tsc --noEmit` — clean
- `npm run build` — passes
- `npm run lint` — 12 errors / 303 warnings, byte-identical to `develop` (verified by stashing); no new problems
- `npm run format-check` — passes

Worth checking by hand: drag-select an error message and confirm the page stays put; normal row clicks still navigate everywhere else; a selection made outside the table does not block the next row click; and the mobile card layout, which shares the guard.


ajhollid (approved):
Looks good

ajhollid (approved):

PR #3913, opened 2026-09-02
Localize formatDuration output

### LINKED ISSUE #3891: i18n: Add localization to the duration helpers
# Add localization to the duration helpers

## The issue

Currently the `formatDuration` function has an argument that expands single-letter durations (i.e. `5s`) into full-word durations (i.e. `5 seconds`). The underlying library used explicitly doesn't support localization ( https://github.com/sindresorhus/pretty-ms/pull/35 ) and has the values for these strings hardcoded, for the benefit of being a really tiny library that just does what it says on the box.

That argument is really only used on the MonitorStatus detail page, where both the label as well as the duration are hardcoded to english:
```tsx
{!isSmall && (
	<>
		<Dot />
		<Typography>
			Checking every {formatDuration(monitor?.interval, true)}
		</Typography>
	</>
)}
```

Having durations be internationalized wherever they are formatted in the longer format would also be beneficial to the UI in other places, where these were manually translated for now, but also my PR for a dynamic interval UI here: https://github.com/bluewave-labs/Checkmate/pull/3881

## Possible fix

I'd be happy to implement a fix, but want to align on the solution first.

Imho the best option is just moving from pretty-ms to a more complete datetime library like dayjs for example (still really tiny at 2kb minified) that has support for loading additional locales out of the box. Changes would be limited to the packages + `client/src/Utils/TimeUtils.ts`

see the following for reference:
https://day.js.org/
https://day.js.org/docs/en/durations/humanize
https://day.js.org/docs/en/i18n/instance-locale

fabiangigler:
Closes #3891

## Summary

- Add options for localized long and relative `formatDuration` output.
- Use `Intl.NumberFormat` and `Intl.RelativeTimeFormat`.
- Apply the localized formatter to monitor intervals and last-check times.

Day.js uses fuzzy relative durations, formatting values such as 8 seconds as “a few seconds,” and its locales generally do not define numeric seconds. The Intl APIs provide localized numeric units and relative wording and are supported by current browsers and Node.js.

## Before and After

Note that the screenshots show two additional differences besides just the localization:
- "Active for" times have been shortened in the current dev branch unrelated to the changes in this PR
- This PR introduces both a plain `long` (just showing the number and long form unit word) and a relative option (showing the number, unit word + "ago") however, both of them round down to the highest unit (only minutes instead of minutes and seconds) which is the only way to get the nicer relative localizations out of the box.

Also domains are blurred just for the screenshots.

### English
grafik
grafik

### German
grafik
grafik

## Additionally

- We could also show a relative time for the certificate (and domain) expiry.
- Show the long, nicer form for the "Active for" duration.

### DIFF
--- a/client/src/Components/design-elements/MonitorStatus.tsx
+++ b/client/src/Components/design-elements/MonitorStatus.tsx
@@ -8,8 +8,11 @@ import { typographyLevels } from "@/Utils/Theme/Palette";
 import useMediaQuery from "@mui/material/useMediaQuery";
 import { LAYOUT } from "@/Utils/Theme/constants";
 import { formatDuration } from "@/Utils/TimeUtils";
+import { useTranslation } from "react-i18next";
+
 export const MonitorStatus = ({ monitor }: { monitor: Monitor }) => {
 	const theme = useTheme();
+	const { t, i18n } = useTranslation();
 	const isSmall = useMediaQuery(theme.breakpoints.down("md"));
 
 	if (!monitor) {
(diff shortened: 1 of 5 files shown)


ajhollid (approved):
Thnaks for this PR!  I pushed two additional commits to handle formatting and pass required values from the call sites after merging your other PR.   Thank you for your contribution here!

PR #3915, opened 2026-09-03
feat(client): define the neutral palette, switch to Geist, and standardize controls

gorkem-bwl:
Design pass over the client theme. No component was restructured — this is almost entirely tokens, plus the handful of component fixes those tokens exposed.

## Why

The palette declared a brand green and six chart hues but never defined the neutrals, so MUI's defaults filled the gap:

- `text.secondary` was `rgba(0,0,0,0.6)` — and every table cell renders through it, which is most of the "washed out" feeling
- `background.default` equalled `background.paper`, so cards and tables had no ground to sit on
- `divider` at 12% black made table rules barely visible
- `success`/`error`/`warning` were stock Material red/green/orange, unrelated to the brand
- Only two font weights existed (h1 500, everything else 400), so h2 was the same weight as body text

## What changed

**Palette.** An opaque neutral ramp biased slightly green, an off-white `background.default` so paper separates from the page, and semantic colours derived from the brand. Dark mode got the same treatment — it defined `background` and `secondary` but its text and divider were still MUI defaults. Every pair passes WCAG AA in both themes; the weakest is 4.63:1.

**Typography.** Geist Variable, self-hosted via `@fontsource-variable/geist` so air-gapped installs keep the typeface. Base size stays 13px. `h1` 700 / `h2` 600 against 400 body, and table headers 500 → 600.

**One control height.** Buttons, inputs, selects, pickers, toggles and tabs declared 34 while IconButton and the PageHeader action sat at 32, so mixed rows were 2px out of alignment. All now read `CONTROL_HEIGHT` from `constants.ts`. Two content-box fixes were needed to make 32 safe, both measured in the browser: MUI's button line-height (1.75 → 22.75px at our base) already overflowed its own box at 34, and MUI's 8.5px inner input padding meant an input's content needed 35.7px inside a 32px shell. Verified in the DOM that button, select and input all render at exactly 32 with `scrollHeight === height`.

**One border story for inputs.** The theme hovered to near-black while Select, TextInput and DatePicker each overrode hover back to `divider` — the resting colour — so those had no hover cue at all. Now defined once and inherited: rest `divider`, hover `text.disabled`, focus brand at 2px, error `error.main`. The per-component overrides are gone.

**Icon sizing.** `Icon` defaulted to 20px, but 17 call sites asked for 16 explicitly against 2 for 20, and the sidebar clamped to 16 in CSS — so 16 was the house size and 20 was a default nobody chose. Anything bare rendered above the cap height of the 13px text beside it. Default is now 16, with the public status page's warning icon pinned at 20 so a published page is unaffected.

**Row selection.** `getRowSx` set a background on the `<tr>`, but every `<td>` paints `background.paper` over it, and the hover rule targets cells — so hover won and selection was invisible. Selection and hover are now both cell-level rules in `DataTable`, with selection first. Uptime, Docker and Infrastructure each had their own broken copy; they now share one implementation via a new `isRowSelected` prop.

## Things worth a reviewer's eye

- **Chart colours shifted.** MUI derives `.light` from `.main`, so histograms, heatmaps and PageSpeed pies moved from Material's bright tones (`#4caf50`, `#ef5350`, `#ff9800`) to brand-derived ones. Intentional, but it touches 8 chart files that no commit message names directly.
- **The offline banner now uses `error.dark`.** In dark mode `error.main` is light enough that MUI derives near-black `contrastText`, which read as a warning rather than an outage. Darkening `error.main` itself was the alternative, but ~30 places use it as text or a chart mark on a dark ground where the lighter value is correct.
- **Dark-mode focus rings use `primary.light`.** The brand green is only 2.86:1 on the dark ground, which made a focused input dimmer than a hovered one. `primary.main` is deliberately untouched — it is the fill for contained buttons, and lightening it would flip their labels to dark text across 31 sites.
- **`isRowSelected` is desktop-only.** The small-screen card branch does not style selection, the same limitation `getRowSx` already had. Documented on the prop rather than fixed here.

## Testing

`npx tsc --noEmit`, `npm run build` and `npm run format-check` all pass. `npm run lint` reports 12 errors / 308 warnings, identical to `develop` — all in untouched files (`Validation/patterns.ts`, `Validation/validation.js`, `vite.config.ts`).

Two rounds of automated review were run against the branch and their findings applied, including a real regression this introduced: once the sidebar moved onto `paper`, the selected nav item composited darker than its ground while hover came out lighter, so hovering any other item made it read as more selected than the actual selection.

Worth exercising by hand: Settings and Create Monitor (densest forms — the input padding change touches every field), dark mode generally, and the public status pages, whose font stack lost its dead `"Inter"` entry.


client/src/Components/design-elements/OfflineBanner.tsx
```diff
@@ -37,8 +37,14 @@ export const OfflineBanner = ({ visible }: OfflineBannerProps) => {
 				left: 0,
 				right: 0,
 				zIndex: theme.zIndex.snackbar,
-				backgroundColor: theme.palette.error.main,
-				color: theme.palette.error.contrastText,
+				// `error.dark` rather than `main`: in dark mode `main` is light
```
malenacaroline: Can you remove these comments? Comments are helpful for explaining complex logic, but most UI changes don’t need them if the code is clear enough. Too many comments can make the code less readable.

client/src/Components/design-elements/Icon.tsx
```diff
@@ -8,7 +8,10 @@ interface IconProps {
 	color?: string;
 }
 
-const Icon = ({ icon: Icon, size = 20, strokeWidth = 1.5, color }: IconProps) => {
+// 16 is the house size: 17 call sites ask for it explicitly against 2 for 20,
+// and the sidebar clamps to 16 in CSS. A 20px default put icons above the cap
```
malenacaroline: Let’s try to find a better place to document these UI decisions. The code itself isn’t the best place for them.

client/src/Components/design-elements/PageHeader.tsx
```diff
@@ -142,7 +146,7 @@ export const PageHeader = ({
 								alignItems: "center",
 								justifyContent: "center",
 								width: 36,
-								height: 32,
+								height: CONTROL_HEIGHT,
```
malenacaroline: What does `CONTROL_HEIGHT` mean? Could we use a more specific name, like `BASE_HEIGHT` or something similar?

malenacaroline (changes requested):

client/src/Components/design-elements/PageHeader.tsx
ajhollid: It's the base height for all control type elements, like buttons, tabs etc.  Maybe `INPUT_BASE_HEIGHT`?

client/src/Components/design-elements/Table.tsx
```diff
@@ -241,7 +259,12 @@ export function DataTable<
 						return (
 							<Fragment key={key}>
 								<TableRow
-									className={isInteractive ? "is-clickable" : undefined}
+									className={[
+										isInteractive ? "is-clickable" : "",
+										isRowSelected?.(row) ? "is-selected" : "",
+									]
+										.filter(Boolean)
+										.join(" ")}
```
ajhollid: This works, but it is not intuitive, and we don't use classnames anywhere else in the application, so a better approach would probably be to use the `sx` props as usual

ajhollid (approved):

PR #3925, opened 2026-09-07
fix: return the created monitor from createMonitor

### LINKED ISSUE #3917: createMonitor never returns the created document — every successful POST /monitors answers data: null
`MonitorService.createMonitor` is declared `Promise<void>`:

```ts
// server/src/domain/monitors/monitor.service.ts
createMonitor = async (teamId: string, userId: string, body: Monitor): Promise<void> => {
  if (body.proxyMode !== "custom") {
    delete body.proxyId;
  }
  const monitor = await this.monitorsRepository.create(body, teamId, userId);
  if (!monitor) {
    throw new AppError({ message: "Failed to create monitor", status: 500, service: SERVICE_NAME, method: "createMonitor" });
  }

  this.scheduler.addJob(monitor.id, monitor);
  // no return
};
```

and the controller writes the return value straight into the response body regardless:

```ts
// server/src/api/controllers/monitorController.ts
createMonitor = catchAsync(async (req: Request, res: Response) => {
  const validatedBody = createMonitorBodyValidation.parse(req.body);
  const userId = requireUserId(req.user?.id);
  const teamId = requireTeamId(req.user?.teamId);

  const monitor = await this.monitorService.createMonitor(teamId, userId, validatedBody);

  return res.status(200).json({
    success: true,
    msg: "Monitor created successfully",
    data: monitor,
  });
});
```

`monitor` is `undefined` at that line on every call that reaches it — the service function never returns anything, successful or not. So `POST /monitors` answers `{"success": true, "msg": "Monitor created successfully", "data": null}` for every monitor it creates, and a caller has no way to learn the new monitor's id from the create response.

## Reproduced against a live instance

Registered a monitor via `POST /monitors` with a valid body; response carried `data: null`. `GET /monitors/team` immediately after shows the monitor exists, fully configured, with a real `id` — so `monitorsRepository.create` and `scheduler.addJob` both ran correctly. Only the id never left the create call.

## Fix

`createMonitor` should return what it creates:

```diff
- createMonitor = async (teamId: string, userId: string, body: Monitor): Promise<void> => {
+ createMonitor = async (teamId: string, userId: string, body: Monitor): Promise<Monitor> => {
    if (body.proxyMode !== "custom") {
      delete body.proxyId;
    }
    const monitor = await this.monitorsRepository.create(body, teamId, userId);
    if (!monitor) {
      throw new AppError({ message: "Failed to create monitor", status: 500, service: SERVICE_NAME, method: "createMonitor" });
    }

    this.scheduler.addJob(monitor.id, monitor);
+   return monitor;
  };
```

and the corresponding line in `IMonitorService`. No controller change needed — it already forwards whatever the service returns.

## Impact

Any API client that needs the created monitor's id right away — to store it, to reference it in a follow-up call, to match it against a local record — cannot get it from the create response and has to fall back to listing monitors and matching by name. That is a workaround for a bug, not a documented API contract, and it costs an extra round trip on every registration.

egeoztass:
## Describe your changes

`createMonitor` was typed `Promise<void>` and never returned the document, so the controller sent `data: null` on every successful `POST /monitors`. It now returns the created monitor and the interface type matches. The existing unit test asserts the return value.

Fixes #3917

### DIFF
--- a/server/src/domain/monitors/monitor.service.ts
+++ b/server/src/domain/monitors/monitor.service.ts
@@ -60,7 +60,7 @@ const computeRestartsInRange = (aggregate: DockerContainerStatsBucket[]): number
 
 export interface IMonitorService {
 	// create
-	createMonitor(teamId: string, userId: string, body: Partial<Monitor>): Promise<void>;
+	createMonitor(teamId: string, userId: string, body: Partial<Monitor>): Promise<Monitor>;
 	createMonitors(monitors: Array<Monitor>): Promise<Monitor[] | null>;
 	addDemoMonitors(args: { userId: string; teamId: string }): Promise<Monitor[]>;
 
@@ -180,7 +180,7 @@ export class MonitorService implements IMonitorService {
 		this.incidentsRepository = incidentsRepository;
 	}
 
-	createMonitor = async (teamId: string, userId: string, body: Monitor): Promise<void> => {
+	createMonitor = async (teamId: string, userId: string, body: Monitor): Promise<Monitor> => {
 		// proxyId is only needed in custom mode
 		if (body.proxyMode !== "custom") {
 			delete body.proxyId;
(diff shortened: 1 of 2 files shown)


ajhollid (approved):
Thanks for the clean PR!

PR #3929, opened 2026-09-07
fix: return ntfy auth fields from the notification repository

### LINKED ISSUE #3928: ntfy channels with authentication send anonymously
The notification repository mapper never returns `ntfyAuthType` or `ntfyUsername`. Alerts are sent with what the mapper returns, so `NtfyProvider` sees no auth type and posts without an Authorization header, even for channels saved with token or basic auth. The same fields are also missing from GET responses.

Introduced by #3839 (my PR, the mapper was missed). Fix is two lines plus a repository test; PR coming right after this.

egeoztass:
## Describe your changes

`toEntity` in the notification repository never returned `ntfyAuthType` or `ntfyUsername`. Since alerts are sent with whatever the repository returns, the ntfy provider treated every channel as unauthenticated and posted anonymously. This adds the two fields to the mapper and a mongodb-memory-server test that reads a stored ntfy channel back through the real repository, which fails on develop and passes here.

Regression from #3839, sorry about that.

Fixes #3928

### DIFF
--- a/server/test/integration/notificationRepository.test.ts
+++ b/server/test/integration/notificationRepository.test.ts
@@ -1,0 +1,74 @@
+import { describe, expect, it, beforeAll, afterAll, beforeEach } from "@jest/globals";
+import mongoose from "mongoose";
+import { MongoMemoryServer } from "mongodb-memory-server";
+import MongoNotificationsRepository from "../../src/domain/notifications/notification.repository.mongo.ts";
+import { NotificationModel } from "../../src/domain/notifications/notification.model.ts";
+
+// ── Real-Mongo harness ─────────────────────────────────────────────────────────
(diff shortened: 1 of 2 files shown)


server/test/integration/notificationRepository.test.ts
```diff
@@ -0,0 +1,74 @@
+import { describe, expect, it, beforeAll, afterAll, beforeEach } from "@jest/globals";
+import mongoose from "mongoose";
+import { MongoMemoryServer } from "mongodb-memory-server";
+import MongoNotificationsRepository from "../../src/domain/notifications/notification.repository.mongo.ts";
```
ajhollid, line 4: Same here, mock DB response should be sufficient

ajhollid (changes requested):

PR #3931, opened 2026-09-07
fix: return systemEmailDisplayName from the settings repository

### LINKED ISSUE #3930: Alert emails ignore the configured sender display name
The settings repository mapper (`toEntity` in `app-settings.repository.mongo.ts`) never returns `systemEmailDisplayName`. `EmailService.sendEmail` reads its transport config from `settingsService.getDBSettings()`, which goes through that mapper, so every real alert email is sent from the bare address. The test-email button passes the form values straight through, so it does show the display name, which makes this easy to miss.

Looks like it was lost in the domain reorg (d022f7d84). One-line fix plus a repository test; PR coming right after this.

egeoztass:
## Describe your changes

Adds `systemEmailDisplayName` to the settings repository mapper so alert emails carry the configured sender name, matching what the test-email path already does. Includes a mongodb-memory-server test that reads settings back through the real repository; it fails on develop and passes here.

Fixes #3930

### DIFF
--- a/server/test/integration/appSettingsRepository.test.ts
+++ b/server/test/integration/appSettingsRepository.test.ts
@@ -1,0 +1,46 @@
+import { describe, expect, it, beforeAll, afterAll, beforeEach } from "@jest/globals";
+import mongoose from "mongoose";
+import { MongoMemoryServer } from "mongodb-memory-server";
+import MongoAppSettingsRepository from "../../src/domain/app-settings/app-settings.repository.mongo.ts";
+import { AppSettingsModel } from "../../src/domain/app-settings/app-settings.model.ts";
+
+// ── Real-Mongo harness ─────────────────────────────────────────────────────────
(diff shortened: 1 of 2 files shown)


server/test/integration/appSettingsRepository.test.ts
```diff
@@ -0,0 +1,46 @@
+import { describe, expect, it, beforeAll, afterAll, beforeEach } from "@jest/globals";
+import mongoose from "mongoose";
+import { MongoMemoryServer } from "mongodb-memory-server";
+import MongoAppSettingsRepository from "../../src/domain/app-settings/app-settings.repository.mongo.ts";
```
ajhollid, line 4: I don't think there's any need to test the actual DB here, it's sufficient to mock the DB response.  We shouldn't have to spin up the database to run tests

ajhollid (changes requested):

PR #3935, opened 2026-09-08
fix(client): credit Annie Hsu on the 404 illustration

gorkem-bwl:
The sushi artwork on the 404 page was designed by **Annie Hsu** and carried no attribution. This adds her name to the bottom-right of the illustration, in a script face at a size that reads as a signature rather than a label.

## Why outlines rather than a font

The glyphs are embedded as vector paths, not a `<text>` element with a `font-family`.

The asset is loaded through an `` tag (`client/src/Pages/NotFound/index.tsx:1`), and an ``-loaded SVG cannot fetch external resources — a web font would silently fall back to whatever cursive face the viewer happens to have, which differs on every platform. Outlines render identically everywhere, and stay sharp at any size in a way a raster export would not.

Cost is ~21KB on the asset.

## Diff

One file. The first commit adds the signature (4 insertions, 1 deletion — the deletion is the closing `</svg>` moving down a line); the second changes one character of the fill colour.

## Testing

Rendered and checked at the page's own `max-height: 25rem` on **both** theme backgrounds. The signature sits clear of the artwork, and the illustration is otherwise untouched — no existing paths were modified.

## Colour choice

The app swaps the whole tree between `lightTheme` and `darkTheme` at `client/src/App.tsx:25`, and the 404 page renders under both. The signature colour has to work on each:

| | on dark `#151518` | on light `#F6F8F7` |
|---|---|---|
| `#9a9a9a` (first pass) | 6.48:1 | **2.64:1** — washed out |
| `#787878` (shipped) | 4.13:1 | 4.14:1 |

`#787878` was chosen by maximising the *worse* of the two contrasts rather than tuning for either ground, so it reads as a signature in both modes without becoming heavy enough to compete with the artwork.


ajhollid (approved):
🚀   Thank you @gorkem-bwl !

PR #3942, opened 2026-09-11
Feat/status page embed origins

### LINKED ISSUE #3941: Allow status pages to be embedded in an iframe from configured origins

Public status pages are served with `Content-Security-Policy: frame-ancestors 'self'` (helmet's default). Embedding a status page in an iframe from another origin, for example an internal operations dashboard, is blocked by the browser. Serving the page on a custom domain does not help, because the embedding page is still a different origin. We currently run a local patch that hard-codes our origins into the global helmet `frame-ancestors` list, which is not something we can contribute as it stands.

An optional list of allowed embedding origins on each status page, set in the status page form alongside the custom domain. The existing status-page document CSP middleware would look the page up (by `/status/public/:url` or by custom domain, cached in the same way as the CORS origin middleware) and emit `frame-ancestors 'self' <origins>` for that document only. Every other document keeps `frame-ancestors 'self'`. Origins would be validated as `https?://host[:port]` with no path, and only honoured for published pages.

An instance-wide environment variable listing allowed origins. Simpler, but it applies to every status page on the instance. The per-page field fits better with the existing custom domain feature.

**Additional context**

Browsers enforce the intersection of all CSP headers, so a second header cannot widen the helmet default. The change would remove `frame-ancestors` from the global helmet policy and emit it per request from the status-page CSP middleware instead. Open question: whether `X-Frame-Options` should also be dropped for status page documents. Browsers that support `frame-ancestors` ignore it when both are present.

Happy to take this on and open a PR against `develop` if it is acceptable.

web-engineer:
## Describe your changes

Public status pages were served with helmet's default `frame-ancestors 'self'`, so embedding one in an iframe from another origin was blocked, and a custom domain did not help because the embedding page is still a different origin.

Each status page now has an optional list of allowed embedding origins, set in the status page form next to the custom domain. Origins are validated as `https?://host[:port]` with no path, limited to 20, and normalised (trimmed, lowercased, deduplicated) before saving. The status page document CSP middleware looks the page up by `/status/public/:url` or by custom domain, cached in the same way as the CORS origin middleware, and emits `frame-ancestors 'self' <origins>` for published pages. Every other non-API document gets `frame-ancestors 'self'`. helmet no longer emits `frame-ancestors`.

One ordering fix on the way: helmet sets the Content-Security-Policy header with a replace, not an append, so the existing status page CSP tightening was being lost on documents served by the SPA catch-all, and documents served straight from the client build carried no helmet policy at all. helmet now runs before `express.static` and before the status page middleware. Verified with curl before and after.

`X-Frame-Options: SAMEORIGIN` is still sent. Browsers that support `frame-ancestors` ignore it when both are present. Happy to drop it for status page documents if preferred.

Tested locally: server lint, format-check, build and unit tests (Node 20); client lint, format-check and build; a published page with two origins checked with curl on the public path and on a custom domain host, plus unknown page, unpublished page, app root and API responses.

Happy to split this into a server PR and a client PR if that is easier to review.

Fixes #3941

```Javascript
<div>{t('add')}</div>
```

Screen shot showing the new addition to the status page area - /status/configure/<PAGEID>

Screenshot 2026-09-11 at 12 32 44

### DIFF
--- a/client/src/Validation/statusPage.ts
+++ b/client/src/Validation/statusPage.ts
@@ -15,11 +15,22 @@ const statusPageHostnamePattern =
 
 const statusPageHostnameRegex = new RegExp(`^(?=.{1,253}$)${statusPageHostnamePattern}$`);
 
-// An embedding origin: scheme + hostname (or localhost) + optional port, no path.
-// Kept in sync with embedOriginRegex in server/src/api/validation/shared.ts.
-const statusPageOriginRegex = new RegExp(
-	`^https?:\\/\\/(?:localhost|${statusPageHostnamePattern})(:\\d{1,5})?$`
-);
+// An embedding origin: scheme + host + optional port, no path. The host may be
+// a hostname, localhost, an IPv4 address, or a bracketed IPv6 address. Kept in
+// sync with isEmbedOrigin in server/src/api/validation/shared.ts.
+const statusPageOriginShapeRegex =
+	/^https?:\/\/([^\s/?#:[\]]+|\[[^\s/?#[\]]+\])(?::\d{1,5})?$/;
+
+const isStatusPageOrigin = (value: string): boolean => {
+	const host = statusPageOriginShapeRegex.exec(value)?.[1];
+	if (host === undefined) return false;
+	if (host.startsWith("[")) return z.ipv6().safeParse(host.slice(1, -1)).success;
+	return (
+		host === "localhost" ||
+		z.ipv4().safeParse(host).success ||
+		statusPageHostnameRegex.test(host)
+	);
+};
 
 const MAX_EMBED_ALLOWED_ORIGINS = 20;
 
@@ -69,16 +80,10 @@ export const statusPageSchema = z.object({
 		.refine((raw) => splitEmbedAllowedOrigins(raw).length <= MAX_EMBED_ALLOWED_ORIGINS, {
 			message: `At most ${MAX_EMBED_ALLOWED_ORIGINS} embedding origins are allowed`,
 		})
-		.refine(
-			(raw) =>
-				splitEmbedAllowedOrigins(raw).every((origin) =>
-					statusPageOriginRegex.test(origin)
-				),
-			{
-				message:
-					"Each origin must be a scheme and host with no path (e.g. https://dashboard.example.com)",
-			}
-		),
+		.refine((raw) => splitEmbedAllowedOrigins(raw).every(isStatusPageOrigin), {
+			message:
+				"Each origin must be a scheme and host with no path (e.g. https://dashboard.example.com)",
+		}),
 	timezone: z.string().optional(),
 	type: z
 		.array(z.enum(["uptime", "infrastructure"]))
--- a/client/src/Validation/statusPage.ts
+++ b/client/src/Validation/statusPage.ts
@@ -10,8 +10,25 @@ import { cssReferencesExternalResource } from "@/Utils/customCss";
 // field definitions.
 export const statusPageStepRegistry = z.registry<{ step: number }>();
 
-const statusPageHostnameRegex =
-	/^(?=.{1,253}$)([a-zA-Z0-9_](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?\.)+[a-zA-Z]{2,63}$/;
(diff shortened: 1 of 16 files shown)


server/src/api/validation/shared.ts
```diff
@@ -44,7 +44,12 @@ export const timezoneValidation = z
 // optionally prefixed with `_` for service labels (e.g. _dmarc, _imaps._tcp).
 // No scheme, port, path, or whitespace. Total length ≤ 253. Kept in sync with
 // the client-side regex in client/src/Validation/monitor.ts.
-export const dnsHostnameRegex = /^(?=.{1,253}$)([a-zA-Z0-9_](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?\.)+[a-zA-Z]{2,63}$/;
+const dnsHostnamePattern = "([a-zA-Z0-9_](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?\\.)+[a-zA-Z]{2,63}";
+export const dnsHostnameRegex = new RegExp(`^(?=.{1,253}$)${dnsHostnamePattern}$`);
+
+// Web origin for CSP frame-ancestors: http(s) scheme, a hostname (or localhost),
+// an optional port, and nothing else — no path, query, or trailing slash.
+export const embedOriginRegex = new RegExp(`^https?://(?:localhost|${dnsHostnamePattern})(?::\\d{1,5})?$`);
```
ajhollid: This will reject bare IP addresses, that could break things for some users

client/src/Validation/statusPage.ts
```diff
@@ -10,8 +10,25 @@ import { cssReferencesExternalResource } from "@/Utils/customCss";
...
+const statusPageHostnamePattern =
+	"([a-zA-Z0-9_](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?\\.)+[a-zA-Z]{2,63}";
+
+const statusPageHostnameRegex = new RegExp(`^(?=.{1,253}$)${statusPageHostnamePattern}$`);
+
+// An embedding origin: scheme + hostname (or localhost) + optional port, no path.
+// Kept in sync with embedOriginRegex in server/src/api/validation/shared.ts.
+const statusPageOriginRegex = new RegExp(
+	`^https?:\\/\\/(?:localhost|${statusPageHostnamePattern})(:\\d{1,5})?$`
+);
```
ajhollid: Same here, this will reject bare IP addresses, we should widen to allow those

server/src/app.ts
```diff
@@ -50,7 +50,29 @@ export const createApp = ({
...
+					"object-src": ["'none'"],
+					"base-uri": ["'self'"],
+					// Emitted per request by statusPageDocumentCsp so status pages can allow configured embedding origins.
+					"frame-ancestors": null,
+				},
+			},
+		})
+	);
+
+	app.use(createStatusPageDocumentCsp(allowedOrigin, services.statusPagesRepository));
```
ajhollid, line 91: Moving this up changes the CSP for the root document.

We would need to set `connect-src` so that users that a base API set that is different from their origin can still connect.

You can get the base API url from `envSettings` and extract the origin from it

ajhollid (changes requested):
Looks good for the most part, a couple of mechanical changes though that I believe are needed.  Thanks for your work here!

client/src/Validation/statusPage.ts
web-engineer: Widened to accept IPv4 and bracketed IPv6 hosts, validated with zod's ipv4/ipv6 checks (same as the DNS server validator). Tests added for both.

server/src/api/validation/shared.ts
web-engineer: Widened to accept IPv4 and bracketed IPv6 hosts, validated with zod's ipv4/ipv6 checks (same as the DNS server validator). Tests added for both.

server/src/app.ts
web-engineer: Added connect-src 'self' plus the origin of CLIENT_CONFIG_API_BASE_URL when it is an absolute URL. Verified with curl that the root document carries connect-src 'self' https://api.example.com with the override set, and 'self' alone without it.

PR #3943, opened 2026-09-11
fix(docker): bump MongoDB to 8.2 so the stack starts on Linux 6.19+ kernels

### LINKED ISSUE #3842: MongoDB fails to start on Linux kernel 6.19+ / 7.0.x (SERVER-121912) - app returns 502, DB connection refused
After upgrading my host to a newer Linux kernel (Proxmox VE 9, kernel `7.0.14-12-pve` - Proxmox 8 goes out of support on 31st August 2026 and Proxmox 9.2, which is the current version, ships by default with trixie kernel 7.0.14), Checkmate started returning 502. The Checkmate server was up but could not reach MongoDB, logging repeated `connect ECONNREFUSED 127.0.0.1:27017`. The root cause is not Checkmate itself but a known MongoDB issue: MongoDB 8.0+ refuses to start on Linux kernel 6.19 and newer due to a TCMalloc/rseq incompatibility, exiting at startup with:

```
MongoDB cannot start: Linux kernel versions 6.19 and newer has a known incompatibility with this version of MongoDB. See https://jira.mongodb.org/browse/SERVER-121912 for more information.
```

## Environment

Checkmate installed via the Proxmox community-scripts LXC (Debian container)
MongoDB 8.0.28, then updated to 8.0.29 - both are affected
Host kernel: `7.0.14-12-pve`

## What I tried that did NOT work

Updating MongoDB within the 8.0 series (8.0.28 to 8.0.29): still gated. The startup check in these builds still blocks 6.19+ and does not yet recognise fixed kernels.
Setting `GLIBC_TUNABLES=glibc.pthread.rseq=0`: no effect.

## What DID work

A systemd drop-in that sets the rseq tunable to 1 for the mongod service:

```
mkdir -p /etc/systemd/system/mongod.service.d
cat <<EOF >/etc/systemd/system/mongod.service.d/rseq.conf
[Service]
Environment=GLIBC_TUNABLES=glibc.pthread.rseq=1
EOF
systemctl daemon-reload
systemctl restart mongod
```

After this, mongod starts normally and Checkmate reconnects. I confirmed it survives a full host reboot.

For Docker Compose deployments, the equivalent is adding to the mongo service:

```
environment:
  GLIBC_TUNABLES: "glibc.pthread.rseq=1"
```

## Suggestion

It may be worth documenting this in the README/troubleshooting, since anyone on a current kernel (Debian 13, Ubuntu 26.04, Proxmox VE 9, Fedora 43+) will hit it. This is a temporary workaround; the real fix won't land until MongoDB ships a patched TCMalloc. But this has been ongoing since May, meanwhile it's now middle of August and no progress on that.

## Sources

MongoDB Jira SERVER-121912 (TCMalloc rseq bug): https://jira.mongodb.org/browse/SERVER-121912
MongoDB 8.0 release notes (affected kernel range 6.19 through 7.0.13): https://www.mongodb.com/docs/manual/release-notes/8.0/

usualdesigner:
## Describe your changes

Bumps the MongoDB image from `mongo:8.0` to `mongo:8.2` in both Compose files, and in the two documented `docker run` dev-database commands.

Fixes #3842.

`mongo:8.0` refuses to start on Linux kernels **6.19 through 7.0.13** — the TCMalloc it vendors violates the upstream rseq ABI, so MongoDB detects those kernels and exits during startup ([MongoDB 8.0 release notes](https://www.mongodb.com/docs/manual/release-notes/8.0/), SERVER-121912). Since the Compose files gate `checkmate` on `mongodb: condition: service_healthy`, the whole stack never comes up: MongoDB crash-loops and `checkmate` sits in `Created`. That is the 502 in #3842, and it now hits anyone on Debian 13, Ubuntu 26.04, Proxmox VE 9 or Fedora 43+.

**8.2 has no kernel gate at all.** Grepping the shipped binaries:

| | `mongo:8.0` (8.0.30) | `mongo:8.2` (8.2.12) |
|---|---|---|
| `"known incompatibility with this version of MongoDB"` | 1 | **0** |
| `"6.19"` | 1 | **0** |
| `SERVER-121912` | 1 | **0** |

SERVER-121912 is closed as *Gone away* — the rseq violation was fixed upstream in TCMalloc — and SERVER-125742 removed the startup gate. 8.2 is also already what this project pins in `docs/hosting/checkmate-so/docker-compose.yaml`, so it is not a new dependency for the team.

## Why not the `GLIBC_TUNABLES` workaround

#3842 documents `GLIBC_TUNABLES=glibc.pthread.rseq=1`, and it does make 8.0 start — I verified that too. I deliberately did **not** submit that, because on 6.19–7.0.13 the crash the gate prevents is real: forcing rseq on keeps users on the unfixed allocator and swaps a loud refusal at startup for a latent memory-allocator bug at runtime. It is the right emergency fix for someone who cannot change images today; it is the wrong thing to ship as the default. The reporter says as much in the issue — *"the real fix won't land until MongoDB ships a patched TCMalloc"*. It has now landed, in 8.2.

## Testing

Ubuntu 26.04.1 LTS, kernel `7.0.0-1011-gcp` (inside the affected range), Docker 29.8.0. Same host, back to back, `docker/docker-compose.yaml` with nothing else changed.

**Before**

```
checkmate   created       Created
mongodb     restarting    Restarting (1)

mongodb-1  | "msg":"MongoDB cannot start: Linux kernel versions 6.19 and newer
              has a known incompatibility with this version of MongoDB."
$ curl http://localhost:52345/   ->  connection refused
```

**After**

```
checkmate   running    Up About a minute (healthy)
mongodb     running    Up About a minute (healthy)

mongodb-1  | "msg":"Waiting for connections","attr":{"port":27017,"ssl":"off"}
$ curl http://localhost:52345/   ->  200   <title>Checkmate</title>
$ mongosh --eval 'db.version()'  ->  8.2.12
```

**Upgrade and rollback for existing installs** — on a volume written by 8.0:

- `mongo:8.2` opens it, `featureCompatibilityVersion` stays `8.0`, documents intact.
- Putting `mongo:8.0` back afterwards still starts and still reads the data.

So this is a tag change for existing deployments, not a data migration, and it is reversible as long as nobody runs `setFeatureCompatibilityVersion: "8.2"`.

The existing `mongosh` healthcheck passes unchanged, so nothing else in either file needed touching.

## Checklist

- [x] Tested locally and confirmed the change works
- [x] Related issue referenced (#3842)
- [x] Addresses only one topic
- [x] No user-facing strings, styling or UI changed — no screenshots apply
- [ ] Issue assigned to me — I can't self-assign; happy for a maintainer to do it

`lint` / `format-check` / `build` are unaffected: this touches no files under `client/` or `server/`. The two remaining `mongo:8.0` strings in the repo are mock container fixtures in `server/test/unit`, deliberately left alone.

## Note on overlap

#3878 also edits both Compose files, but only the `healthcheck:` line, while this touches only `image:`. They shouldn't conflict, and I'm happy to rebase on whichever lands first.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -37,7 +37,7 @@ npm run format           # Prettier formatting
 ```bash
 docker build -f docker/Dockerfile -t checkmate .          # build the mono image (server + built client)
 docker compose -f docker/dev/docker-compose.yaml up       # local full stack (builds image, runs mongo + worker)
-docker run -d -p 27017:27017 -v uptime_mongo_data:/data/db --name uptime_database_mongo mongo:8.0   # just a dev database
+docker run -d -p 27017:27017 -v uptime_mongo_data:/data/db --name uptime_database_mongo mongo:8.2   # just a dev database
 ```
 
 ## Environment Setup
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -91,7 +91,7 @@ cd Checkmate
 Run MongoDB container:
 
 ```bash
-docker run -d -p 27017:27017 -v uptime_mongo_data:/data/db --name uptime_database_mongo mongo:8.0
+docker run -d -p 27017:27017 -v uptime_mongo_data:/data/db --name uptime_database_mongo mongo:8.2
 ```
 
 #### Step 3: Set Up the Backend (Server)
--- a/docker/dev/docker-compose.yaml
+++ b/docker/dev/docker-compose.yaml
@@ -39,7 +39,7 @@ services:
       mongodb:
         condition: service_healthy
   mongodb:
-    image: mongo:8.0
+    image: mongo:8.2
     restart: always
     ports:
       - "27017:27017"
--- a/docker/docker-compose.yaml
+++ b/docker/docker-compose.yaml
@@ -29,7 +29,7 @@ services:
         condition: service_healthy
 
   mongodb:
-    image: mongo:8.0
+    image: mongo:8.2
     restart: always
     command: ["mongod", "--quiet", "--bind_ip_all"]
     volumes:


usualdesigner:
Closing this — **my recommendation was wrong, and merging it would have made things worse.** Sorry for the noise; the evidence I missed is below, because it invalidates several workarounds circulating for #3842, not just mine.

I claimed 8.2 "has no kernel gate, therefore it carries the fix." I had the causation backwards. The gate was *introduced* partway through the 8.0 line, and I only checked builds on one side of it:

```
mongo:8.0.12   gate absent
mongo:8.0.20   gate absent
mongo:8.0.28   gate present
mongo:8.0.29   gate present
mongo:8.0.30   gate present
mongo:8.2.12   gate absent     <- what this PR pinned
mongo:8.3.9    gate present    <- current mongo:8 and mongo:latest
```

8.2 is an EOL rapid-release series — it is absent from MongoDB's release-notes index, which lists 8.3 as current and 8.0 as the previous stable. Its last build simply predates the guard. It is not a fixed build; it is an **unguarded** one.

The decisive point I should have caught: **8.3.9 is newer than 8.2.12 and still gates.** If the vendored TCMalloc had been fixed, the newest release would not need the check. And MongoDB's own docs still state the remedy in kernel terms, not version terms:

> Linux kernel version 7.0.14 and later resolves this incompatibility. To run MongoDB on an affected system, upgrade to Linux kernel version 7.0.14 or later.

So pinning 8.2 would have swapped a loud, correct refusal at startup for a silent run on the buggy allocator — which is precisely the objection I raised in this PR against `GLIBC_TUNABLES=glibc.pthread.rseq=1`. Both do the same thing. My version was worse, because it looked like an upgrade.

### What this means for #3842

Every image-side workaround I can find is a guard-removal, not a fix:

- `GLIBC_TUNABLES=glibc.pthread.rseq=1` — forces rseq on under the unfixed allocator
- pinning `mongo:8.2` or an 8.0.x before .28 — predates the guard
- `mongo:8` / `mongo:latest` — resolves to **8.3.9 today, which gates and will not start at all**, so anyone following "just use `mongo:8`" gets a stack that still won't boot

On kernels **6.19 through 7.0.13** the crash is real and the guard is doing its job. The actual remedy is kernel **7.0.14+**.

That does leave a genuine gap: the failure surfaces as `checkmate` stuck in `Created` behind `condition: service_healthy` and a 502, with the cause only visible in `docker compose logs mongodb`. If a short troubleshooting note — affected kernel range, the one-line log signature, and "upgrade to 7.0.14+" — would be useful in the README or `docker/`, I'm happy to open that as a docs-only PR. That's a maintainer call, so I won't presume this time.

Verified throughout on Ubuntu 26.04.1, kernel `7.0.0-1011-gcp` (inside the affected range), Docker 29.8.0.

PR #3945, opened 2026-09-11
Add instance egress self-check (#3937)

### LINKED ISSUE #3937: Mitigating false incidents temporary egress failiure
When our server or cluster loses outbound internet connectivity for a few minutes, every monitor's checks fail at the same time, since they all run from the same egress path. Checkmate has no way to distinguish this from a genuine outage of the monitored targets, so it fires incidents and alerts for every monitor at once, even though in reality nothing may actually be down.

An instance-level "egress self-check": in Settings, let an admin specify a small list of known-reliable domains or IPs to probe on a schedule (e.g. 1.1.1.1, 8.8.8.8, or a custom list). If all of them fail, treat this as evidence that the instance itself has lost connectivity, and suppress incident creation and alerting until the check succeeds again. Include an option to send a notification when this state is detected, using the same notification channels (email, Slack, Discord, webhook, etc.) already configurable for monitors.

- Raising `statusWindowSize`/`statusWindowThreshold` on monitors to require more failures before flipping to down. Reduces false positives somewhat but delays real-incident detection too, and doesn't address the root cause.
- Routing monitors through a proxy on a separate network path (`proxyMode: custom`), so their checks don't share the instance's own egress. Effective per monitor, but requires hosting a proxy elsewhere and doesn't help monitors left on the default path.
- Using the existing Geo-Distributed Checks (Globalping) data as a cross-check against a monitor's own failed check before creating an incident. More precise, but only covers monitors with geo-checks enabled (HTTP/ping only), and ties the incident-decision path to a third-party API rather than just a display feature.

**Additional context**
Happy to help scope or implement this — I'd been exploring how it could hook into the existing `AppSettings` singleton and notification pipeline. Would like to know if this sounds useful to others too.

web-engineer:
## Describe your changes

Adds an instance egress self-check so that a loss of outbound connectivity from the Checkmate host is not reported as an outage of every monitored target. Design and discussion are on the issue; this PR is the first implementation, opened as a draft for architecture review.

**Worker**

- When a check fails, `CheckProducer` asks `EgressService` to probe the configured targets (default `1.1.1.1`, `8.8.8.8`; bare host = ICMP, `host:port` = TCP, `http(s)://` = HTTP, via the existing `NetworkService` providers). If none are reachable the check is written with `egressStatus: "degraded"`.
- `CheckEvaluator` short-circuits on a degraded check: no status change, no incident, no notification, and the check stays out of the status window and `MonitorStats`, so monitors do not flip to down and do not fire a spurious "up" on recovery.
- Egress state is a persisted Mongo singleton (`domain/egress/egress-state.*`) with atomic ok→degraded and degraded→ok transitions, so the API and one or more worker processes see the same state and exactly one worker sends the recovery notification.
- While degraded, a dedicated unref'd timer re-probes at `egressPollIntervalSeconds` (default 30 s). On recovery it sends a single `egress_recovered` notification to the channels configured in settings.
- Degraded checks are excluded from uptime percentage aggregations (summary, daily buckets, date-range uptime, hardware and docker totals). Response-time graph series are not filtered.

**API**

- `GET /api/v1/egress` (JWT) returns the current state for the client banner. OpenAPI spec updated.
- `AppSettings` gains `egressCheckEnabled`, `egressCheckTargets`, `egressPollIntervalSeconds`, `egressNotifications`. PATCH validates the notification ids exist.

**Client**

- Admin-only "Egress self-check" card on Settings → Monitoring.
- Banner on every page while egress is degraded, polling `/egress` every 30 s.
- Checks table labels degraded checks.

**Points I'd like a view on**

- PagerDuty treats every notification as an incident, so `egress_recovered` would open one. Suppress for that provider, or leave to the user?
- Probes deliberately bypass the global proxy, on the basis that the question is "can this host reach the internet", not "can the proxy".
- `settingsController` calls `notificationsRepository` directly to verify ids exist. Happy to move that into a service if preferred.

Tested on our cluster: synthetic failure (unreachable targets) shows the banner, flags the checks, opens no incidents, clears within one poll interval of restoring the targets, and sends one Slack recovery message.

Settings card:

Egress self-check settings card

Banner while degraded:

Degraded egress banner

Recovery notification:

Slack recovery notification

Fixes #3937

```Javascript
<div>{t('add')}</div>
```

- [x] I have **not** included any files that are not related to my pull request, including package-lock and package-json if dependencies have not changed. Note: `server/openapi.json` was regenerated with `npm run build:openapi`, and the committed spec was already behind the routes, so the diff there includes some pre-existing drift as well as the new egress route. Happy to trim it to the egress change only if you'd rather.

### DIFF
--- a/client/src/Components/design-elements/EgressBanner.tsx
+++ b/client/src/Components/design-elements/EgressBanner.tsx
@@ -1,0 +1,59 @@
+import Box from "@mui/material/Box";
+import Stack from "@mui/material/Stack";
+import Typography from "@mui/material/Typography";
+import { useTheme } from "@mui/material/styles";
+import { useTranslation } from "react-i18next";
+import { useSelector } from "react-redux";
+import { TriangleAlert } from "lucide-react";
+import type { RootState } from "@/Types/state";
+import type { EgressState } from "@/Types/Egress";
+import { formatDateWithTz } from "@/Utils/TimeUtils";
+import { LAYOUT } from "@/Utils/Theme/constants";
+
+interface EgressBannerProps {
+	state: EgressState | null;
+}
+
+export const EgressBanner = ({ state }: EgressBannerProps) => {
+	const theme = useTheme();
+	const { t } = useTranslation();
+	const uiTimezone = useSelector((s: RootState) => s.ui.timezone);
+
+	if (state?.status !== "degraded") return null;
+
+	const message = state.degradedSince
+		? t("components.egressBanner.degraded", {
+				time: formatDateWithTz(state.degradedSince, "ddd, MMM D, HH:mm", uiTimezone),
+			})
+		: t("components.egressBanner.degradedNoTime");
+
+	return (
+		<Box
+			position="fixed"
+			top={0}
+			left={0}
+			right={0}
+			zIndex={theme.zIndex.snackbar}
+			bgcolor={theme.palette.warning.main}
+			color={theme.palette.warning.contrastText}
+			px={theme.spacing(LAYOUT.MD)}
+			py={theme.spacing(LAYOUT.XS)}
+			role="status"
+		>
+			<Stack
+				direction="row"
+				alignItems="center"
+				justifyContent="center"
+				gap={theme.spacing(LAYOUT.XS)}
+			>
+				<TriangleAlert size={20} />
+				<Typography
+					variant="body2"
+					fontWeight={500}
+				>
+					{message}
+				</Typography>
+			</Stack>
+		</Box>
+	);
+};
--- a/client/src/Components/design-elements/index.tsx
+++ b/client/src/Components/design-elements/index.tsx
@@ -24,6 +24,7 @@ export * from "./SplitBox";
 export * from "./TextLink";
 export * from "./NoticeBanner";
 export * from "./OfflineBanner";
+export * from "./EgressBanner";
 export * from "./Avatar";
 export * from "./StatusCodeLabel";
 export * from "./StrategyBadge";
--- a/client/src/Components/layout/AppLayout.tsx
+++ b/client/src/Components/layout/AppLayout.tsx
@@ -1,8 +1,14 @@
 import Box from "@mui/material/Box";
 import { useState, useEffect, useRef } from "react";
 import { useTheme } from "@mui/material/styles";
-import { OfflineBanner } from "@/Components/design-elements";
+import { useSelector } from "react-redux";
+import { OfflineBanner, EgressBanner } from "@/Components/design-elements";
 import { setServerUnreachableCallback, get } from "@/Utils/ApiClient";
+import { useGet } from "@/Hooks/UseApi";
+import type { RootState } from "@/Types/state";
+import type { EgressState } from "@/Types/Egress";
+
+const EGRESS_POLL_MS = 30_000;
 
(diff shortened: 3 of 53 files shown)


ajhollid (changes requested):
@web-engineer This is a really good feature, but 2k loc change is not really reasonable to review in one go.

I would really like to see this feature merged in, but we should discuss the design first and break this up into reasonably reviewable chunks so we can properly review the pieces.

Something like

1.  Egress check functionality
2.  Egress check result persistence
3.  Notifications
4.  Front end / API route

I appreaciate the idea and the work so far, but the team just can't review a PR that's roughly 10% of the codebase.

web-engineer:
I’ll take a look and break it up, apologies - just wanted to get this done as we’ve had a patch of intermittent issues with egress and the amount of alerts was getting unmanageable.

web-engineer:
I've split this into four PRs, each building on the last, in the order you suggested. Every branch compiles and passes the server test suite on its own, and each PR shows only its own diff once the previous one has merged.

| # | Branch | Scope | Size |
|---|---|---|---|
| 1 | `feat/egress-1-check-service` | `EgressService` (probe targets, degraded/ok state machine, recovery poll loop), persisted `EgressState` singleton, `AppSettings` egress fields and validation, DI wiring. Nothing calls it yet, so behaviour is unchanged. | +1080 in 17 files |
| 2 | `feat/egress-2-check-persistence` | `Check.egressStatus`, `CheckProducer` calls the probe on a failed check, `CheckEvaluator` short-circuits degraded checks, uptime aggregations exclude them. | +172 in 11 files |
| 3 | `feat/egress-3-recovery-notification` | `egress_recovered` notification type, message builder, `NotificationsService.sendEgressRecoveredNotification`, settings controller validates the notification ids. | +245 in 11 files (two commits: a no-op refactor, then the feature) |
| 4 | `feat/egress-4-api-and-ui` | `GET /api/v1/egress`, OpenAPI route, settings card, banner, checks-table label. | +517 in 20 files |

PR 1 is open now: https://github.com/bluewave-labs/Checkmate/pull/3959. I'll open each of the others once the previous one has merged. Compare views for the rest in the meantime:

- 2: https://github.com/web-engineer/Checkmate/compare/feat/egress-1-check-service...feat/egress-2-check-persistence
- 3: https://github.com/web-engineer/Checkmate/compare/feat/egress-2-check-persistence...feat/egress-3-recovery-notification
- 4: https://github.com/web-engineer/Checkmate/compare/feat/egress-3-recovery-notification...feat/egress-4-api-and-ui

Three things in the implementation differ from the plan on #3937, worth knowing before you read PR 1:

1. Degraded state is a persisted Mongo singleton (`domain/egress/egress-state.*`), not in memory. The worker can run as a separate process from the API and there may be several workers, so the transitions are atomic compare-and-set and only one worker sends the recovery notification.
2. Suppression happens in `CheckEvaluator`, not only in `MonitorStatusPolicy`. A degraded check has to stay out of the status window and `MonitorStats` too, otherwise the monitor flips to down silently and fires a spurious "up" on recovery.
3. The recovery poll runs on a dedicated unref'd timer in the worker (open question 2 on the plan), default 30 s.

The unrelated `openapi.json` regeneration noise from this PR is gone; each PR now carries only its own spec change. Closing this one in favour of the four above.

PR #3957, opened 2026-09-13
fix(notifications): don't report a temperature breach exactly at the threshold

### LINKED ISSUE #3956: Notification content lists a temperature breach at exactly the threshold, inconsistent with the incident-driving check
**Describe the bug**

`NotificationMessageBuilder.extractThresholdBreaches` reports a temperature breach with `maxTemp >= threshold`, while `StatusService.computeHardwareStatus` — the check that actually drives the "breached" status transition and incident creation — uses a strict `maxTemp > threshold` for the same comparison (as do cpu, memory and disk in both files; temp is the only metric where the two files disagree).

**To reproduce**

1. A hardware monitor has `tempAlertThreshold: 80` and is already `breached` for some other reason (say `cpuAlertThreshold` exceeded).
2. A check reports `cpu.temperature: [80]` — exactly at the threshold, not over it.
3. `StatusService` correctly does not count temp among the breaching metrics (`80 > 80` is false).
4. `extractThresholdBreaches`, called separately to build the notification content, evaluates `80 >= 80` as true and lists temperature as a breaching metric in the alert.

**Expected behavior**

Temperature exactly at the threshold should not be reported as a breach, matching cpu/memory/disk's own boundary behavior and the status-transition check for the same metric.

PR with the one-character fix (`>=` → `>`) plus a regression test coming right after this.

Asgabani:
Fixes #3956.

## What changed

`NotificationMessageBuilder.extractThresholdBreaches` compared temperature with `maxTemp >= threshold`, while `StatusService.computeHardwareStatus` — the check that actually drives the "breached" status transition and incident creation — uses `maxTemp > threshold` for the same metric. cpu, memory and disk already use `>` consistently in both files; temp was the only one where the two diverged.

Practical effect: a monitor already `breached` for some other metric (say CPU) would have its notification content list temperature as a breaching metric when temp sat exactly at its threshold, even though that same reading would never have caused a "breached" transition on its own.

One-line fix (`>=` → `>`) plus a regression test covering the exact-threshold boundary, mirroring the existing "below threshold" tests for the same metric.

## Testing

- `npm run lint` (server): clean.
- `npm run format` (server): unchanged, already formatted.
- `npx jest --no-coverage test/unit/services/notificationMessageBuilder.test.ts`: 57/57 pass (new boundary test included).
- `npx jest --no-coverage test/unit/services/statusService.test.ts`: 94/94 pass (confirms no regression on the check this fix aligns with).
- Full suite (`npx jest --coverage`): 93 suites / 1728 tests pass; `notification.message-builder.ts` at 100% line coverage.
- `npm run build` (server): succeeds.


ajhollid (approved):
Thanks for catching this, appreciate your contribution to the project!

PR #3958, opened 2026-09-14
Feat/invite status and duration

MahlonW:
## Describe your changes

Adds pending-invite management to the Team tab: admins can now see outstanding invites, change how long an invite stays valid, and revoke one outright.

Previously every invite expired exactly 1 hour after creation, enforced by a fixed-window Mongo TTL index (`expires: 3600` anchored to creation time), there was no way to see outstanding invites, and no way to revoke one short of waiting for it to lapse.

**Backend**

- `invite.model.ts` now stores an absolute expiry timestamp with a per-document TTL index (`expireAfterSeconds: 0`), so each invite carries its own duration.
- New repository methods: `findById`, `findByTeamId`, `updateExpiryById`, `deleteById` — all team-scoped, returning 404 on cross-team access.
- New endpoints, all behind `verifyJWT` + `isAllowed(["admin", "superadmin"])`:
  - `GET /invite` — list pending invites for the team
  - `PATCH /invite/:id/expiry` — change how long an invite stays valid
  - `DELETE /invite/:id` — revoke a pending invite
- `InviteService.getInvites` / `updateInviteExpiry` strip the invite `token` before returning — it is a bearer credential for accepting the invite and must never leave the send endpoint.
- `updateInviteExpiry` and `deleteInvite` are both gated by the same role check used on invite creation: an actor may only act on invites for roles they are allowed to manage, so an admin cannot revoke or extend a superadmin invite.
- Duration bounds (1 hour – 30 days, whole hours) centralised in `invite.constants.ts` and shared by both the create and update validators.
- Migration `0013_convertInviteExpiryToAbsoluteTtl` advances existing invites' stored `expiry` by the legacy 1-hour window so already-issued invites expire at the same wall-clock moment as before, then drops the old fixed-window index so the post-migration `syncIndexes()` pass can create the new per-document one.

**Frontend**

- `PendingInvitesTable` — email, role, status (pending/expired), expiry, and a row actions menu with "Change duration" and a red "Delete".
- `ChangeInviteDurationDialog` + `DurationSelectField`, sharing a single `INVITE_DURATION_OPTIONS` list.
- `DeleteInviteDialog` — destructive confirm naming the invitee, warning that the existing link stops working.
- `InviteTeamMemberDialog` gains a duration picker so the duration can be set at creation time.

Fixes #

```Javascript
<div>{t('add')}</div>
```

image
image
image
image
image


ajhollid (changes requested):
Conceptually this looks fine, but there are some issues to resolve.

As far as I can tell, expiry is never actually enforced.  All queries should now be filtered by expiry date, unless I'm missing something.

The migration can also be applied more than once, please make sure it is guarded correctly.

Please remove all the narrative comments as well, they mostly just add noise.

Additionally please make sure to format, this PR is failing format checks.  Thank you!

PR #3959, opened 2026-09-14
Egress self-check (1/4): egress state and probe service (#3937)

web-engineer:
## Describe your changes

First of four PRs for the instance egress self-check (#3937), split from #3945 as requested. This one adds the service and state and wires it up; nothing calls it yet, so runtime behaviour is unchanged.

- `EgressService` (`server/src/domain/egress/egress.service.ts`): probes a list of reliability targets using the existing `NetworkService` providers (bare host = ICMP, `host:port` = TCP, `http(s)://` = HTTP). `assessAfterFailure()` is the entry point the check pipeline will call in PR 2: if egress is already degraded it returns `degraded`; otherwise it probes, and if no target is reachable it marks the instance degraded and starts the recovery poll loop. The loop re-probes at `egressPollIntervalSeconds` and clears the state when any target answers. Recovery is silent in this PR; the notification arrives in PR 3.
- `EgressState` persisted singleton (`egress-state.model.ts`, `.repository.mongo.ts`): atomic `markDegraded` / `markRecovered` so that with several worker processes only one performs each transition. `EgressStateService` is the read-only API-side accessor for PR 4.
- `AppSettings` gains `egressCheckEnabled`, `egressCheckTargets`, `egressPollIntervalSeconds`, `egressNotifications`, with Zod validation on the settings PATCH body. An empty target list falls back to the defaults (`1.1.1.1`, `8.8.8.8`).
- DI: `egressStateRepository` in shared services, `EgressStateService` in API services, `EgressService` constructed and `init()`-ed in the worker so a degraded episode survives a restart.
- Tests: `egressService.test.ts`, `egressStateService.test.ts`, `settingsValidation.test.ts`.

Next: PR 2 (`Check.egressStatus`, producer/evaluator hook, uptime exclusions), PR 3 (recovery notification), PR 4 (API route and UI). Compare views on the fork:

- https://github.com/web-engineer/Checkmate/compare/feat/egress-1-check-service...feat/egress-2-check-persistence
- https://github.com/web-engineer/Checkmate/compare/feat/egress-2-check-persistence...feat/egress-3-recovery-notification
- https://github.com/web-engineer/Checkmate/compare/feat/egress-3-recovery-notification...feat/egress-4-api-and-ui

- [x] I have **not** included any files that are not related to my pull request. `server/openapi.json` is regenerated and the diff is limited to the new settings fields.

🔧


server/src/domain/egress/egress.service.ts
```diff
@@ -0,0 +1,290 @@
...
+		return { ...base, type: "ping", url: target } as unknown as Monitor;
+	};
+
+	private probeTarget = async (target: string): Promise<EgressProbeResult> => {
+		const monitor = this.toProbeMonitor(target);
+		let timer: NodeJS.Timeout | undefined;
+
+		const { response, responseTime, error } = await timeRequest(() =>
+			Promise.race([
+				this.networkService.requestStatus(monitor),
```
ajhollid: There's no context passed here, so this will bypass a proxy if one has been set by the user.  Egress would always fail.  Proxy setting should be extracted and passed along as the second param.

server/src/domain/egress/egress.service.ts
```diff
@@ -0,0 +1,290 @@
...
+			return { ...base, type: "http", url: target } as unknown as Monitor;
+		}
+
+		const hostPort = HOST_PORT_PATTERN.exec(target);
+		if (hostPort) {
+			const host = hostPort[1] ?? hostPort[2];
+			return { ...base, type: "port", url: host, port: Number(hostPort[3]) } as unknown as Monitor;
+		}
+
+		return { ...base, type: "ping", url: target } as unknown as Monitor;
```
ajhollid: Ping provider sanitizes bare IPs

```
	private sanitizeHost(url: string): string {
		return url
			.replace(/^https?:\/\//, "")
			.replace(/\/.*$/, "")
			.replace(/:.*/, "");
	}
```

This would break the app if an IPV6 address was used for egress checks

server/src/domain/egress/egress.service.ts
```diff
@@ -0,0 +1,290 @@
...
+				target,
+				reachable: false,
+				responseTime,
+				message: error instanceof Error ? error.message : error ? String(error) : "No response",
+			};
+		}
+
+		return {
+			target,
+			reachable: response.status === true,
```
ajhollid: Not sure this is the right call here.  Does a 400 response mean egress failed?  I don't think so 🤔

server/src/domain/egress/egress.service.ts
```diff
@@ -0,0 +1,290 @@
...
+		const intervalMs = await this.readPollIntervalMs();
+		if (!this.polling) {
+			return;
+		}
+		this.pollTimer = setTimeout(() => void this.tick(), intervalMs);
+		// Never hold the process open for this loop; the state is persisted and init() resumes it on restart.
+		this.pollTimer.unref();
+	};
+
+	private tick = async () => {
```
ajhollid: Unless I'm missing something here, there's no reset of the singleton in the recovery path.

IE If i start with the service enabled, reach a degraded status, disable the service, and restart, I will be stuck in degraded service forever.

The singleton should have its state reset on recovery, as well as on init if the service is disabled.

server/src/domain/egress/egress.service.ts
```diff
@@ -0,0 +1,290 @@
...
+		const targets = (configured ?? []).map((target) => target.trim()).filter((target) => target.length > 0);
+		// An empty list would make "all unreachable" vacuously true, so fall back to the defaults.
+		return targets.length > 0 ? targets : [...DEFAULT_EGRESS_TARGETS];
+	};
+
+	// ****************************
+	// Event-triggered assessment
+	// ****************************
+
+	assessAfterFailure = async (): Promise<EgressStatus | null> => {
```
ajhollid: Presumanbly this will be called for _every_ check, which would be very expensive if you have a significant number of monitors.

This particular issue already has an established pattern in `ProxyResolver` for handling concurrent inflight checks.  For egress it can be even simpler, we should just store the egress check promise, and any concurrent checks get the same promise rather than creating a new one.

Check out the implementaiton in `ProxyResolver`.

server/src/domain/egress/egress.service.ts
```diff
@@ -0,0 +1,290 @@
...
+
+	probeTargets = async (targets: string[]): Promise<EgressProbeResult[]> => {
+		return await Promise.all(targets.map((target) => this.probeTarget(target)));
+	};
+
+	private resolveTargets = (configured: string[] | undefined): string[] => {
+		const targets = (configured ?? []).map((target) => target.trim()).filter((target) => target.length > 0);
+		// An empty list would make "all unreachable" vacuously true, so fall back to the defaults.
+		return targets.length > 0 ? targets : [...DEFAULT_EGRESS_TARGETS];
+	};
```
ajhollid: This creates a potentially hard to debug issue.  If for whatever reason you submitted an empty array for your targets, you would expect egress to fail, but it succeeds silently because of the default egress targets.

I suggest either blocking empty arrays at the validation level, or normalizing to the default values on submission of an empty array.  At least that way there's always an indication of what's going on.

server/src/domain/egress/egress-state.repository.mongo.ts
```diff
@@ -0,0 +1,67 @@
...
+			{ $setOnInsert: { singleton: true, status: "ok" } },
+			{ upsert: true, new: true, setDefaultsOnInsert: true }
+		).lean<EgressStateDocument>();
+		return this.toEntity(doc);
+	};
+
+	recordProbe = async (results: EgressProbeResult[], now: Date) => {
+		const doc = await EgressStateModel.findOneAndUpdate(
+			{ singleton: true },
+			{ $set: { lastProbeAt: now, lastProbeResults: results }, $setOnInsert: { singleton: true, status: "ok" } },
```
ajhollid: The set here is unguarded, so we have a concurrency issue.  Each probe assumes that the status it read is the current status, which may not be the case in the event of a slow probe.  I think the solution here is probably to pass the egress status along with the result.

server/src/domain/egress/egress-state.repository.mongo.ts
```diff
@@ -0,0 +1,67 @@
...
+		};
+	};
+
+	findSingleton = async () => {
+		const doc = await EgressStateModel.findOneAndUpdate(
+			{ singleton: true },
+			{ $setOnInsert: { singleton: true, status: "ok" } },
+			{ upsert: true, new: true, setDefaultsOnInsert: true }
+		).lean<EgressStateDocument>();
+		return this.toEntity(doc);
```
ajhollid: I think the intention here was something like

"Read if the doc exists, write a new one if it doesn't"

However what actually happens here is that `$set: {updatedAt: <now>}` is called on every read due to how timestamps work.


https://mongoosejs.com/docs/timestamps.html

This should be separated into clean read and writes to avoid this.

ajhollid:
@web-engineer on closer review here I think we're doing way too much duplication of existing functionality, notably the polling.

I think a more elegant solution here is to do something like

1. A check fails, call `assessAfterFailure` to check for egress
2. Instead of the timer and recovery loop currently implemented, insert an egress check job into the scheduler that handles the recovery situation.

There are some details that need to be worked out like deduping in flight egress checks, but there is an existing pattern for that in `ProxyResolver`

This requires some familiarity with how the job scheduler works, but it will reduce complexity significantly here.

I'm happy to jump in and handle that, as the rest of the persistence implementation looks fine to me.

Let me know if you want to work on it, want to pass it off to me, or have other thoughts on the implementation

web-engineer:
> Let me know if you want to work on it, want to pass it off to me, or have other thoughts on the implementation

If it’s ok with you @ajhollid I’ll take a quick look as I’ve several chunks to follow to get this across the line …

ajhollid:
> > Let me know if you want to work on it, want to pass it off to me, or have other thoughts on the implementation
> 
> If it’s ok with you @ajhollid I’ll take a quick look as I’ve several chunks to follow to get this across the line …

Sure, take your time, theres no rush.

web-engineer:
Reworked as suggested. The timer loop is gone. When `assessAfterFailure` marks egress degraded it upserts a global `egress` job row (interval from `egressPollIntervalSeconds`), and `DBQueueWorker` runs `EgressService.checkRecovery` for that job type. Each run records the probe while degraded and deletes its own row once any target answers, so a restart mid-episode needs no startup hook and any worker can pick the job up.

Concurrent failures share one in-flight assessment, same shape as the `ProxyResolver` cache with a 5 s window, so a burst of failing checks produces one probe rather than one per monitor. `deleteByIdAndType` now accepts a null refId for the global row.

Tests updated; the full server suite passes. Verified locally by seeding a degraded state and an `egress` job with an unreachable target, watching the job run and record probes, then switching the target to a reachable URL: state returned to ok, the job row was removed and the recovery was logged.

server/src/worker/worker.db-queue.ts
```diff
@@ -175,6 +181,9 @@ export class DBQueueWorker extends JobScheduler implements IQueueWorker {
 				case "cleanup-retention":
 					await this.helper.getCleanupRetentionJob()(); // Get job and execute
 					break;
+				case "egress":
+					await this.egressService.checkRecovery(); // Re-probe while degraded; removes its own row on recovery
```
ajhollid: Jobs belong in the worker helper, and should be called with a factory method like the above `getCleanupRetentionJob()`

server/src/domain/egress/egress.service.ts
```diff
@@ -0,0 +1,245 @@
...
+		await this.jobsRepository.upsertJob(seed);
+	};
+
+	private removeRecoveryJob = async () => {
+		await this.jobsRepository.deleteByIdAndType(null, "egress");
+	};
+
+	// Runs on the queue at the configured interval while degraded. Errors propagate so the queue records
+	// the failure and retries with its usual backoff.
+	checkRecovery = async (): Promise<void> => {
```
ajhollid: As far as I can tell, disabling the feature once it is started doesn't reset the state of the singleton.  IE start with enabled, enter degraded state, disable, restart -> we're still in degraded state.

Enabling/disabling the service should reset the state.

server/src/domain/egress/egress.service.ts
```diff
@@ -0,0 +1,245 @@
...
+		return { ...base, type: "ping", url: target } as unknown as Monitor;
+	};
+
+	private probeTarget = async (target: string): Promise<EgressProbeResult> => {
+		const monitor = this.toProbeMonitor(target);
+		let timer: NodeJS.Timeout | undefined;
+
+		const { response, responseTime, error } = await timeRequest(() =>
+			Promise.race([
+				this.networkService.requestStatus(monitor),
```
ajhollid: There's no context passed here, so if the user is using a global proxy their proxy is not passed along and all egress checks will always fail

server/src/domain/egress/egress.service.ts
```diff
@@ -0,0 +1,245 @@
...
+			return { ...base, type: "http", url: target } as unknown as Monitor;
+		}
+
+		const hostPort = HOST_PORT_PATTERN.exec(target);
+		if (hostPort) {
+			const host = hostPort[1] ?? hostPort[2];
+			return { ...base, type: "port", url: host, port: Number(hostPort[3]) } as unknown as Monitor;
+		}
+
+		return { ...base, type: "ping", url: target } as unknown as Monitor;
```
ajhollid: IPV6 addresses will fail here due to the sanitization in the ping service

server/src/api/validation/settingsValidation.ts
```diff
@@ -1,6 +1,15 @@
...
+import { MAX_EGRESS_POLL_INTERVAL_SECONDS, MAX_EGRESS_TARGETS, MIN_EGRESS_POLL_INTERVAL_SECONDS } from "@/domain/egress/egress.type.js";
 import { z } from "zod";
 
+// A reliability target is a bare host/IP (ICMP ping), host:port (TCP connect) or an http(s) URL.
+const egressTargetValidation = z
+	.string()
+	.trim()
+	.min(1)
+	.max(253)
+	.regex(/^(https?:\/\/\S+|[A-Za-z0-9.\-:[\]]+)$/, "Invalid egress target");
```
ajhollid: This will allow malformed egress targets like addresses with unbalanced brackets.  If you only had one invalid egress target you'd forever be stuck in degraded state

server/src/api/validation/settingsValidation.ts
```diff
@@ -42,6 +51,11 @@ export const updateAppSettingsBodyValidation = z
 			.optional(),
 		globalProxyEnabled: z.boolean().optional(),
 		globalProxyId: z.string().nullable().optional(),
+
+		egressCheckEnabled: z.boolean().optional(),
+		egressCheckTargets: z.array(egressTargetValidation).max(MAX_EGRESS_TARGETS).optional(), // empty list falls back to DEFAULT_EGRESS_TARGETS
+		egressPollIntervalSeconds: z.number().int().min(MIN_EGRESS_POLL_INTERVAL_SECONDS).max(MAX_EGRESS_POLL_INTERVAL_SECONDS).optional(),
+		egressNotifications: z.array(z.string().regex(/^[a-f\d]{24}$/i, "Invalid notification id")).optional(),
```
ajhollid: The `/i` flag will break the Open API spec since it is passed directly to Zod's `regex`.  The regex should be updated to handle upper and lower case instead of using the flag

server/src/domain/egress/egress-state.repository.mongo.ts
```diff
@@ -0,0 +1,67 @@
...
+	};
+
+	recordProbe = async (results: EgressProbeResult[], now: Date) => {
+		const doc = await EgressStateModel.findOneAndUpdate(
+			{ singleton: true },
+			{ $set: { lastProbeAt: now, lastProbeResults: results }, $setOnInsert: { singleton: true, status: "ok" } },
+			{ upsert: true, new: true, setDefaultsOnInsert: true }
+		).lean<EgressStateDocument>();
```
ajhollid, line 64: I believe the intention here was to read if the doc exists, write if it does not.  However, since timestamps are set on this doc, Mongo appends `$set: { updatedAt: new Date() }` whenever `findOneAndUpdate` is called.  

https://mongoosejs.com/docs/timestamps.html

server/src/domain/egress/egress.service.ts
```diff
@@ -0,0 +1,245 @@
...
+		if (this.assessment && this.assessment.expiresAt > now) {
+			return this.assessment.value;
+		}
+		// Cache the in-flight promise so concurrent failures share one probe
+		const value = this.assess();
+		this.assessment = { value, expiresAt: now + ASSESSMENT_TTL_MS };
+		// Don't cache an internal failure, so the next failing check retries
+		void value.then((status) => {
+			if (status === null && this.assessment?.value === value) this.assessment = null;
+		});
```
ajhollid: `null` is the sentinel returned by `assess` when the service is disabled, so this actually evicts the `disabled` status and we get excess reads. We need a dedicated sentinel for `disabled` instead of `null`

server/src/domain/egress/egress.service.ts
```diff
@@ -0,0 +1,245 @@
...
+
+	// Runs on the queue at the configured interval while degraded. Errors propagate so the queue records
+	// the failure and retries with its usual backoff.
+	checkRecovery = async (): Promise<void> => {
+		const state = await this.egressStateRepository.findSingleton();
+		if (state.status !== "degraded") {
+			// Another worker recorded the recovery, or the state was reset; nothing left to poll for.
+			await this.removeRecoveryJob();
+			return;
+		}
```
ajhollid: This is a race.

W1:  Runs recovery job, egress is good state set to OK
W2:  Check fails, reads state, see's OK, probes, egress fails, sets state to degraded, upserts
W1:  Reaches next line and deletes the row W2 just upserted

The end state of the race is service is degraded with no recovery job running.

The deletion of the row cannot be unconditional

server/src/domain/egress/egress.service.ts
```diff
@@ -0,0 +1,245 @@
...
+	// ****************************
+
+	assessAfterFailure = (): Promise<EgressStatus | null> => {
+		const now = Date.now();
+		if (this.assessment && this.assessment.expiresAt > now) {
+			return this.assessment.value;
+		}
+		// Cache the in-flight promise so concurrent failures share one probe
+		const value = this.assess();
+		this.assessment = { value, expiresAt: now + ASSESSMENT_TTL_MS };
```
ajhollid: ASSESMENT_TTL_MS == PROBE_TIMEOUT_MS so this doesn't actually work as expected in the slow case where request >= PROBE_TIMEOUT_MS.  Concurrent failures for the fast case work, but in the slow case new probe rounds will be started every time

server/src/domain/app-settings/app-settings.model.ts
```diff
@@ -42,6 +43,10 @@ const AppSettingsSchema = new Schema<AppSettingsDocument>(
 		globalThresholds: { type: thresholdsSchema },
 		globalProxyEnabled: { type: Boolean, default: false },
 		globalProxyId: { type: String, default: null },
+		egressCheckEnabled: { type: Boolean, default: false },
+		egressCheckTargets: { type: [String], default: () => [...DEFAULT_EGRESS_TARGETS] },
+		egressPollIntervalSeconds: { type: Number, default: DEFAULT_EGRESS_POLL_INTERVAL_SECONDS },
+		egressNotifications: { type: [String], default: () => [] }, // Notification ids, stored as strings like globalProxyId
```
ajhollid, line 49: Deleting a notification needs to cascade to this or we'll have dangling notification IDs and failing notifications

ajhollid (changes requested):
This is more along the lines of what I was thinking!  There are significant issues to resovle though, so please have a look at my comments in the code review.

Thanks for your work here!

server/src/worker/worker.db-queue.ts
web-engineer: Moved. `WorkerHelper.getEgressRecoveryJob()` returns the handler and `DBQueueWorker` calls it like the cleanup jobs. The only difference is that it takes the claimed `Job`, which the conditional delete below needs.

server/src/domain/egress/egress.service.ts
web-engineer: Right, it did not. `EgressStateService.reset()` now clears the singleton and removes any pending recovery job, and the settings controller calls it whenever `egressCheckEnabled` changes in either direction. `checkRecovery` also releases the job if it finds the feature disabled mid-episode.

server/src/domain/egress/egress.service.ts
web-engineer: Fixed. Probes go through `ProxyResolver.resolve()` with `proxyMode: "inherit"` and pass `{ proxyUrl }` in the check context, as `CheckProducer` does.

server/src/domain/egress/egress.service.ts
web-engineer: Fixed in `PingProvider.sanitizeHost`: a bracketed address is unwrapped and a bare IPv6 address is left intact; only a single colon is treated as a port separator. Tests added.

server/src/api/validation/settingsValidation.ts
web-engineer: Replaced the regex with `parseEgressTarget()`, one function used by both the validator and the probe, so anything the validator accepts is something the probe can address. Unbalanced brackets, non-address bracket contents and out-of-range ports are rejected. Tests cover each.

server/src/api/validation/settingsValidation.ts
web-engineer: Yes, the generated pattern was `^[a-f\d]{24}$/i`, which is invalid. Now `^[0-9a-fA-F]{24}$`, matching the status page validators. `openapi.json` updated.

server/src/domain/egress/egress-state.repository.mongo.ts
web-engineer: Confirmed against the Mongoose timestamps docs. `findSingleton` now does a plain `findOne` and only upserts when the document is missing, with the first-insert duplicate-key race handled the way `upsertEvaluate` does. An integration test checks that a read leaves `updatedAt` alone.

server/src/domain/egress/egress.service.ts
web-engineer: Fixed. `assess` returns `"disabled"` as its own result and it is cached like `ok` and `degraded`; only an internal failure (`null`) is evicted.

server/src/domain/egress/egress.service.ts
web-engineer: 😱 The row is now removed with `deleteGlobalJobIfUnchanged(type, job.nextScheduledAt)`. A worker that flips ok to degraded re-sets `nextScheduledAt` when it schedules the job, so the delete from the stale run does not match and its job survives. Workers that only observe an existing degraded state use `upsertJob`, which never touches the schedule. The job is also scheduled before `markDegraded` now, so a crash between the two leaves a stray row that the first run removes rather than a degraded state nothing polls. Integration test added for the conditional delete.

server/src/domain/egress/egress.service.ts
web-engineer: Fixed. An in-flight assessment is shared for as long as it runs; the settled result is then reused for 5 s. A slow round can no longer be duplicated.

server/src/domain/app-settings/app-settings.model.ts
web-engineer: Added. `NotificationsService.deleteById` calls `settingsService.removeEgressNotification(id)`, an atomic `$pull` in the settings repository, after the team-scoped delete succeeds.

web-engineer:
@ajhollid thanks for the review, all eleven were confirmed. Each is addressed in the two latest commits with a test per point.

A second pass over the diff turned up a few more, also fixed: the recovery job is scheduled before the degraded transition is recorded; an http probe counts any HTTP response (including 4xx/5xx) as reachable and only a transport failure against egress, so a single target returning 503 no longer degrades the instance; global job rows are addressed by one key (`deleteGlobalJob`); `PingProvider` takes `net` through its constructor like `PortProvider`; integration tests added for `MongoEgressStateRepository` and the global-row delete methods.

Verified live against a throwaway Mongo: enabled the check through the API, seeded a degraded episode with unreachable targets including IPv6, watched the job run and record probes, then disabled the check and confirmed the state returned to ok with the job row removed. Full server suite passes.

The default targets are ICMP only (`1.1.1.1`, `8.8.8.8`). Checked from a Checkmate pod on our cluster: `ping_group_range` is open, `/bin/ping` carries `cap_net_raw`, and both targets answer, as do TCP 53 and HTTPS. So the defaults work in the image as shipped. A deployment that drops `NET_RAW` would lose ICMP; if you would rather not depend on that, adding `1.1.1.1:53` and `8.8.8.8:53` to the defaults gives a TCP fallback. `egressNotifications` ids are checked to exist in PR 3, where the notification itself arrives, alongside the cascade added here.

ajhollid:
> @ajhollid thanks for the review, all eleven were confirmed. Each is addressed in the two latest commits with a test per point.
> 
> A second pass over the diff turned up a few more, also fixed: the recovery job is scheduled before the degraded transition is recorded; an http probe counts any HTTP response (including 4xx/5xx) as reachable and only a transport failure against egress, so a single target returning 503 no longer degrades the instance; global job rows are addressed by one key (`deleteGlobalJob`); `PingProvider` takes `net` through its constructor like `PortProvider`; integration tests added for `MongoEgressStateRepository` and the global-row delete methods.
> 
> Verified live against a throwaway Mongo: enabled the check through the API, seeded a degraded episode with unreachable targets including IPv6, watched the job run and record probes, then disabled the check and confirmed the state returned to ok with the job row removed. Full server suite passes.
> 
> The default targets are ICMP only (`1.1.1.1`, `8.8.8.8`). Checked from a Checkmate pod on our cluster: `ping_group_range` is open, `/bin/ping` carries `cap_net_raw`, and both targets answer, as do TCP 53 and HTTPS. So the defaults work in the image as shipped. A deployment that drops `NET_RAW` would lose ICMP; if you would rather not depend on that, adding `1.1.1.1:53` and `8.8.8.8:53` to the defaults gives a TCP fallback. `egressNotifications` ids are checked to exist in PR 3, where the notification itself arrives, alongside the cascade added here.

Awesome, thanks for making changes.  I'm liking the shape of this now, it's starting to feel solid.  I will do another review pass.  Thanks for being patient with this process, it's important to get the building blocks right!

ajhollid (approved):
This looks good, thanks for working with me on figuring out the implementation.

One caveat here is that eventually I think it would be good to include some sort of increasing backoff cap, be it arithmetic or geometric, on the scheduling of the egress job, and for that matter the clean up jobs as well.

Currently the backoff there is capped at 5 seconds :thinking: 

These types of job rarely exist so there's no need to poll so frequently in most cases.

Any road, that's a job for another day!  Thank you for your work here 🚀

PR #3963, opened 2026-09-15
Egress self-check (2/4): flag checks taken during degraded egress (#3937)

web-engineer:
## Describe your changes

Second of four PRs for the instance egress self-check (#3937), following #3959. This one connects the egress service to the check pipeline. With `egressCheckEnabled` off (the default) nothing changes at runtime beyond an optional field on the check schema.

- `Check.egressStatus` (`check.type.ts`, `check.model.ts`): optional, `enum` from the `EgressStatuses` tuple. Set on failing checks while the egress check is enabled: `degraded` when no reliability target was reachable (or the instance was already marked degraded), `ok` when one was. Successful checks and older documents carry no value, and `$ne` matches a missing field, so nothing needs backfilling. The repository maps it in `toEntity`, so the evaluate job sees it when it reads unevaluated checks.
- `CheckProducer`: consults `egressService.assessAfterFailure()` only when the target was not reached. Providers that can tell the two apart say so through `peerResponded` on `MonitorStatusResponse` — gRPC reports the peer as having answered for `NOT_SERVING` and every code but `DEADLINE_EXCEEDED` and `UNAVAILABLE`, DNS separates the c-ares codes where the server replied from timeout and connection refused. HTTP, port, WebSocket and Docker share `peerAnsweredFromError`, which holds the errno codes that require the target's own stack to have replied — a refusal, a reset, a TLS handshake that reached a certificate — and HTTP and Docker additionally treat any status code as an answer. Ping and game cannot tell the two apart and stay conservative, as does a proxied HTTP request that fails with no response, where the refusal may be the proxy's. Providers that leave the flag unset fall back to the HTTP-status heuristic, so a 4xx, a 5xx or a content mismatch still proves the target was reached. Attribution follows whether the check leaves the instance rather than what type it is (`isEgressAttributable` in `monitor.type.ts`): a hardware check is an outbound HTTP request to the Capture agent and a Docker daemon can be reached over TCP, so both count; a daemon on a `unix://` socket is local IPC and does not. The check is built before the assessment so the probe's duration stays out of `createdAt`, and a rejection from the egress service is logged and the check recorded unchanged.
- `CheckEvaluator`: a check with `egressStatus: "degraded"` is returned without calling `statusService.updateMonitorStatus`, with a decision that creates no incident and sends no notification. The monitor comes back unchanged, so the status window, running stats and `recentChecks` do not see the check; `lastEvaluatedAt` still advances so it is not re-read.
- Aggregations (`check.query.ts` exports the match): a degraded check is left out of every aggregate, so a spell reads as a gap in monitoring rather than as downtime or as a dip in response time. The exclusion is spread into the top-level `$match` of each query rather than into `$group` accumulators, so no pipeline can count a degraded check by omission. Applied to the team summary, the daily status buckets, the whole uptime facet (`uptimePercentage`, `groupedChecks`, `groupedUpChecks`, `groupedDownChecks`, `groupedAvgResponseTime`), the pagespeed series, and the docker and hardware totals, bucket counts and latest-check reads. Check listings keep degraded checks, so they still appear in the checks table; PR 4 labels them.
- Tests: producer and evaluator unit tests for each branch (success, degraded, ok, disabled, rejection, HTTP response received, non-attributable type); `checksRepositoryEgressExclusion.test.ts` runs each aggregation against a live mongod; `heartbeatJobDegradedEgress.test.ts` runs the full produce → persist → evaluate pipeline through the shared heartbeat harness and asserts no status change, incident or notification while degraded, normal detection otherwise, and normal handling of a 503 and of a docker failure during an episode.
- Maintenance: leaving a maintenance window is now explicit in `CheckProducer`, alongside the entry it already owned. It had been a side effect of evaluating the resulting check, which the degraded short-circuit skips, so a monitor whose window ended during a spell would keep reporting `maintenance`. The monitor is set to `initializing`, which the first evaluated check resolves to up or down.
- Status window: the window is cleared when the first degraded check of a spell lands, beside the maintenance gate that already clears it on entering a window. Without that, the results either side of a spell ended up adjacent — a monitor at `[t, t, t, f, f]` when egress dropped resumed at `[t, t, f, f, f]` on its first real failure afterwards, crossing the threshold on results taken before the outage. It is guarded on the window's length, so it is one write per spell. The monitor still goes down once the rebuilt window holds enough real failures.
- Recovery cadence: the poll interval is fixed at `EGRESS_RECOVERY_POLL_SECONDS` (5 s) and `egressPollIntervalSeconds` is removed from the settings, the validator and the OpenAPI spec. While degraded, checks are flagged from stored state without probing, so the interval bounds the window in which a recovered instance still discards its checks; raising it only lengthened that window. Measured on a live instance, recovery followed the network's return by 3.2 s with an address as the reliability target. With a hostname it took 12–17 s, because `HttpProvider`'s `CacheableLookup` holds a failed lookup for 30 s (`errorTtl`) and the poll keeps reading the cached `ENOTFOUND`; the defaults are addresses and are unaffected. This touches code merged in PR 1.

Known limits, and what was agreed on review:

- `MonitorStats.lastCheckTimestamp` is not advanced by a degraded check, so "last checked" goes stale for the length of a spell. Agreed on review to leave as it is: it is the last time we got a response.
- `/checks/team/summary` excludes degraded checks while `/checks/team` keeps them, so during a spell the Checks page header and the table below it disagree — measured on a live instance at the same date range, the summary returned 67 checks while the listing returned 89 rows, the difference being exactly the degraded ones. Agreed on review to leave until PR 4, which labels degraded rows in the uptime details table; the team Checks table and a filter are on that list too.
- `recentChecks` snapshots (`check.snapshot.ts`, `checkSnapshotResponseSchema`) do not carry `egressStatus`. Degraded checks never reach the window, so nothing is lost today. If you would like the flag on the recent-checks bars in PR 4, the snapshot and its schema need the field.
- `findSummaryByTeamId` previously counted on `metadata.teamId` and `createdAt` only, both resolvable from the timeseries bucket metadata. `egressStatus` is a measurement field, so that predicate now applies after unpacking, and this is the only query where the added clause is the sole non-meta predicate. At 300k checks over 30 days the count goes from 6 ms to 13 ms, examining the same 4327 buckets either way: the cost is the unpacking, not extra reads.

Known follow-ups in `EgressService` (merged in #3959), kept out of this PR to keep it to the check pipeline:

- The probe waits for every target (`Promise.all`, up to the 5 s probe timeout) even when the first has answered. An early return on the first reachable target would bound the delay a failing check pays outside the 5 s reuse window.
- An HTTP response received while the instance is marked degraded is evidence of recovery that is currently left to the poll job.

Next: PR 3 (recovery notification), PR 4 (API route and UI). Compare views on the fork:

- https://github.com/web-engineer/Checkmate/compare/feat/egress-2-check-persistence...feat/egress-3-recovery-notification
- https://github.com/web-engineer/Checkmate/compare/feat/egress-3-recovery-notification...feat/egress-4-api-and-ui

- [x] I have **not** included any files that are not related to my pull request. The only OpenAPI change is the removal of `egressPollIntervalSeconds`.


server/src/domain/monitors/monitor.type.ts
```diff
@@ -50,6 +50,21 @@ export const UptimeDetailsSupportedTypes = ["http", "ping", "port", "game", "grp
 export type UptimeDetailsSupportedType = (typeof UptimeDetailsSupportedTypes)[number];
 export const supportsUptimeDetails = (type: MonitorType): type is UptimeDetailsSupportedType => UptimeDetailsSupportedTypes.some((t) => t === type);
 
+// Types whose transport failure can be the instance's own loss of egress. Hardware (Capture agent) and docker
+// (local socket or host) failures are local, so they are never attributed to egress.
```
ajhollid: Capture agent can be running on any machine anywhere, so it is also subject to egress failure.

Ditto for Docker monitors, remote access to the Docker daemon is supported so egress failure is possible there too.

server/src/worker/worker.check-producer.ts
```diff
@@ -42,6 +46,26 @@ export class CheckProducer implements ICheckProducer {
 		return dockerTlsKey ?? undefined;
 	}
 
+	// Only a failure to reach the target at all can be the instance's own fault. Any HTTP response (4xx, 5xx,
+	// or a 200 with a content mismatch) proves the target was reached, and hardware/docker failures are local.
+	private isTransportFailure(monitor: Monitor, status: MonitorStatusResponse): boolean {
+		return status.status === false && isEgressAttributable(monitor.type) && !isHttpStatusCode(status.code);
```
ajhollid: This doesn't work for gRPC and DNS providers, please check their return values.  They will be marked as transport failed when they are in fact responding

server/src/worker/worker.check-evaluator.ts
```diff
@@ -4,17 +4,52 @@ import { Check } from "@/domain/checks/check.type.js";
...
+			check,
+			statusChange: { monitor, statusChanged: false, prevStatus: monitor.status, code: status.code, timestamp: Date.now() },
+			decision,
+		};
+	};
+
 	evaluate = async (status: MonitorStatusResponse, check: Check, monitor: Monitor) => {
+		if (check.egressStatus === "degraded") {
+			return this.skipDegradedEgressCheck(status, check, monitor);
+		}
```
ajhollid, line 51: On reflection there is an issue here with the degraded path.

Once egress is marked as degraded, all failed checks are marked degraded until egress status is OK, which could be an interval of up to 600 seconds given the current constraints.

That means if egress fails, and then the issue is immediately resolved, we blindly mark all checks for the next 600 seconds as degraded.  

No matter what there's going to be a blind window here, but I suggest we  just set the egress poll interval at something reasonable like 5 seconds rather than allowing a range.

server/src/worker/worker.check-evaluator.ts
```diff
@@ -4,17 +4,52 @@ import { Check } from "@/domain/checks/check.type.js";
...
+			monitor,
+			status,
+			check,
+			statusChange: { monitor, statusChanged: false, prevStatus: monitor.status, code: status.code, timestamp: Date.now() },
+			decision,
+		};
+	};
+
 	evaluate = async (status: MonitorStatusResponse, check: Check, monitor: Monitor) => {
+		if (check.egressStatus === "degraded") {
```
ajhollid, line 49: This early return breaks the maintenance status, which was updated as a side effect of evaluating a check.  

This will now need to be handled explicitly, probably the correct place to do it is in the check producer.

server/src/worker/worker.check-producer.ts
```diff
@@ -74,6 +98,10 @@ export class CheckProducer implements ICheckProducer {
 			throw new Error("No network response");
 		}
 
+		// Step 1c: On a transport failure, ask whether the instance itself can reach anything before blaming the target.
+		// Null means the egress check is disabled (or failed internally) and the check is treated as usual.
+		const egressStatus = this.isTransportFailure(monitor, status) ? await this.assessEgress(monitor.id) : null;
```
ajhollid: This is an async operation, and since it is called before creating the check it's op time is added to the check's `createdAt` timestamp.  Minor issues, but solved for free by moving the creation of the check up before the egress status check, and then setting the flag.

server/src/domain/checks/check.repository.mongo.ts
```diff
@@ -387,11 +390,14 @@ class MongoChecksRepository implements IChecksRepository {
 						monitorId: "$metadata.monitorId",
 						day: { $dateTrunc: { date: "$createdAt", unit: "day", timezone } },
 					},
-					totalChecks: { $sum: 1 },
-					upChecks: { $sum: { $cond: [{ $eq: ["$status", true] }, 1, 0] } },
+					// Counts exclude degraded-egress checks; the response-time average keeps them.
+					totalChecks: { $sum: { $cond: [IS_NOT_DEGRADED_EGRESS_EXPR, 1, 0] } },
+					upChecks: { $sum: { $cond: [{ $and: [{ $eq: ["$status", true] }, IS_NOT_DEGRADED_EGRESS_EXPR] }, 1, 0] } },
 					avgResponseTime: { $avg: "$responseTime" },
```
ajhollid: The average need to be wrapped in the degraded conditional as well

server/src/domain/checks/check.repository.mongo.ts
```diff
@@ -466,6 +472,8 @@ class MongoChecksRepository implements IChecksRepository {
 			{
 				$facet: {
 					uptimePercentage: [
+						// Response-time series (groupedUpChecks / groupedDownChecks) deliberately keep degraded checks so they still show as failures on the graph.
+						{ $match: EXCLUDE_DEGRADED_EGRESS_MATCH },
```
ajhollid: This should be moved up in to the outer match.  The degraded checks should be removed from all facets

ajhollid (commented):
The changes produce some serious inconsistencies with the various graphs and statistics.

The problem is that degraded checks are excluded from some aggergates but inlcuded in others.  

I think the correct approach here is to exclude them from all aggregates and treat it as a gap in monitoring.

As I see it, the only place we should see a "degraded" check si in the list of checks on the Checks page.

server/src/domain/monitors/monitor.type.ts
web-engineer: You're right, and the comment was wrong. A hardware check is `httpProvider.handle(monitor)` — an ordinary outbound request from the instance — so it goes out through our egress like any http monitor, wherever the agent happens to live. A Docker daemon reached over TCP does the same.

The test that matters is whether the check leaves the instance, not what type it is, so `isEgressAttributable` now takes the monitor rather than its type. `hardware` and `docker` are both attributable; the one exemption is a daemon on a `unix://` socket, which is local IPC and cannot fail this way.

The docker and hardware aggregations exclude degraded checks again as a result, along with the docker latest-check reads, so a spell does not leave stale container data showing as current.

Worth separating, since it is a different failure: the probe tells us whether *this instance* can reach the network, not whether the agent's own host can. I ran both halves against a live instance on one hardware monitor, minutes apart — stopping the agent's host with our egress intact, then cutting our egress with the agent's host running:

| What was broken | Stored check | Counted? |
|---|---|---|
| agent host stopped, egress intact | `connect EHOSTUNREACH`, `egressStatus: "ok"` | yes — status window went `[t,t,t,t,f]` |
| egress cut, agent host running | request timeout, `egressStatus: "degraded"` | no — window unchanged |

So a hardware check does consult the assessment, and the assessment answers correctly either way.

server/src/worker/worker.check-producer.ts
web-engineer: Confirmed in both, and this is the one that mattered most, because it fails towards silence rather than towards noise.

`GrpcProvider` returns 5000 for a `NOT_SERVING` reply, the same sentinel as a connection failure, and its error path returns gRPC codes 0–16, none of which are HTTP statuses — so `PERMISSION_DENIED`, `UNIMPLEMENTED`, `UNAUTHENTICATED` and the rest were all read as transport failures despite the server having answered. `DNSProvider` turns every resolver error into `NETWORK_ERROR`, including an authoritative `NXDOMAIN`, which is close to the reason a DNS monitor exists.

Fixed in the providers rather than by dropping the two types, so the distinction lives where the knowledge is. `MonitorStatusResponse` gains `peerResponded`, which a provider sets when it can tell the two apart:

- gRPC reports the peer as having answered for `NOT_SERVING` and for every code except `DEADLINE_EXCEEDED` and `UNAVAILABLE`, or a failure carrying no gRPC code at all.
- DNS separates the c-ares codes where the server replied — `ENOTFOUND`, `ENODATA`, `ESERVFAIL`, `EREFUSED`, `EFORMERR`, `ENOTIMP`, `EBADRESP` — from timeout, connection refused and local faults.

Providers that leave it unset are unchanged: the producer falls back to the HTTP-status heuristic, which is right for anything speaking HTTP. Where both are present the flag wins.

Checked against real servers rather than only in unit tests — a dnsmasq resolver and a `@grpc/grpc-js` health server, both placed where they stay reachable while the instance's egress is cut, so the two cases can be told apart under a real spell:

```
instance marked degraded

dns-answered      queryA ENOTFOUND nope.invalid            (no egressStatus — attributed to the target)
dns-unreachable   queryA ETIMEOUT example.com              egressStatus: "degraded"
grpc-notserving   code 5000  "gRPC service unhealthy (NOT_SERVING)"   (no egressStatus)
grpc-unreachable  code 14    "Name resolution failed …"    egressStatus: "degraded"
```

`dns-answered` went `down` with `statusWindow: [f,f,f,f,f]` during the outage, which is the point: a genuine NXDOMAIN and a service reporting itself unhealthy are still reported, rather than being swallowed with everything else.

server/src/worker/worker.check-evaluator.ts
web-engineer: Agreed. While the state says degraded the assessment short-circuits to stored state without probing, so the blind window is bounded by the recovery poll interval.

Fixed at 5 s and no longer configurable — `EGRESS_RECOVERY_POLL_SECONDS`. `egressPollIntervalSeconds` is removed from the settings type, the model, the repository, the validator and the OpenAPI spec, since the only effect of raising it was a longer blind window. Folded in here rather than raised separately, now PR 1 has merged.

One qualification I did not expect, from measuring it on a live instance by cutting and restoring the network and timing the flip back to `ok`. With an address as the reliability target the window is the poll interval, as you would hope — 3.2 s, against a network that was verifiably usable 3 ms after it came back. With a *hostname* it was 12–17 s across three runs, because `HttpProvider` builds `new CacheableLookup({ maxTtl: 300, errorTtl: 30 })`, so the failed lookup is held for 30 s and the recovery poll keeps reading the cached `ENOTFOUND` until it ages out.

The defaults are unaffected — `1.1.1.1` and `8.8.8.8` are pinged, not resolved — and the same negative cache already applies to ordinary monitors, so this is existing behaviour surfacing in a new place rather than something this PR introduces. Happy to leave it, or to have the probe path bypass the cache if you would rather the ceiling were the poll interval in every configuration.

server/src/worker/worker.check-evaluator.ts
web-engineer: Confirmed, and a regression from this PR. `statusService.ts:202` is what lifts a monitor out of `maintenance` and `initializing`, and the short-circuit skips it, so a monitor whose window ended during a spell would keep reporting `maintenance` while it was in fact being checked and failing.

Moved to the producer as you suggest. It already writes `status: "maintenance"` when a window opens, so both transitions now sit together and neither depends on evaluation running. On leaving the window the monitor is set to `initializing`, which the first evaluated check resolves to up or down, so the end state after that check is what it was before.

Reproduced on a live instance before fixing it — a window ending while egress was down left the monitor reporting `maintenance` three minutes later and counting. With the block in place the same run goes `maintenance` → `initializing` at the moment the window closes, holds `initializing` for the length of the spell, and resolves to `up` on the first real check afterwards.

server/src/worker/worker.check-producer.ts
web-engineer: Agreed — `checkService.toCheck` stamps `createdAt` at build time, so the probe's duration was landing in the timestamp, up to the 5 s probe timeout on the first failure of a spell. Reordered exactly as you describe: build the check, assess, then set the flag.

server/src/domain/checks/check.repository.mongo.ts
web-engineer: Done, by removing the conditional rather than extending it. The exclusion now sits in the top-level `$match`, so the average only ever sees attributable checks and there is no second place to keep in step.

server/src/domain/checks/check.repository.mongo.ts
web-engineer: Done. The facet-level `$match` is gone and the predicate sits in the `$match` feeding the `$facet`, so every facet inherits it — including the response-time series, which now shows a degraded spell as a gap rather than as a run of slow failures.

web-engineer:
Thanks — all eight points are addressed.

The aggregate fix took your framing: a degraded check is now left out of every aggregate, so a spell reads as a gap in monitoring. The exclusion is applied once per query in the top-level `$match` and never inside a `$group` accumulator, which is what let the first version drift — any facet or accumulator nobody thought to change kept counting them. `findPageSpeedDateRangeChecks` was missed entirely before and is covered now.

Your gRPC and DNS point turned out to be the important one. Rather than narrow the set of attributable types, providers can now report whether the peer answered, so an authoritative `NXDOMAIN` or a `NOT_SERVING` reply is never mistaken for our own loss of egress. Together with your Capture and Docker point, attribution now follows whether a check leaves the instance rather than what type it is.

The maintenance regression is fixed in the producer, the recovery poll is fixed at 5 s with the setting removed, and the egress assessment no longer inflates the check's timestamp.

Server suite is 100 suites / 1853 tests. The only OpenAPI change is the removed setting.

I also ran the whole thing against a live instance rather than trusting the stubs — two Docker networks, so the instance's outbound connectivity can be removed without taking its database with it. With 22 degraded checks stored alongside 54 up and 2 genuinely down, `MonitorStats` read `totalChecks: 56, upChecks: 54, downChecks: 2`, and the minutes in which every check was degraded produced no bucket at all in the response-time and up/down series — a hole rather than a downtime bar. Counting them as downtime would have read 69.2% instead of 96.4%.

One measurement for the timeseries question below: at 300k checks over 30 days, the summary count goes from 6 ms to 13 ms with the added `egressStatus` predicate, examining the same 4327 buckets in both cases.

Three things I have left alone and would value a view on:

1. `statusWindow` is frozen across a spell, so the results either side end up adjacent: a monitor at `[t, t, t, f, f]` when egress drops resumes at `[t, t, f, f, f]` on its first real failure afterwards, which can cross the threshold on stale results. Clearing the window on the first non-degraded check would fix it, in the way the producer already clears it when entering maintenance.
2. `MonitorStats.lastCheckTimestamp` is not advanced by a degraded check, so "last checked" goes stale for the length of a spell. Intended, since the check told us nothing about the target.
3. `/checks/team/summary` excludes degraded checks while `/checks/team` keeps them, so during a spell the Checks page header and the table below it disagree. On the instance I was testing against, same team and same date range: the summary returned `totalChecks: 67` while the listing returned 89 rows, the difference being exactly the 22 degraded checks. PR 4 labels degraded rows in the uptime details table; the team Checks table and a filter for them are on the list for PR 4 as well.

server/src/worker/worker.check-producer.ts
```diff
@@ -42,6 +46,31 @@ export class CheckProducer implements ICheckProducer {
 		return dockerTlsKey ?? undefined;
 	}
 
+	// Only a failure to reach the target at all can be the instance's own fault. A provider that can tell the
+	// two apart says so outright; for the rest, any HTTP status code proves the target answered (a 4xx, a 5xx,
+	// or a 200 with a content mismatch).
+	private peerAnswered(status: MonitorStatusResponse): boolean {
+		return status.peerResponded ?? isHttpStatusCode(status.code);
```
ajhollid, line 53: This has surfaced an additional problem I didn't really think about.

What happens if we're in an egress failure state, and a service running on our local network goes down?

For DNS and gRPC it now is marked correctly since they set `peerResponded`, but the other providers do not.  They will be marked "degraded" and their status will stay as "up", when in fact they are down.

We should make sure all services set `peerResponded` to `true`  when the service actually responds.

ajhollid (changes requested):
This is looking pretty good.  I missed a couple questions from your initial PR, so I'll address them here.

1.  Frozen status window:  Let's clear the stauts window on the first non-degraded check, ala Maintenance Windows. 

2.  Stale timestamps are fine, that was the last time we got a response, that can stay as is.

3.  We can leave the summary / check list mismatch until PR 4 when the degraded rows wil be shown.

4.  DNS cache:  Let's leave it alone, that's existing behaviour.

server/src/worker/worker.check-producer.ts
web-engineer: Confirmed, and it is the same failure as the gRPC and DNS one: with `peerResponded` set by only two providers, every other failure fell through to the HTTP-status heuristic, which answers "nothing replied" for anything not speaking HTTP. A service on the local network refusing connections during a spell was read as our own loss of egress, recorded as degraded, and the monitor stayed up.

`peerAnsweredFromError` in `service/network/utils.ts` now holds the errno codes that require the target's own stack to have replied — a refusal, a reset, or a TLS handshake that got as far as presenting a certificate. Four providers share it:

| Provider | Peer reported as having answered when |
|---|---|
| `http` | any status code came back, or the errno proves a reply |
| `port` | `ECONNREFUSED` or `ECONNRESET` from the socket |
| `websocket` | the same errno cases from `ws` |
| `docker` | the daemon returned a status, or the errno proves a reply |
| `hardware`, `pagespeed` | inherited — both are `HttpProvider` calls |

The rule I applied is that `false` is the value that costs nothing. It only changes anything while egress is actually degraded, and there it defers to the probe, which is what the feature is for. `true` is the claim with a price, because it blames the target and can open an incident. So it is set only on evidence, never on inference.

Three cases stay conservative, and are commented where they sit rather than left to be rediscovered:

- **`ping`** — no echo reply is silence, which is what our own loss of egress also looks like.
- **`game`** — gamedig reports a failed query the same way whichever leg failed.
- **`http` through a proxy, with no response** — `ECONNREFUSED` there may be the proxy refusing us, and the proxy is part of this instance's route out. Covered by a test; a status code through a proxy still counts as an answer.

One narrower limit: `ws` rejects a non-101 handshake with a plain `Error` carrying only `Unexpected server response: <code>`, so a WebSocket server answering the handshake with an HTTP error is not distinguished. Parsing that message would work but breaks on any `ws` change; listening for `unexpected-response` would give the status structurally, at the cost of taking over the abort path. Left alone for now unless you would rather it were covered.

web-engineer:
Thanks — taking those in order.

**1. Status window.** Done, and the window is cleared exactly as the maintenance gate does it. Worth flagging one difference from your wording: the maintenance code clears on *entering* a window, not on leaving, so following it means clearing when the first degraded check lands rather than on the first non-degraded one. The end state is the same either way — the window rebuilds from empty and the first real check after the spell lands in a window too short to transition — and clearing at the start needs no extra state to remember that a spell happened. Guarded on the window's length, so it is one write per spell. Say if you meant the other one and I will move it.

Two integration tests asserted the old behaviour and are rewritten, including the one that expected failures either side of a spell to combine. The monitor still goes down, once the rebuilt window holds enough real results to cross the threshold.

**2, 3, 4.** Agreed on all three. The timestamp stays as it is, the summary and listing mismatch goes on the PR 4 list alongside the degraded row labels and the filter, and the DNS negative cache is left alone.

Suite is 101 suites / 1933 tests. No OpenAPI change in this round.

ajhollid (approved):
Thanks for making changes, this looks to be in pretty good shape.

There is an issue with the unix docker sockets being treated as external, but I'll fix that quickly in a separate PR.  Good to go here!

PR #3968, opened 2026-09-16
feat: remove mock data and add redux state to dashboard page

malenacaroline:
## Describe your changes

Replaces the mock data on the Dashboard page with live monitor data fetched from the existing monitors API, and adds empty state messages when no data is available.

Changes:

- Added useDashboardData hook (Pages/Dashboard/hooks/useDashboardData.ts) — uses the existing useMonitorListController to fetch monitors and their summary; computes the type breakdown from the monitors list; refreshes every 30s
- Removed Pages/Dashboard/types.ts — DashboardTypeCount moved to @/Types/Monitor alongside the rest of the monitor types, renamed to MonitorTypeCount
- MonitorStatusCard — now accepts MonitorsSummary | null; shows "No data to show at this moment." when no monitors exist
- MonitorsByTypeCard — shows "No data to show at this moment." when the list is empty. Uses MonitorTypeCount from the shared types

- Added pages.dashboard.noData key

No fix/issue related.

- [x] I took a screenshot or a video and attached to this PR if there is a UI change - no visual changes affected for these changes


client/src/Pages/Dashboard/hooks/useDashboardData.ts
```diff
@@ -0,0 +1,29 @@
...
+			.map(([type, count]) => ({ type, count }))
+			.sort((a, b) => b.count - a.count);
+	}, [monitors]);
+
+	return {
+		isLoading,
+		summary,
+		monitorsByType,
+	};
+};
```
ajhollid: I don't think we need to wrap a hook in a hook, this just transforms and forwards some data.  Let's just call `useMonitorList` and do the transform in the dashboard page so it's clear at a glance what's going on.

client/src/Pages/Dashboard/hooks/useDashboardData.ts
```diff
@@ -0,0 +1,29 @@
+import { useMemo } from "react";
+import { useMonitorListController } from "@/Hooks/useMonitorListController";
+import type { MonitorTypeCount } from "@/Types/Monitor";
+
+export const useDashboardData = () => {
+	const { monitors, summary, isLoading } = useMonitorListController({
+		types: "selectable",
```
ajhollid: This should be the full list of monitor types instead of "selectable"

ajhollid (changes requested):
Looks good for the most part!  Biggest issue is the monitor type list, it should be the full list of monitors.  

Less important but worth a change is teh `useDashboardData` just wraps a hook in a hook, obfuscating what's actually going on.  I don't think the abstraction gains us anything and adds a level of digging a maintainer has to do to figure out what's going on.

ajhollid (approved):
Awesome, looks great!

PR #3973, opened 2026-09-18
Support Capture-backed Docker monitors

### LINKED ISSUE #2328: Monitor docker containers on other machines
From what I can see it is only possible to monitor docker containers on the same machine as the checkmate server is running on, exposing docker.sock access via volume.

What if I want to monitor docker containers on different machines? Would I have to spin up checkmate server installations on each and every one of them?

fabiangigler:
## Describe your changes
Adds support for monitoring Docker containers on remote machines through Capture.

Existing Docker socket and TLS monitors continue to work as before. The monitor form automatically recognizes Capture URLs, requests the Capture secret, and hides unsupported log settings.

Fixes #2328,  https://github.com/bluewave-labs/Checkmate/issues/2389#issuecomment-4639313685

   - [x] (Do not skip this or your PR will be closed) I deployed the application
 locally.
   - [x] (Do not skip this or your PR will be closed) I have performed a
 self-reviewing and testing of my code.
   - [x] I have added i18n support to visible strings.
   - [x] I have **not** included any files that are not related to my pull request,
 including package-lock and package-json if dependencies have not changed.
   - [x] I didn't use any hardcoded values that should be shared or configurable.
   - [x] I made sure font sizes, color choices, and dimensions are referenced from the
 theme.
   - [x] I ran the formatter in the server and client directories.
   - [x] I took a screenshot or video and attached it to this PR.
   
   
grafik


fabiangigler:
Doesn't depend on, but goes nicely with this PR to fix requesting status for more than 4 docker containers in one go via the capture agent: https://github.com/bluewave-labs/capture/pull/159 (works reliably with 47 containers after that, so far)

Once this is finalized, I think it would be nice to make it easier for users to automatically add their infrastructure monitor added via capture as docker host as well, if docker containers are detected on it. I implemented this locally for testing, but it likely introduces a bunch of new i18n labels, so I left it out of the scope of this PR so far!

server/src/service/network/DockerProvider.ts
```diff
@@ -40,6 +42,54 @@ const DOCKER_LOG_MAX_LINE_BYTES = 4096;
 const DOCKER_LOG_TRUNCATION_MARKER = " …[truncated]";
 const DOCKER_LOG_TS_REGEX = /^(\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2})(?:\.(\d{1,9}))?Z /;
 const DOCKER_TLS_DEFAULT_PORT = 2376;
+const CAPTURE_FATAL_ERROR_METRICS = new Set(["docker.client", "docker.container.list"]);
+
+const captureDockerResponseSchema = z.object({
```
ajhollid, line 47: We don't generally use zod to create network response types, though perhaps this is a better approach.

Rather than have two competing ways of doing this, let's stick with the established pattern of declaring network response types in the `/types/network`

If we want to change to a new approach that should be a separate PR that refactors all the response types.

ajhollid (changes requested):
This is well executed 👍   My only change request here is to declare the Caputre payload following the established approach.  I think your approach here is probably better, but I don't want to introduce two compething methods.

Thanks for your work here, it looks good!

server/src/service/network/DockerProvider.ts
fabiangigler: Addressed in 2f3de9797. The Capture response is now declared in `types/network.ts` and the provider uses the typed network payload; the local Zod schema was removed.

fabiangigler:
> This is well executed 👍   My only change request here is to declare the Caputre payload following the established approach.  I think your approach here is probably better, but I don't want to introduce two compething methods.
> 
> Thanks for your work here, it looks good!

Yea that makes sense, updated the validation!

ajhollid (approved):
There are some minor issues here but not worht another round of PR review, we'll fix them in a follow up.  Thanks for working on this feature!

PR #3979, opened 2026-09-21
Egress self-check (3/4): send a notification when instance egress recovers (#3937)

web-engineer:
## Describe your changes

Third of four PRs for the instance egress self-check (#3937), following #3959 and #3963. This one sends the single alert the design allows per episode: when the instance regains outbound connectivity. Nothing is sent when it degrades, since the instance most likely cannot deliver it at that point, and the recovery message carries the start time and duration instead. With `egressNotifications` empty (the default) the recovery is logged and nothing is sent.

- `NotificationType` gains `egress_recovered`. It is the first instance-level notification: every existing type is about a monitor, and every provider reads `message.monitor`, so the message describes the instance through a fixed `MonitorInfo` (`EGRESS_MONITOR_INFO`: id `egress`, type `system`, url set to the client host) rather than widening the message shape. Providers build "view incident" links only when `content.incident` is set, and Rocket.Chat's link builder returns nothing for a type that has no details page, so the message carries no dead links.
- `NotificationMessageBuilder.buildEgressRecoveredMessage(state, clientHost)`: title, summary and details from the persisted egress state (degraded since, recovered at, duration, targets probed), severity `success`.
- `NotificationsService.sendEgressRecoveredNotification(state, ids)`: resolves the ids from `egressNotifications` and fans out through the same `send()` the monitor alerts use. The fan-out and its success/failure summary moved into `sendToAll()` in a separate refactor commit so both callers share it; `send()` lost three parameters it never read.
- `EgressService.checkRecovery`: sends once, on the process whose `markRecovered` performed the transition, after the recovery job row has been released. A second worker that polls in the same window sees `recovered` as null and sends nothing. The send is last and best effort: by then the row is gone and the state reads `ok`, so a rerun would have nothing to send against. A rejection is logged at error level with the notification ids rather than thrown, so it does not surface as a queue failure for a job that has completed. A crash between the transition and the send loses that one notification rather than risking two.
- `SettingsController.updateAppSettings`: rejects unknown ids in `egressNotifications` with 422, through `notificationsService.findNotificationsByIds`, as the `globalProxyId` check does through the proxies service. Settings are instance-wide and admin-only, so the ids are not team-scoped, matching `globalProxyId`.
- Email gets a subject for the new type. PagerDuty maps `egress_recovered` to `event_action: "resolve"` alongside the other recoveries: a resolve for an unknown dedup key is accepted as a no-op, whereas the previous default of `trigger` would have opened an incident for connectivity being restored. PagerDuty users therefore see nothing for this event; the other channels carry it. I have not confirmed the no-op against a live routing key; if you know otherwise, say and I will switch it.
- Settings are read through the worker's cached copy (60 s TTL, per process, as for `egressCheckEnabled` since #3959), so a channel added within a minute of a recovery can miss that one notification.
- Tests: message content and severity; a rejected send being logged with the job still released; fan-out to the configured channels, the no-channels branch, and the false return when a provider fails; `checkRecovery` sending exactly once and not when another process won the transition; the PagerDuty mapping.

No OpenAPI change: `egressNotifications` was added to the settings schema in #3959, and notification types are not in the spec.

Rebased onto `develop` after #3963 and #3977. Against #3978 it conflicts in four hunks, all mechanical: the `EgressService` constructor call in `services.worker.ts`, and the `MonitorActionDecision` import path in `notification.message-builder.ts` and its two test files, which #3978 moves to `worker.interface.ts`. Happy for either to go first.

Next: #3980 (API route, banner, checks labels and filter, settings card), open as a draft stacked on this PR.

- [x] I have **not** included any files that are not related to my pull request.

🔧


server/src/api/controllers/settingsController.ts
```diff
@@ -79,6 +83,16 @@ class SettingsController implements ISettingsController {
 		}
 
 		const previousSettings = await this.settingsService.getDBSettings();
+		if (validatedBody.egressNotifications && validatedBody.egressNotifications.length > 0) {
+			const requestedIds = [...new Set(validatedBody.egressNotifications)];
+			const notifications = await this.notificationsService.findNotificationsByIds(requestedIds);
```
ajhollid, line 88: This method is unscoped, meaning an admin from Team A will be able to see Team Bs notifications.  It should either be scoped by teamId or restricted to super admin only.  I believe scoping by teamID makes most sense.

server/src/domain/notifications/providers/pagerduty.ts
```diff
@@ -74,8 +74,11 @@ export class PagerDutyProvider extends NotificationProvider {
 	}
 
 	private buildPagerDutyPayload(notification: Notification, message: NotificationMessage): AlertPagerDutyPayload {
-		// Map our notification type to PagerDuty event_action
-		const eventAction = message.type === "monitor_up" || message.type === "threshold_resolved" ? "resolve" : "trigger";
+		// Map our notification type to PagerDuty event_action. Every recovery resolves, including egress_recovered,
+		// which has no matching trigger. A resolve with no open alert under its dedup key is expected to be accepted
+		// and do nothing, whereas a trigger would open an incident for connectivity being restored.
+		const eventAction =
+			message.type === "monitor_up" || message.type === "threshold_resolved" || message.type === "egress_recovered" ? "resolve" : "trigger";
```
ajhollid, line 81: As noted in the comments, PagerDuty now silently does nothing.

We should either use a low severity trigger for egress or just disallow pagerduty at all from egress notifications.  I suggest the former.

server/src/api/controllers/settingsController.ts
```diff
@@ -79,6 +83,16 @@ class SettingsController implements ISettingsController {
 		}
 
 		const previousSettings = await this.settingsService.getDBSettings();
+		if (validatedBody.egressNotifications && validatedBody.egressNotifications.length > 0) {
+			const requestedIds = [...new Set(validatedBody.egressNotifications)];
```
ajhollid, line 87: The validation allows for hex ids with both upper and lower case letters, however mongoose always returns lower case.  Most expedient solution is to normalize to lower case before (or while) creating the set.

server/src/api/controllers/settingsController.ts
```diff
@@ -79,6 +83,16 @@ class SettingsController implements ISettingsController {
...
+			const notifications = await this.notificationsService.findNotificationsByIds(requestedIds);
+			const foundIds = new Set(notifications.map((notification) => notification.id));
+			const missing = requestedIds.filter((id) => !foundIds.has(id));
+			if (missing.length > 0) {
+				throw new AppError({ message: `Referenced notification does not exist: ${missing.join(", ")}`, status: 422 });
+			}
+		}
+
```
ajhollid, line 95: Above we filter out duplicates but then the raw validated body is passed straight along.  We should pass on the deduped values instead.

server/src/domain/notifications/notification.service.ts
```diff
@@ -107,13 +110,31 @@ export class NotificationsService implements INotificationsService {
...
-				method: "sendNotifications",
+				method,
 			});
 		}
 		// Return true if all notifications succeeded
 		return succeeded === notifications.length;
 	};
 
+	sendEgressRecoveredNotification = async (state: EgressState, notificationIds: string[]) => {
+		const notifications = await this.notificationsRepository.findNotificationsByIds(notificationIds);
```
ajhollid, line 121: No need for a query if `notificationIds` is length 0, lets return early if that's the case.

server/src/domain/notifications/notification.service.ts
```diff
@@ -107,13 +110,31 @@ export class NotificationsService implements INotificationsService {
...
+			this.logger.info({
+				message: "Egress recovered but no notification channels are configured for it",
+				service: SERVICE_NAME,
+				method: "sendEgressRecoveredNotification",
+			});
+			return true;
+		}
+
+		const settings = this.settingsService.getSettings();
+		const clientHost = settings.clientHost || "Host not defined";
```
ajhollid, line 132: "Host not defined" is declared twice now, this should be extracted or they must be manually kept in sync.

ajhollid (changes requested):

